PR #56, opened 2026-05-01
Add 'Visit web UI' click-through on the device card

bdraco:
## Summary

Restores the legacy ESPHome dashboard's click-through to the device's on-board web UI when the YAML includes a `web_server:` block.

The backend already exposes `web_port: number | null` on `ConfiguredDevice` (populated from `StorageJSON.web_port`, set by esphome at compile time). This PR wires the frontend side: when `web_port` is non-null, the device's context (kebab) menu shows a **Visit web UI** item that opens `http://<address>[:port]` in a new tab.

URL shape matches the legacy formula:
- `http://` (the `web_server` component is plain HTTP)
- omit the port when it's 80
- prefer `device.address` (mDNS hostname like `foo.local`) and fall back to `device.ip` if address is empty

Implemented as a menu item rather than an inline button on the card so the action row stays compact — the visit action is useful but not as frequent as Edit / Install / Update. Pure navigation, so the click handler calls `window.open(..., "_blank", "noopener")` directly instead of bouncing through the parent like the other menu items do.

Translations added for en / fr / nl.

## Test plan

- [x] Existing 93 tests still pass.
- [x] TypeScript lint clean.
- [ ] Visual check: a device with \`web_server:\` shows the menu item; one without doesn't.
- [ ] Click opens \`http://<host>[:port]\` in a new tab; non-default ports render correctly.

### DIFF
--- a/src/components/dashboard/table-row-menu.ts
+++ b/src/components/dashboard/table-row-menu.ts
@@ -8,6 +8,7 @@ import {
   mdiDownload,
   mdiFileDownloadOutline,
   mdiKeyVariant,
+  mdiOpenInNew,
   mdiPencil,
   mdiRenameOutline,
   mdiUpload,
@@ -31,6 +32,7 @@ registerMdiIcons({
   download: mdiDownload,
   "file-download-outline": mdiFileDownloadOutline,
   "key-variant": mdiKeyVariant,
+  "open-in-new": mdiOpenInNew,
   pencil: mdiPencil,
   "rename-outline": mdiRenameOutline,
   upload: mdiUpload,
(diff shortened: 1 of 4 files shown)


bdraco:
Addressed both Copilot points in 61ff6eb:

1. **Menu item no longer renders when there's no host.** Pulled into a `_renderVisitWebUi` helper that returns `nothing` unless we have both a non-null `web_port` *and* a non-empty `device.address || device.ip`. `ip` stays empty until the device is seen online, so the previous wider gate on just `web_port` could surface a click that did nothing.

2. **Switched from `window.open(url, "_blank", "noopener")` to a real `<a target="_blank" rel="noopener noreferrer">`.** Browser-enforced `rel` attributes are uniform across browsers and also suppress the `Referer` header (which the `window.open` flag doesn't). Matches the codebase's existing pattern (`markdown.ts`, `wizard-step-board.ts`, `device-board-info.ts`, etc.). CSS reset on `.menu-item--link` keeps the anchor looking like the surrounding `<div>` items.

PR #62, opened 2026-05-02
Link loaded-integration tags in the device drawer to esphome.io docs

bdraco:
## Summary

Each tag in the device drawer's **Loaded Integrations** section is now a clickable link when the integration name resolves to an esphome.io page; tags without a docs hit fall back to the plain `<span>`.

Linked tags get the dashboard's primary colour with a subtle hover so "this opens something" reads at a glance, but the rounded-pill shape is unchanged so the row stays visually consistent.

### Where the data comes from

The mapping is fetched once per WebSocket connection from the backend's `components/get_integration_docs` WS command (added in **esphome/device-builder#57**), stored on `app-shell.ts`, and provided down the tree via the new `integrationDocsContext`. A failed fetch is non-fatal — `_integrationDocs` stays `{}` and every tag renders as plain text, same behaviour as today.

The backend's lookup is catalog-derived (top-level component → category landing → stem alias, in priority order), so a present URL is also a guarantee that the docs page exists.

### Files

- `src/api/esphome-api.ts` — new `getIntegrationDocs()` client method.
- `src/context/{contexts,index}.ts` — new `integrationDocsContext`.
- `src/components/app-shell.ts` — provides the context, fetches once on `onConnected`.
- `src/components/dashboard/device-drawer-content.ts` — consumes the context, swaps `<span class="tag">` for `<a class="tag tag--link">` when a URL is present, plus a `.tag--link` style block.

## Test plan

- [x] Lint clean.
- [x] All 108 existing tests pass.
- [ ] Pairs with backend PR #57 — once that lands, open the device drawer; integrations like `api`, `wifi`, `mdns`, `sensor`, `ltr390` show as links and open the matching esphome.io page in a new tab.
- [ ] Integrations the catalog doesn't know (`runtime_stats`, `web_server_id`, etc.) stay rendered as plain tags.
- [ ] Without the backend deployed, the drawer still renders — every tag falls back to plain text.

### DIFF
--- a/src/api/esphome-api.ts
+++ b/src/api/esphome-api.ts
@@ -674,6 +674,19 @@ export class ESPHomeAPI {
     return this.sendCommand<PagedComponentsResponse>("components/get_components", args);
   }
 
+  /**
+   * Map of integration name → esphome.io docs URL for every
+   * loaded-integration name we can resolve. Names with no docs page
+   * are simply absent from the map; the dashboard renders those as
+   * plain text. Fetched once at app load — the dataset only refreshes
+   * with a backend release.
+   */
+  async getIntegrationDocs(): Promise<Record<string, string>> {
+    return this.sendCommand<Record<string, string>>(
+      "components/get_integration_docs",
+    );
+  }
+
   // ─── Config Commands ──────────────────────────────────────
 
   /** Get ESPHome and server version. */
--- a/src/api/esphome-api.ts
+++ b/src/api/esphome-api.ts
@@ -680,11 +680,31 @@ export class ESPHomeAPI {
    * are simply absent from the map; the dashboard renders those as
    * plain text. Fetched once at app load — the dataset only refreshes
    * with a backend release.
+   *
+   * The WS layer doesn't enforce a shape, so we filter the payload to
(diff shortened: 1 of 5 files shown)


bdraco:
Addressed all four Copilot points in 8b90759:

1. **Untyped WS payload** (`esphome-api.ts`) — `getIntegrationDocs()` now coerces the response: anything that isn't a plain object becomes `{}`, and entries with non-string keys/values are filtered out. Consumers can rely on the result being `Record<string, string>` without further checks.

2. **Untrusted href in the drawer** (`device-drawer-content.ts`) — added an `_isSafeDocsUrl()` guard that requires `https:` scheme and `esphome.io` host (where the catalog actually points). Anything else falls back to the plain `<span>` so a smuggled `javascript:` / `data:` URL can't trigger script execution.

3. **`_loadIntegrationDocs()` shape** — point 1 already enforces this at the API boundary, so the existing app-shell try/catch fallback to `{}` is enough; no extra sanitization needed in app-shell.

4. **Missing focus state** — `:focus-visible` on `.tag--link` now gets the hover-style background plus a 2px outline so keyboard tabs are visible and consistent with other interactive elements.

PR #69, opened 2026-05-02
Cancel the prior follow stream when the install dialog reopens

bdraco:
## Symptom

Clicking the busy spinner on a card while a job was still running reopened the install/progress dialog. The trailing log lines duplicated once per reopen — five clicks, five copies:

```
INFO Upload took 24.95 seconds, waiting for result...
INFO Upload took 24.95 seconds, waiting for result...
INFO Upload took 24.95 seconds, waiting for result...
INFO Upload took 24.95 seconds, waiting for result...
INFO Upload took 24.95 seconds, waiting for result...
INFO OTA successful
INFO OTA successful
INFO OTA successful
INFO OTA successful
INFO OTA successful
```

## Cause

`command-dialog.ts` had several sites that bared `this._streamId = ""` (in `open()`, `followJob()`, `close()`, `_onDialogHide()`, `_stop()`, `_start()`). That cleared the local id but never told the backend to cancel or asked the API client to drop the local handler. The previous `firmwareFollowJob`'s `onOutput` closure was bound to the dialog instance and kept appending to `this._lines` — each click layered another live subscription, so the same output came in N times.

## Fix

Extract a `_detachStream()` helper that:

- Calls `this._api.stopStream(this._streamId)` when set — that drops the local handler (so its closure stops appending) and tells the backend to cancel the queued streaming task.
- Zeros `_streamId`.

Use it at every site that previously bare-cleared `_streamId`, so a reopen always starts from a single live subscription.

Also applied the same fix to `firmware-install-dialog._close()` — same shape, less likely to be hit on the user's path but the bug is identical.

## Test plan

- [x] Lint clean.
- [x] All 124 existing tests pass.
- [ ] Reproduce the original report: click "Installing" on the card 5x while the job runs. Output should appear once per line, not five times.
- [ ] Reopen mid-stream still shows the live tail (the new follow's snapshot replay covers this — same as today).

### DIFF
--- a/src/components/command-dialog.ts
+++ b/src/components/command-dialog.ts
@@ -245,7 +245,7 @@ export class ESPHomeCommandDialog extends LitElement {
     this._lines = [];
     this._statusMessage = "";
     this._jobId = "";
-    this._streamId = "";
+    this._detachStream();
     this._dialog.open = true;
     this._resetAnsiLogScroll();
     this._start();
@@ -276,15 +276,35 @@ export class ESPHomeCommandDialog extends LitElement {
     this._lines = [];
     this._statusMessage = "";
     this._jobId = job.job_id;
-    this._streamId = "";
+    // Cancel any prior follow before starting a new one. Without
+    // this, every reopen of the dialog (clicking the busy spinner
+    // again while a job is still running) layered on a fresh
+    // ``firmwareFollowJob`` while the previous one was still pumping
+    // ``onOutput`` callbacks into ``this._lines`` — each new line
+    // appeared once per leaked subscription, so output duplicated
+    // five times after five clicks.
+    this._detachStream();
     this._dialog.open = true;
     this._resetAnsiLogScroll();
     this._followJob(job.job_id);
   }
 
   public close() {
-    this._streamId = "";
+    this._detachStream();
     this._dialog.open = false;
+  }
+
+  /**
+   * Tear down the active stream subscription, both client-side
+   * (drops the local handler so its closure stops appending to
+   * ``_lines``) and backend-side (the queued task can stop pushing
+   * lines for a job we're no longer watching). Safe to call when
+   * no stream is active.
+   */
+  private _detachStream() {
+    if (!this._streamId) return;
+    this._api.stopStream(this._streamId).catch(() => {});
+    this._streamId = "";
   }
 
   private get _title(): string {
@@ -350,7 +370,7 @@ export class ESPHomeCommandDialog extends LitElement {
   // ─── Command execution ─────────────────────────────────────
 
   private async _start() {
-    this._streamId = "";
+    this._detachStream();
     this._jobId = "";
     this._state = "running";
     this._lines = [];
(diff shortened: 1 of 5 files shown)


bdraco:
Rebased on latest main and picked up the third Copilot point from PR #68 (the Web Serial reader leak that was logged as a follow-up).

**What's covered now:**

1. **Backend WS streams** (`command-dialog`'s `firmwareFollowJob` / `validate`) — `_detachStream()` helper called from `open`, `followJob`, `close`, `_onDialogHide`, `_stop`, `_start` retry. Resolves the original duplicate-output report.

2. **Web Serial reader loop** (`dashboard-actions.streamSerialToDialog`) — now returns a cancel function that calls `reader.cancel()` to settle the pending `read()` and `releaseLock()`s in `finally` so the decoder pipeline tears down cleanly. `pipeTo()` rejection that follows is caught so it doesn't surface as an unhandled promise rejection. `logs-dialog` stores the cancel on `_serialCancel` and disposes it at the same three lifecycle points WS streams already get cleaned up at: `setSerialCancel` (new caller registers, prior is disposed), `openPassive` (fresh session), `_stopStreaming` (Stop / Close / dialog hide).

3. **firmware-install-dialog** — same `stopStream` hygiene applied to `_close()` even though it's less likely on the user's path; the bug shape is identical.

bdraco:
Addressed the new Copilot point in 1dc00cc. They're right — `_close()` only fires from the explicit Close button. `wa-dialog`'s header X and Escape route through `wa-after-hide → _onClose`, which previously just flipped `_open` and left the `firmwareFollowJob` subscription attached. The next `installWebSerial` then ran `_init` (which bare-cleared `_streamId`) and the leaked subscription's `onOutput` kept feeding into the new session's `_logLines`.

Fix: extracted a `_detachStream()` helper (mirrors the one in `command-dialog.ts`) and called it from every close-shaped path:

- `_close()` — explicit Close button
- `_onClose()` — `wa-after-hide` (header X / Escape / light-dismiss / programmatic close)
- `_init()` — pre-reopen reset, in case the prior run was dismissed via `_onClose` and never landed in `_close`

Lint clean, 124 tests pass.

bdraco:
Two more Copilot finds, both addressed in 56f48e8:

1. **`streamSerialToDialog` never closed the `SerialPort`** — only released the reader lock. Each passive-logs reopen leaked an open OS port handle; enough reopens eventually trip the per-tab Web Serial ceiling and the device can't be reconnected without a page refresh. The cancel now chains `reader.cancel().finally(port.close())` so the handle goes back to the browser.

2. **`_detachStream()` stranded `_compileAndWait`** — removing the local follow_job handler meant `onResult` / `onError` could no longer fire, so the Promise inside `_compileAndWait` (only settled by those callbacks) hung forever after a mid-flight detach. The awaiter, `_startWebSerialInstall`, leaked one pending task per reopen.

   Fix: capture the Promise's `reject` on `this._compileReject` when the body runs, clear it on natural completion (onResult / onError / the firmwareCompile catch), and have `_detachStream()` reject any pending promise + `firmwareCancel` the underlying job so the backend stops working too. Null-guarded so a normal completion can't double-reject.

Tightened the surrounding paths while in there: `_close()` and `_init()` no longer bare-clear `_jobId` since `_detachStream` covers it. Lint clean, 124 tests pass.

PR #72, opened 2026-05-02
Collapse the Logs button to icon-only when the Visit Web UI button is shown

bdraco:
## Symptom

In a four-up card grid (and on tighter desktop viewports generally), cards that expose a **Visit Web UI** button — i.e. devices with `web_server:` in their YAML, so the backend reports a non-null `web_port` — have one extra button-width to fit on the row. Edit + Install + Logs + Visit + kebab squeezed the labelled buttons into truncation territory, so the row read as a smoosh of icons with cropped text.

## Fix

When `webUrl` is set, drop the **Logs** label and render the button icon-only (same `--tile` shape Visit already uses). Edit and Install keep their labels — they're more frequent and the label carries weight beyond the icon — and Logs is the right one to collapse because the `console` icon is already used elsewhere in the dashboard for the same action. `aria-label` and `title` carry the accessible name so screen readers and hover tooltips still say "Logs".

When there's no Visit button (no `web_server` block), the row keeps the old layout — the extra horizontal real estate goes to the labelled Logs button as before.

## Test plan

- [x] Lint clean.
- [x] All 124 existing tests pass.
- [ ] Visual on a 4-up card grid: cards with Visit button — Edit + Install labelled, Logs icon-only, Visit icon, kebab. No truncation.
- [ ] Cards without Visit button — unchanged from today (Edit + Install + labelled Logs + kebab).
- [ ] Hover the icon-only Logs button: tooltip says "Logs".

### DIFF
--- a/src/components/device-card.ts
+++ b/src/components/device-card.ts
@@ -586,11 +586,15 @@ export class ESPHomeDeviceCard extends LitElement {
                       `
                     : nothing}
                 <button
-                  class="action-btn action-btn--ghost"
+                  class="action-btn action-btn--ghost ${this.webUrl ? "action-btn--tile" : ""}"
                   @click=${() => this._emit("open-logs")}
+                  aria-label=${this._localize("dashboard.drawer_logs")}
+                  title=${this._localize("dashboard.drawer_logs")}
                 >
                   <wa-icon library="mdi" name="console"></wa-icon>
-                  ${this._localize("dashboard.drawer_logs")}
+                  ${this.webUrl
+                    ? nothing
+                    : this._localize("dashboard.drawer_logs")}
                 </button>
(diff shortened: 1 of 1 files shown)


bdraco:
Addressed Copilot's point in 1100bb9. The previous version attached `title` / `aria-label` unconditionally, so the labelled variant (no `webUrl`) picked up a redundant hover tooltip even though its visible text was already the accessible name.

Split the render into two branches keyed on `webUrl`:

- **`webUrl` present** — `action-btn--tile` (icon-only padding), `title` + `aria-label` carry the name now that the visible text is gone.
- **`webUrl` missing** — exactly today's shape: bare button, no `title`, the icon + label is the accessible name.

Lint clean, 124 tests pass.

PR #224, opened 2026-05-07
Show core configs in dep-detour catalog search (fixes #383)

### LINKED ISSUE #383: [Bug] Core Configuration dependant Component
### What happened?

Using debug: as an example 
 
- Create/Open Device
- Select Components/Add Component
- Search debug, two sensors are shown (free and device... I think this is also a bug, should be several more sensors, filtered by cpu support ideally)
- Click Add in the Debug Component free option, the following dialog is shown
Image
 
- As requested select Add debug
- The wizard takes the user back to the Component search dialog where the Debug Core Configuration is not available so the debug sensors can not be added

For components that depends on a Core Configuration that wasnt add previously, the Add Component wizard cannot be completed because the user cannot add Core Configuration elements from this wizard and not useful information is presented to the user.

### Dashboard version

beta21

### How did you install it?

Wheel from a GitHub release

### Where is it running?

Local — Linux

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

Fixes esphome/device-builder#383.

The reproduction from the issue:

1. Open a device, click "Add Component", search "debug" → catalog returns the two platform sensors (`sensor.debug` / `text_sensor.debug`).
2. Pick `sensor.debug.free`. The form opens, its missing-deps banner detects that `debug` (the bare top-level `debug:` block) isn't in the YAML and prompts "Add debug".
3. Clicking the prompt forwards the user back to the catalog filtered to `debug` — but the **Debug Core Configuration** entry (`core.debug` in the catalog) doesn't appear, so the user can't actually add the dep the form just told them they needed.

Cause: the regular "Add component" dialog passes `excludeCategories=CORE_CATEGORIES` to the catalog query so users go through the dedicated "Add core configuration" entry point for `core` / `ota` / `update` instead of seeing those mixed into a sensor browse. That exclusion is correct in the regular flow but wrong in the dep-detour case — the form has already narrowed the user's intent to "I need a top-level `<domain>` block", so a core-categorized entry matching that domain is exactly what they want.

Flip the exclusion off when `_returnTo !== null` (the dialog's existing dep-detour signal). Now the catalog filtered to `debug` surfaces `core.debug` (alongside the platform sensors), and the user can add the bare `debug:` block from the same dialog the form forwarded them to.

The exclusion logic moves to a tiny pure helper (`chooseExcludeCategories`) so the three-state contract — *regular browse* / *core-locked* / *dep-detour* — can be unit-tested without spinning up a DOM env to render the dialog. Four tests cover the matrix and pin the regression so a refactor that drops the `isInDepDetour` term fails loudly.

- fixes esphome/device-builder#383

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/add-component-dialog-categories.ts
+++ b/src/components/device/add-component-dialog-categories.ts
@@ -1,0 +1,36 @@
+import { CORE_CATEGORIES, type ComponentCategory } from "../../api/types.js";
+
+/**
+ * Decide which categories to *exclude* from the catalog query
+ * for the given dialog mode. Lives in its own module so the
+ * three-mode contract can be unit-tested without spinning up a
+ * DOM env to render the dialog.
+ *
+ * Three states the caller passes:
+ *
+ * - **Core-config dialog** (``isCoreLocked = true``). The
+ *   "Add core configuration" entry point already locks the
+ *   query *to* ``CORE_CATEGORIES``; excluding them too would
+ *   produce an empty catalog. Returns ``[]``.
+ * - **Dep-detour mode** (``isInDepDetour = true``). The form
+ *   just told the user "you need a top-level <domain> first"
+ *   and forwarded them back to the catalog filtered to that
+ *   domain. The missing dep is sometimes itself a core block
+ *   (e.g. a ``sensor.debug`` needs the bare ``debug:`` block
+ *   whose catalog entry is ``core.debug``); hiding it from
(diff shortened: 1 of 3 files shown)


masterofnonemx:
Ummm, now I see why only to components show up.

PR #340, opened 2026-05-15
Skip auto-id for top-level singleton components

### LINKED ISSUE #776: Singleton components get misleading `_1` suffix in auto-generated ids (e.g. `web_server_1`)
When adding `web_server` via the builder, the generated YAML includes `id: web_server_1`. The id itself is fine (useful for `!extend` overrides in packages), but the `_1` suffix is misleading. `web_server` is a singleton, so there will never be a `web_server_2`. The numbering implies a multiplicity that ESPHome doesn't actually support.

This likely affects other singleton-style components too: `captive_portal`, `mdns`, `logger`, `api`, `ota`, `improv_serial`, `wifi`, `time`, `safe_mode`, etc. Worth auditing which generated ids should drop the numeric suffix.

**Reproduction:**
1. Open device-builder
2. Add a `web_server` component to a device
3. Inspect the generated YAML, which contains `id: web_server_1`

**Expected:** `id: web_server` (no numeric suffix) for singleton components.

**Suggested fix:** Maintain a list of singleton component types and emit their ids without the `_N` suffix.

bdraco:

When a top-level singleton component (`multi_conf: false`, no `.` in the catalog id — e.g. `web_server`, `mdns`, `logger`, `api`, `ota`, `captive_portal`, `wifi`, `improv_serial`, `safe_mode`, `time`) is added via the catalog, the dashboard no longer pre-fills its `id:` field. The previous `id: web_server_1` was misleading (it implied a non-existent `_2`), and the bare-slug alternative would collide with ESPHome's `web_server::` C++ namespace at codegen time. These components aren't referenced by id from elsewhere in the YAML, so skipping the auto-id is harmless; power users who need an id for `!extend` overrides in packages can type one in.

Platform entries (`switch.gpio`, `sensor.dht`, ...) and repeatable top-level blocks (`multi_conf: true`, e.g. `script`, `i2c`, `spi`) keep their suffixed auto-ids (`switch_gpio_1`, `script_1`, ...). The dashboard's id-reference pickers (`config-entry-id-reference-renderer.ts` → `findReferencedComponents` in `util/config-entry-yaml-scan.ts`) only surface components whose YAML block has an `id:` line — a switch without an id is invisible to every "pick a switch" dropdown in the editor (automation actions, `on_press → switch.toggle`, etc.); an `i2c:` bus without an id can't be selected as `i2c_id:` from devices on it. So the prefilled id is load-bearing for cross-references inside the dashboard, not just convenience for hand-edited YAML.

The singletons we skip have no such pickers pointing at them — nothing in the schema has `references_component: web_server` (or `mdns`, `api`, ...) — so omitting the id costs nothing in the editor.

The id-generation and YAML id-scanning helpers were extracted from `esphome-add-component-form` into `src/util/default-component-id.ts` so the policy is unit-testable; the form delegates to those helpers and skips id-seeding when `generateDefaultComponentId` returns `null`.

- fixes https://github.com/esphome/device-builder/issues/776

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/add-component-form.ts
+++ b/src/components/device/add-component-form.ts
@@ -17,6 +17,10 @@ import {
   type ValidationError,
 } from "../../util/config-validation.js";
 import { seedBoardPinDefaults } from "../../util/board-pin-defaults.js";
+import {
+  collectExistingIds,
+  generateDefaultComponentId,
+} from "../../util/default-component-id.js";
 import { renderMarkdown } from "../../util/markdown.js";
 import { setIn } from "../../util/nested-values.js";
 import { registerMdiIcons } from "../../util/register-icons.js";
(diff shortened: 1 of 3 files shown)


jesserockz:
This wont work, the id cannot be a component name. This is due to the variable name then conflicting with the c++ namespace

jesserockz:
We ideally should simply omit automatically generating ids, or use the same format that esphome actually generates in codegen if it is not specified in config

bdraco:
> This wont work, the id cannot be a component name. This is due to the variable name then conflicting with the c++ namespace

For sure, you are too quick. Its still processing my feedback.

bdraco:
so we can't drop them from the platforms as otherwise the code breaks that allows you to select a switch.  But the top level ones can go away

bdraco:
@bluetoothbot review

PR #344, opened 2026-05-16
Skip save-time validate_yaml when the linter just validated the buffer

bdraco:

The editor's CodeMirror linter
(`util/yaml-lint-backend.ts`) runs `validateYaml` at typing-stop +
600ms, and `pages/device.ts._saveYaml` runs it again on Save
click. Same buffer, same backend pass, two round-trips — visible
as 7-9s click-to-popup latency on a slow machine when the
underlying `esphome vscode` validation is heavy
(`external_components` / `packages:` / `!include`).

Cache the linter's last successful result keyed on
`(configuration, content)` and expose `getLastValidatedResult`.
`_saveYaml` consults it first and skips its own `validateYaml`
call when the buffer matches exactly. Different content or no
linter result → original validate-then-save path unchanged, so
edge cases (network blip during linter, save fired before any
linter run) still validate authoritatively.

Companion to the backend PR esphome/device-builder#892 which
adds a `_EditorSession` content-hash cache. Either alone is an
improvement; together the save click on an already-linted buffer
makes zero validate work happen.

- fixes <Slack report from MasterOfNone, 2026-05-16>

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/pages/device.ts
+++ b/src/pages/device.ts
@@ -34,6 +34,7 @@ import { UnsavedGuard } from "../util/unsaved-guard.js";
 import { registerMdiIcons } from "../util/register-icons.js";
 import { sectionAtLine, sectionKeyOf } from "../util/yaml-sections.js";
 import { resolveSectionForUrlLine } from "../util/url-line-resolver.js";
+import { getLastValidatedResult } from "../util/yaml-lint-backend.js";
 import { summarizeValidation } from "../util/yaml-validation-summary.js";
 import { devicePageStyles } from "./device-styles.js";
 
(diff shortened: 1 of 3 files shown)


bdraco:
@bluetoothbot review

bdraco:
Thanks for the merge-ready ack. One follow-up landed in 1ec86e7 (post-review): the cache had no TTL while the backend has 60s, so a save more than a minute after the linter ran could serve a stale 'all clear' and skip the backend re-validate that would have caught it. Bounded the frontend cache to 60s as well so staleness semantics are symmetric on both paths. New test `returns null when the cached entry is past the TTL window` uses a `performance.now` spy to drive deterministic expiry.

bdraco:
x-post from backend pr:

definitely better to work with on heavy configs. still a lot of cpu churn for revalidate on heavy configs, but any improvement there is win.  Will need to dig into lvgl more as https://github.com/esphome/esphome/pull/16478 is just the tip of it and we have 0 benchmarks for it.

PR #355, opened 2026-05-19
command-dialog: batch streamed output lines per animation frame

### LINKED ISSUE #348: Opening a historical firmware build log is O(n²) — 2000-line replays take 2–4 s to render
### Summary

Opening a previous firmware build's log (or any other historical job output via `firmware/follow_job`) takes 2–4 seconds to render on a typical 1500–2000-line log because the frontend does O(n²) work replaying historical lines. The slowness is purely a frontend pattern; the backend snapshot/replay is fine.

### Reproduction

1. Run a compile or install job that produces ~1500+ output lines (any non-trivial ESP-IDF build hits this trivially).
2. Close the dialog after it completes.
3. Reopen it from the firmware jobs list.
4. Observe: dialog opens immediately, then the log area is blank/sluggish for 2–4 seconds before all historical lines pop in.

### Root cause

The backend (`device-builder/.../firmware/controller.py:418-422` in `_send_initial`) replays each historical line as a separate `output` WebSocket event:

```python
async def _send_initial(controls: StreamControls) -> None:
    for line in snapshot:
        await client.send_event(message_id, StreamEvent.OUTPUT, line)
```

The frontend handler in `src/components/command-dialog.ts:992` (and `:1099` for the alt code path) reacts to each event by copying the entire accumulated array:

```typescript
this._streamId = this._api.firmwareFollowJob(jobId, {
  output: (line) => {
    this._lines = [...this._lines, line];   // ← new array every line
  },
  ...
});
```

For an `n`-line replay this is `1 + 2 + ... + n = O(n²)` element copies — about **2 million copies for a 2000-line log** — just to assemble `_lines`.

Every assignment is a reactive write, so Lit fires `render()` on `esphome-ansi-log` 2000 separate times. Each `render()` in `src/components/ansi-log.ts:316`:

- Runs `_chunksToVisualLines(this.lines)` over the *whole* array (O(n)).
- Calls `_renderLine(line, state)` for every line, re-parsing ANSI escapes from scratch (O(n)) — the ANSI state has to be threaded sequentially, so Lit can't skip lines without help.
- Diffs the resulting child list against the previous DOM (O(n)).

Combined: roughly O(n²) work driven by stream cadence. For n=2000 that's a few million regex executions plus full DOM diffs on every frame, which is exactly the observed 2–4 s "thinking" pause.

### Suggested fix

Two complementary changes; either alone helps materially, both together effectively eliminate the delay.

#### A. Batch line appends within an animation frame (biggest single win)

In `src/components/command-dialog.ts`, coalesce rapid `output` callbacks so we don't trigger a render per line:

```typescript
private _pendingLines: string[] = [];
private _flushScheduled = false;

private _enqueueLine(line: string): void {
  this._pendingLines.push(line);
  if (this._flushScheduled) return;
  this._flushScheduled = true;
  requestAnimationFrame(() => {
    if (this._pendingLines.length > 0) {
      this._lines = [...this._lines, ...this._pendingLines];
      this._pendingLines = [];
    }
    this._flushScheduled = false;
  });
}
```

Replace both call sites:

```typescript
output: (line) => {
  this._lines = [...this._lines, line];
}
```

with:

```typescript
output: (line) => this._enqueueLine(line);
```

This alone collapses 2000 renders into roughly 60 (one per frame at 60 fps) and the array-copy work becomes ~O(n) total. In practice this takes the visible delay from 2–4 s to under 100 ms.

Also remember to flush the buffer on dialog close / stream end so the last partial batch isn't lost.

#### B. Memoize `_chunksToVisualLines` and the per-line ANSI parse in `ansi-log.ts`

The visual-lines computation only changes when `this.lines` does, and parsing the same line with the same entry-state always yields the same spans. Two caches:

- `WeakMap<readonly string[], string[]>` keyed by the `lines` array reference for the chunk-fold result.
- `Map<string, AnsiSpan[]>` keyed by `${entryState.color}|${entryState.bold}|${...}|${line}` for the per-line parse.

With B, each `render()` only does work proportional to the *new* lines, not the whole buffer — useful even after the batching in (A) for very long buffered logs (2000 retained lines × 60 renders per second during a live build is still a lot).

### Why A is the priority

Most of the wall time is the DOM diffing inside Lit, not the regex parsing. Dropping ~30× redundant render frames per second is the biggest single lever. (B) optimizes the per-render cost but if (A) already takes the render count from 2000 to 60, the per-render cost matters far less.

### Backend-side alternative (not recommended)

The backend could send the historical snapshot as a single bulk event instead of one event per line. That would also fix the slowness, but:

- It changes the wire protocol (less attractive — other clients may exist).
- It doesn't help live streaming when a tool emits a burst of output (pip install progress, esptool block writes) — that path still benefits from frontend batching.
- The frontend `_enqueueLine` change above gracefully handles *both* historical replay and bursty live streams with one code path.

So fixing it on the frontend is the right call.

swoboda1337:

Replaying a 2000-line historical build log used to do
`host._lines = [...host._lines, line]` per chunk — 1 + 2 + … + n
array copies (~2M for n=2000) and one ansi-log render per write.
Dialog-open delay was a visible 2–4 s pause before the log painted.

Coalesce streamed lines through `enqueueLine` / `flushPendingLines`:
each `onOutput` callback pushes into a pending buffer and schedules
a single `requestAnimationFrame` flush. n=2000 collapses to ~60
renders (one per frame at 60 fps) and the array-copy work becomes
O(n) total. Flush is synchronous at stream-terminal callbacks
(`onResult` / `onError`), `detachStream`, and `_downloadOutput`,
so consumers never race the rAF. Reset on every `_lines = []`
site (`open`, `followJob`, `startCommand`, secrets-toggle restart)
cancels the pending frame so a stale batch can't paint onto the
fresh buffer.

The same path applies to both `validate` and `follow_job` streams,
so live bursty output (pip install, esptool block writes) benefits
too — not just the historical-replay burst at attach time.

### ansi-log auto-scroll fix (second commit)

Surfaced by the batching change: the pre-existing auto-scroll
path in `ansi-log.ts` did `scrollTop = scrollHeight` inside its
own `requestAnimationFrame`, deferring the scroll by one frame
after Lit committed the new DOM. With one line per render that
one-line gap was invisible. With many lines per render (the new
common case during chatty builds), the scroll lagged far enough
behind the new content height that the last line in each burst
was visibly clipped below the viewport — only catching up when
the stream slowed down.

Move the auto-scroll path to a synchronous `_syncScrollToBottom`
called directly from `updated()`. `updated()` runs after Lit's
DOM commit, so reading `scrollHeight` there forces a layout that
already accounts for the appended lines. The public
`scrollToBottom` keeps its rAF wrapper for the dialog-open path,
where the post-open animation needs a frame to settle before the
scroll target is meaningful.

### Verified locally

- Previously-completed compile log: dialog-open delay went from a
  visible 2–3 s pause to instant paint on the same buffer.
- Live ESP-IDF compile burst (~30 "Compiling …" lines per frame
  during the mbedtls stage): bottom line tracks the burst cleanly
  instead of getting clipped until the stream slows.

- fixes #348

Pairs naturally with the storage-side CR-collapse landing in
esphome/device-builder#903 — together, a chatty ninja build's
~2000 stored chunks become ~300–500 real lines, replayed in
~10 render frames instead of 2000.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes (1112 existing tests; no regressions).

Note on tests: rAF coalescence is awkward to assert cleanly without
a vitest fake-timer harness for `requestAnimationFrame`. Happy to
add one if you'd like — the natural shape is to drive a sequence
of `enqueueLine` calls, advance the frame, and assert `_lines`
contains the full batch in one write. Left out of the prototype to
keep the diff focused on the behaviour change.

### DIFF
--- a/src/components/ansi-log.ts
+++ b/src/components/ansi-log.ts
@@ -347,7 +347,15 @@ export class ESPHomeAnsiLog extends LitElement {
 
   protected updated(changedProperties: Map<string, unknown>) {
     if (changedProperties.has("lines") && this.autoScroll && !this._isUserScrolled) {
-      this._scrollToBottom();
+      // Sync scroll: ``updated`` runs after Lit has committed the
+      // DOM mutation, so reading ``scrollHeight`` here forces a
+      // layout that reflects the just-appended lines. Deferring
+      // to rAF (as the public ``scrollToBottom`` does for the
+      // dialog-open path) lags by one frame, which is invisible
+      // when one line lands per render but produces a visible
+      // "bottom line clipped" cliff during the rAF-batched
+      // streaming bursts now coming from command-dialog.
+      this._syncScrollToBottom();
     }
   }
 
@@ -420,13 +428,14 @@ export class ESPHomeAnsiLog extends LitElement {
     this._isUserScrolled = scrollHeight - scrollTop - clientHeight > 40;
   }
 
+  private _syncScrollToBottom() {
+    if (!this._container) return;
+    this._ignoreNextScroll = true;
+    this._container.scrollTop = this._container.scrollHeight;
+  }
+
   private _scrollToBottom() {
-    requestAnimationFrame(() => {
-      if (this._container) {
-        this._ignoreNextScroll = true;
-        this._container.scrollTop = this._container.scrollHeight;
-      }
-    });
+    requestAnimationFrame(() => this._syncScrollToBottom());
   }
 
   /** Public method to scroll to bottom programmatically. */
--- a/src/components/ansi-log.ts
+++ b/src/components/ansi-log.ts
@@ -347,14 +347,8 @@ export class ESPHomeAnsiLog extends LitElement {
 
   protected updated(changedProperties: Map<string, unknown>) {
     if (changedProperties.has("lines") && this.autoScroll && !this._isUserScrolled) {
-      // Sync scroll: ``updated`` runs after Lit has committed the
-      // DOM mutation, so reading ``scrollHeight`` here forces a
-      // layout that reflects the just-appended lines. Deferring
-      // to rAF (as the public ``scrollToBottom`` does for the
-      // dialog-open path) lags by one frame, which is invisible
-      // when one line lands per render but produces a visible
-      // "bottom line clipped" cliff during the rAF-batched
-      // streaming bursts now coming from command-dialog.
+      // Sync (not rAF-deferred): ``updated`` runs post-DOM-commit,
+      // and a one-frame lag clips the bottom line during bursts.
       this._syncScrollToBottom();
     }
   }
--- a/src/components/command-dialog.ts
+++ b/src/components/command-dialog.ts
@@ -210,7 +210,7 @@ export class ESPHomeCommandDialog extends LitElement {
     this._port = options?.port ?? "OTA";
     this._state = null;
     this._lines = [];
-    this.resetPendingLines();
+    this._resetPendingLines();
     this._statusMessage = "";
     this._jobId = "";
     this._jobStatus = null;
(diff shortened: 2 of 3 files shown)


stvncode (approved):
LGTM

PR #356, opened 2026-05-19
yaml-editor: stop active-line bg from hiding the selection

### LINKED ISSUE #326: [Bug] Selected string is not highlighted
### What happened?

When you doubleclick on a string in the editor it gets selected, but it is not visually highlighted as selected. Such highlighting works only when selecting multiple lines.

Can be reproduced by selecting a word that is found on the page multiple times - all other words are highlighted (class `.cm-selectionMatch`), just the main one stays without highlight. It is not a theme issue - light & dark have the same problem.

Image

Image

### Dashboard version

0.1.0b46

### How did you install it?

From source (script/setup)

### Where is it running?

Local — Linux

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

swoboda1337:

`.cm-activeLine` is rendered as a line decoration on the
`.cm-line` div inside `.cm-content`, while the selection layer
(`.cm-selectionLayer`) sits *behind* the content. An opaque
`backgroundColor` on `.cm-activeLine` therefore completely covers
the selection rectangle on the active line — which produced the
symptom in #326: double-clicking a word inside the active line
showed the `.cm-selectionMatch` highlight on every *other*
matching word but left the selected word itself looking
unhighlighted (the search extension explicitly skips the
overlapping main match, so the active line was relying entirely
on the now-hidden selection background).

Switch the `.cm-activeLine` background to a translucent
white-on-dark / black-on-light overlay so the selection layer
beneath shows through. The gutter highlight
(`.cm-activeLineGutter`) keeps its opaque colour since nothing
sits behind it. CodeMirror's own base theme uses the same
translucent approach (`#99eeff33` dark / `#cceeff44` light) for
exactly this reason.

Tint is matched to the previous opaque value's perceived contrast
against the editor background, so the active-line affordance
reads the same as before in the absence of a selection.

### Verified locally

- Double-click on a word inside the active line: selected text
  now shows the VSCode-blue selection rectangle; identical words
  on other lines still get the green `.cm-selectionMatch` tint.
- Active-line affordance still reads as a subtle row highlight
  in both dark and light themes when no selection is present.

- fixes #326

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes (1112 existing tests; no regressions).

### DIFF
--- a/src/util/yaml-editor-theme.ts
+++ b/src/util/yaml-editor-theme.ts
@@ -20,6 +20,9 @@ const DARK_FG = "#d4d4d4";
 const DARK_GUTTER_FG = "#858585";
 const DARK_SELECTION = "#264f78";
 const DARK_ACTIVE_LINE = "#2a2d2e";
+// Translucent variant for the content-layer overlay; opaque
+// active-line bg hides the selection layer behind it (#326).
+const DARK_ACTIVE_LINE_TINT = "#ffffff0a";
 const DARK_CURSOR = "#aeafad";
 
 const LIGHT_BG = "#ffffff";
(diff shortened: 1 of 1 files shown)


stvncode (approved):
LGTM

Issue #41, opened 2026-05-01
Implement mobile view


marcelveldt:
Currently only a desktop view has been implemented and only some very rough initial mobile layout.

- [x] Nearly all dialogs do not align to middle of vertical space on mobile
- [x] Download window renders in the middle of the page on mobile and overflows
- [x] Validate  window renders in the middle of the page on mobile and overflows
- [x] Download window overflows when build status expands
- [x] Kebab menu on cards opens below the page

<img width="401" height="331" alt="Image" src="https://github.com/user-attachments/assets/7adee6ae-c75f-467f-badb-8d2bb8cabcdd" />

jesserockz:
Multiple select footer bar is broken on mobile - too wide

jesserockz:
New device dialog shows two columns of "other boards", should be one single column

swoboda1337:
The table view (esphome-device-table) doesn't fit on mobile viewports. Currently 7 columns are default-visible (select, status, name, address, platform, config, actions) and the only responsive behaviour at narrow widths is a horizontal scroll bar on .table-scroll. Phones get a tiny, scroll-only view of a table designed for desktop widths.

bdraco:
All the major points are done. Separate issues opened for the remaining issues

PR #401, opened 2026-05-28
Cascade table column hiding below 920/820/700px breakpoints

bdraco:

The dashboard table shows 7 default-visible columns (select, status, name, address, platform, config, actions). The existing breakpoint ladder in `table-cell-styles.ts` only hid per-row action buttons; whole columns kept their slots and the table horizontal-scrolled on tablet portrait / phone-landscape widths. Extend the ladder so whole columns drop in the same priority order the action-button ladder uses: below 920px the `config` column hides, below 820px also `platform`, below 700px also `address`. `select / status / name / actions` stay as the four-column spine at every width. Mobile users still default to the card grid; this only affects users who toggle to table view on a narrow viewport.

- refs https://github.com/esphome/device-builder-frontend/issues/41 (umbrella mobile-view tracker; this addresses @swoboda1337's "table view doesn't fit on mobile viewports" comment)

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/dashboard/table-cell-styles.ts
+++ b/src/components/dashboard/table-cell-styles.ts
@@ -253,9 +253,32 @@ export const tableCellStyles = css`
     .cell-action-btn--logs {
       display: none;
     }
+    /* Progressive column hiding for the table-view fallback on
+       narrow viewports (mobile users default to the card grid; this
+       only kicks in when they toggle to table view on a small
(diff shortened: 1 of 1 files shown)


bdraco:
changed my mind on this. If they asked for they want it.. wrong approach.. rethinking

PR #407, opened 2026-05-28
Pair Select-multiple with the device-count in card view

bdraco:

Select-multiple was sitting in the toolbar-row between the view-switcher and the facet chips (PR-F). Better fit: pair it with the device-count on its own row — both reference the device list ("operate on these N devices"), and grouping them visually frees the toolbar-row above for filter-related controls only.

Adds a new `.device-count-row` flex container that puts the count on the left and the toggle on the right via `justify-content: space-between`. Table view (`render-content.ts`) keeps PR-F's placement because the table doesn't have an external device-count in its toolbar.

- refs https://github.com/esphome/device-builder-frontend/issues/41 (umbrella mobile-view tracker)

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/dashboard/render-content.ts
+++ b/src/components/dashboard/render-content.ts
@@ -4,10 +4,10 @@ import type { AdoptableDevice, ConfiguredDevice } from "../../api/types.js";
 import { downloadYaml, editDevice } from "./actions.js";
 import {
   renderAddDeviceCard,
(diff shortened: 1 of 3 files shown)


bdraco:
Closing; the layout cascade got out of hand. Will start fresh with a smaller-scope PR.

PR #428, opened 2026-05-28
Fix Visual Editor dropping values for map keys containing a dot

### LINKED ISSUE #1005: [Bug] Visual Editor doesn't keep values with dotted tag-specific log levels
### What happened?

When a tag-specific log level with a dotted tag, e.g. i2c.idf, is added in the Visual Editor, the value is lost after selection in the combobox, but it is set in the YAML. If set in the YAML editor, the value is not reflected in the Visual Editor.

Image

### Dashboard version

Beta66

### How did you install it?

PyPI

### Where is it running?

Local — Linux

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

In the Visual Editor, a logger `logs:` row keyed with a dot (for example `i2c.idf`) lost its level in the dropdown even though the value was written to YAML; a dotted key set in the YAML editor also did not show up in the Visual Editor.

The post-render `<wa-select>` value sync rebuilt a field's path from its `data-field-key` with `key.split(".")`, while the renderers built that attribute with `path.join(".")`. The round-trip over-segmented any path whose segment is a user-supplied map key that itself contains a dot, so the value lookup missed and the select was blanked. The fix encodes the path as JSON in `data-field-key` and decodes it back, with a dot-split fallback for the non-path pin-advanced toggle key.

- fixes https://github.com/esphome/device-builder/issues/1005

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/config-entry-form.ts
+++ b/src/components/device/config-entry-form.ts
@@ -30,6 +30,7 @@ import type { LocalizeFunc } from "../../common/localize.js";
 import { localizeContext } from "../../context/index.js";
 import { type ValidationError } from "../../util/config-validation.js";
 import { _isStructuralType, filterRenderable } from "./config-entry-render-filter.js";
+import { fieldKeyAttr, parseFieldKey } from "./config-entry-renderers-shared.js";
 import { getIn, isPrimitiveOrNullish } from "../../util/nested-values.js";
 import { registerMdiIcons } from "../../util/register-icons.js";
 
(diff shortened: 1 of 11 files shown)


bdraco:
Thanks, both addressed in 11bf06b.

1. Flipped the straggling `renderYamlOnlyFallbackIfNonPrimitive` site to `fieldKeyAttr(path)` so the helper is used uniformly across the module.

2. Added a renderer-level regression test (`config-entry-form-map-sync.test.ts`) that renders the `logger.logs` map for a row keyed `i2c.idf` through the real select renderer and asserts the emitted `data-field-key` decodes back to `["logs","i2c.idf"]`, with a companion assertion that it does not over-segment into `["logs","i2c","idf"]`; that catches a producer reverting to `join`, which the codec test alone would miss. A full `<wa-select>` form mount isn't possible under happy-dom (its form-associated base reads `ElementInternals.validity`), so the test walks the `TemplateResult` instead, matching the pattern the other renderer tests use.

PR #437, opened 2026-05-29
Stop the add-automation dialog from leaking wa-selects on every open

bdraco:

Opening the Add automation dialog leaked memory; each open rebuilt the whole form and recreated its wa-select dropdowns, and webawesome does not fully release those selects when they are removed, so the tab grew without bound. Now the spinner only shows before the first load; afterwards the form stays mounted across reopens, so the selects are created once and reused. Verified with a heap probe, opening the catalog dialogs repeatedly is now flat at zero added nodes and listeners.

- Reported on Discord, no issue filed.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/add-automation-dialog.ts
+++ b/src/components/device/add-automation-dialog.ts
@@ -260,7 +260,7 @@ export class ESPHomeAddAutomationDialog extends LitElement {
         })
       : this._localize("device.add_automation");
     return html`<wa-dialog light-dismiss label=${title}>
-      ${this._loading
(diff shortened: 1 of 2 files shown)


bdraco:
Replaced the source-scan test with a behavioral one; it mounts the dialog under happy-dom, asserts the spinner only shows before the first load, and asserts the same wa-select element survives a reopen instead of being recreated. Confirmed it fails on the pre-fix code and passes on the fix.

PR #439, opened 2026-05-29
Restore the OTA-format firmware download

### LINKED ISSUE #1033: [Bug] No OTA firmware download possible
### What happened?

When choosing Install / Download firmware binary in the new UI, I always ONLY get the full binary.

The previous option to download the OTA format binary is missing which prevents an OTA update of devices using HTTP update method.
See attached screenshot for previous option during download that allows the download of the OTA binary.

Image

### Dashboard version

0.1.0b71

### How did you install it?

ESPHome container (preview toggle)

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

The manual Download firmware binary flow always handed back the factory image and never offered the other formats, so the OTA binary that the legacy dashboard exposed (needed for HTTP method OTA updates) was unreachable. The backend already lists every format via firmware/get_binaries, the UI just wasn't surfacing them. Now, when a compiled device produces more than one format, the dialog shows a picker listing each one (factory, OTA, ...) with its title and description; single format devices and the Web Serial / web.esphome.io paths are unchanged. The picker downloads whatever the backend lists, so it works for ESP32, RP2040, LibreTiny, and nRF52.

- fixes https://github.com/esphome/device-builder/issues/1033

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/api/types.ts
+++ b/src/api/types.ts
@@ -1361,6 +1361,8 @@ export interface FirmwareJob {
 export interface FirmwareBinary {
   title: string;
   file: string;
+  // Optional subtext from ESPHome's get_download_types; not every platform supplies one.
+  description?: string;
 }
 
 export interface FirmwareDownload {
--- a/src/components/firmware-install-dialog.ts
+++ b/src/components/firmware-install-dialog.ts
@@ -12,7 +12,7 @@ import {
 import { LitElement, html } from "lit";
 import { customElement, state } from "lit/decorators.js";
 import type { ESPHomeAPI } from "../api/index.js";
-import { JobSource, type ConfiguredDevice } from "../api/types.js";
+import { JobSource, type ConfiguredDevice, type FirmwareBinary } from "../api/types.js";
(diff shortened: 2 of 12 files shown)


bdraco:
Thanks. Addressed all three suggestions in the amended commit:

1. Typed the binding explicitly: `let binaries: FirmwareBinary[]`.
2. Softened `choose_binary_desc` to a format-agnostic line in all five locales, since each row already shows its own backend-supplied description.
3. Added tests for the get-binaries error path and the re-pick (download another format).

bdraco:
Fixed the footer nit too. Added a dedicated `downloading` step so the byte fetch no longer reuses `queued`; `renderFooter` shows Close (not Stop) for both `choose-binary` and `downloading`, so the re-pick no longer flickers a dead Stop button, and the Stop affordance stays only for the actually cancelable compile/flash phases. Added a footer test pinning Close-not-Stop on those steps and Stop while compiling.

PR #445, opened 2026-05-29
Confirm dialogs with Enter consistently

### LINKED ISSUE #1035: [Bug] Cannot confirm dialog boxes by pressing ENTER
### What happened?

Most dialog boxes can be confirmed by pressing ENTER, but ENTER has no effect in at least the following cases. In some cases, you can't use the TAB key to navigate to the button either.

Image

Image

Image

Image

### Dashboard version

v0.1.0b71

### How did you install it?

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

Several dialogs could not be confirmed with Enter; wa-dialog only handles Escape, and these dialogs use plain click buttons rather than forms, so Enter did nothing. Adds a shared EnterController (mirrors the existing EscapeController) that fires the dialog's primary action on a plain Enter, while ignoring modifiers, IME composition, already-handled events, and focus on elements that act on Enter themselves (button, link, textarea, select, contenteditable). It is wired into confirm-dialog, unsaved-changes-dialog, yaml-validation-dialog, onboarding-wifi-dialog, and edit-pairing-endpoint-dialog; accept-peer-dialog is fixed transitively since it wraps confirm-dialog. A destructive confirm does nothing on Enter so a stray keypress can't delete, and the yaml-invalid dialog jumps to the first error instead of force-saving. The two input dialogs also autofocus their first field so Enter works the moment they open. Tests are behavioral happy-dom mounts, not source scans.

- fixes https://github.com/esphome/device-builder/issues/1035

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/confirm-dialog.ts
+++ b/src/components/confirm-dialog.ts
@@ -112,8 +112,7 @@ export class ESPHomeConfirmDialog extends LitElement {
 
   private _decided = false;
 
-  // Enter confirms, but never a destructive prompt — a stray Enter must
-  // not delete. Cancel stays on Escape / the Cancel button.
+  // Enter confirms, but never a destructive prompt (a stray Enter must not delete).
   private _enter = new EnterController(this, () => {
     if (!this.destructive) this._confirm();
   });
--- a/src/components/confirm-dialog.ts
+++ b/src/components/confirm-dialog.ts
@@ -7,6 +7,7 @@ import { localizeContext } from "../context/index.js";
 import { dialogCloseButtonStyles } from "../styles/dialog-close-button.js";
 import { modalDialogStyles } from "../styles/modal-dialog.js";
 import { espHomeStyles } from "../styles/shared.js";
+import { EnterController } from "../util/enter-controller.js";
 import { registerMdiIcons } from "../util/register-icons.js";
 
 import "@home-assistant/webawesome/dist/components/dialog/dialog.js";
@@ -111,9 +112,16 @@ export class ESPHomeConfirmDialog extends LitElement {
 
   private _decided = false;
 
+  // Enter confirms, but never a destructive prompt — a stray Enter must
+  // not delete. Cancel stays on Escape / the Cancel button.
+  private _enter = new EnterController(this, () => {
+    if (!this.destructive) this._confirm();
+  });
+
   open() {
     this._decided = false;
     this._dialog.open = true;
+    this._enter.set(true);
   }
 
   close() {
@@ -174,6 +182,7 @@ export class ESPHomeConfirmDialog extends LitElement {
   }
 
   private _onAfterHide() {
+    this._enter.set(false);
     if (!this._decided) {
       this.dispatchEvent(new CustomEvent("cancel", { bubbles: true }));
     }
--- a/src/components/confirm-dialog.ts
+++ b/src/components/confirm-dialog.ts
@@ -163,6 +163,7 @@ export class ESPHomeConfirmDialog extends LitElement {
   }
 
   private _confirm() {
+    if (this._decided) return; // a repeated Enter must not dispatch twice
     this._decided = true;
     this.close();
     // composed:false (omitted) so wrappers like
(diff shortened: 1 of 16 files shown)


bdraco:
Addressed all three. EnterController now also ignores Shift+Enter, and documents the single-active-modal assumption (an onEnter can preventDefault to stop a co-active controller). onboarding-wifi now focuses the SSID field via updateComplete rather than the autofocus attribute, matching edit-pairing-endpoint, so both input dialogs reliably focus their first field on open; added a Shift+Enter case to the controller test. Also tightened the comments to the terse why.

bdraco:
Addressed both pre-merge items plus the test gap.

- onboarding _save() now bails on `if (this._saving) return;`, so the Enter path can't double-submit while a save is in flight (the concrete bug). Added a test that a second _save during the first does not re-issue setOnboardingWifi.
- Made the close-on-action dialogs idempotent on their existing resolved flags (confirm _decided, unsaved _resolved, yaml _resolvedExit), so a repeated Enter in the gap before wa-after-hide can't double-dispatch. Did the same for the wizard steps (_done / _finished). Kept the controller firing-while-active rather than one-shot so the wizard's name -> wifi -> finish Enter still works.
- Added double-Enter tests (confirm fires once, empty-config creates once).
- Verified in a real browser that a button-only confirm-dialog opens with focus on the native dialog element, not Cancel (path: esphome-confirm-dialog > wa-dialog > dialog), and a real Enter from there confirms; so the controller is not pre-empted by a focused self-handling button.

Also folded in two more spots from the same issue (wizard board-setup and create-empty-config steps) since the reporter wanted Enter there too.

bdraco:
Addressed both.

- Wired the co-active-modal defense instead of just documenting it: EnterController now calls preventDefault() when it acts, so a second co-active controller bails on its defaultPrevented guard and a single Enter acts once, matching EscapeController. Updated the comment so it no longer promises an undefended behaviour, and added a test with two controllers on one target asserting only the first fires.
- Extracted the duplicated pressEnter helper to test/_press-enter.ts and imported it in the three dialog/wizard tests.

bdraco:
Addressed both Copilot inline comments (wizard-step-setup and wizard-step-empty-config).

Good catch: the wizard steps stay mounted inside create-config-dialog's wa-dialog while it is only hidden (light-dismiss / Escape / close), so attaching the Enter listener in connectedCallback left a stray window listener that could dispatch next-step / finish-setup / create-empty-config from an Enter typed elsewhere on the page. Fixed by mirroring the edit-pairing-endpoint pattern: the parent now drives an `active` property from its open state (set true on open, false on wa-after-hide), and each step gates the EnterController via `willUpdate` (`set(this.active)`). So Enter is dead the moment the dialog is hidden, even though the step stays mounted. Added a test asserting Enter is ignored when `active` is false.

bdraco:
Both warnings are the wizard fire-while-hidden case, fixed in 33fdf48 (pushed after this review's last referenced commit 0aa3e44). create-config-dialog now tracks _open (true on open, false on wa-after-hide) and binds ?active onto the two Enter-wired steps; each step gates the controller via willUpdate (set(this.active)) instead of connectedCallback, mirroring edit-pairing-endpoint. So the listener drops the moment the dialog hides even though the step stays mounted, and a test asserts Enter is ignored when active is false. The next re-review should pick it up.

bdraco:
Addressed the latest round.

- Suggestion #2 (latches never reset / lockout on reopen + failed create): removed the _done/_finished step latches entirely. The parent create-config-dialog already de-dupes via _runCreate's synchronous _submitting guard, so the step-level latches were redundant for double-dispatch and harmful for retry (a failed create left the step mounted with the latch stuck, killing both the button and Enter). Now a retry after failure works and a reopen isn't dead. Added tests: create-config-dialog drops a second create while one is in flight and allows a retry after a failed create; the step test asserts it no longer latches.
- Suggestion #1 (co-active routing): documented that the single-active-modal assumption is load-bearing and "first" is window registration (open) order, not z-order, so Enter routing is undefined if two ever stack.
- Suggestion #3 (security-sensitive accept-peer): accept-peer-dialog already renders the confirm-dialog with `destructive`, so the controller's `if (!this.destructive)` already makes Enter a no-op there; the grant requires an explicit click. (The PR description's "non-destructive" note was wrong; the behaviour is the safe one.)

PR #450, opened 2026-05-30
Split api/types.ts into per-domain files

esphbot:

Split `src/api/types.ts` (2476 lines) into per-domain files under `src/api/types/`, and point every consumer at the per-domain file directly. `types.ts` was the largest file in the repo at roughly 2476 lines, about 4x the 500-600 line cap, and it mixed ~14 unrelated domains (WS protocol, devices, boards, components, config entries, automations, firmware jobs, remote build, and more) into a single file.

The new layout mirrors how the Home Assistant frontend organises `src/data`; one file per domain, consumers import the domain file directly, and there is no central barrel. Each domain's declarations move into its own file; cross-domain references use relative `import type` between siblings. The ~184 consumers now import from the specific per-domain file (for example `../api/types/devices.js`); `src/api/types.ts` is deleted, and `src/api/index.ts` keeps only its `ESPHomeAPI` and `APIError` re-exports. No behaviour change, no public API change, no translation changes; the largest resulting file is 436 lines, all under the cap. Brought up to date with main.

- N/A

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


bdraco:
It does need to split but we should drop the big file and update the imports. Nothing else uses this

bdraco (approved):

PR #461, opened 2026-05-30
Send the raw display name from the create wizard

### LINKED ISSUE #1070: [Bug] Device name not set as friendly_name
### What happened?

Follow-up for issue #926 

When I add a new device and use a descriptive name, any special characters are removed, and the name is used for both the name: and friendly_name: fields.

The fix in #926 seems to be incomplete. If I'm not mistaken, there's still a frontend-PR missing?

### Dashboard version

v0.1.0b74

### How did you install it?

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

The new-device wizard slugified the name before calling `devices/create`, so the backend received the slug for both the hostname and the friendly name; the descriptive name the user typed was stripped to the slug and `name == friendly_name`. This forwards the raw display name from both generated-config flows (empty config and basic setup). The backend (companion PR esphome/device-builder#1074) derives the hostname slug and keeps the cleaned original as `esphome.friendly_name`, so a name like `Living Room #2` lands as `name: living-room-2`, `friendly_name: Living Room #2`. The hostname slug is not lost; it just lives server-side now, and the dialog navigates via the backend-returned configuration. The `friendly-name-slugify` util is left in place but no longer called from the dialog; the upload path is untouched.

- fixes https://github.com/esphome/device-builder/issues/1070

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


bdraco:
Both addressed in 7032e1c.

1. Dead code: removed `src/util/friendly-name-slugify.ts` and its test; the backend owns slugging now (#931/#1074), so there's no remaining caller and nothing downstream per the lockstep policy. Also retired the stray `friendlyNameSlugify` mention in esphome-id.test.ts's comment.

2. Coverage: added a sibling test that emits `finish-setup` (board + WiFi + name) and asserts `createDevice` is called with the raw name and `config_type: basic`, so both generated-config flows are now pinned.

PR #467, opened 2026-05-30
Add an "All" option to the list view pagination

bdraco:

Adds an "All" choice to the table list view's rows-per-page selector, so the device list can show every device on one page; the default stays 25, this is purely an added option. "All" is the sentinel page size 0, which persists like any other choice through the existing table_page_size preference. The table can't hand TanStack a page size of 0 (the page count would be infinite and the slice empty), so it translates 0 into an effective page size of the row count at the boundary and hides the page navigation while "All" is active. Adds the pagination_all key to every locale. No backend change; table_page_size already accepts any integer.

### Note on large fleets

"All" renders every row to the DOM; there is no virtualization. That's the explicit, opt-in request, the default stays 25, and rows are rendered with a keyed `repeat` so re-renders reuse row DOM. Windowing for very large fleets is a deliberate, separate follow-up.

- implements https://github.com/orgs/esphome/discussions/3682

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/dashboard/device-table.ts
+++ b/src/components/dashboard/device-table.ts
@@ -49,8 +49,9 @@ import { tableLayoutStyles } from "./table-styles.js";
 
 import "@home-assistant/webawesome/dist/components/icon/icon.js";
 import "@home-assistant/webawesome/dist/components/spinner/spinner.js";
+import { ALL_PAGE_SIZE, effectiveTablePageSize } from "./pagination.js";
 import "./table-column-toggle.js";
-import { ALL_PAGE_SIZE } from "./table-pagination.js";
+import "./table-pagination.js";
 import "./table-row-menu.js";
 
 registerMdiIcons({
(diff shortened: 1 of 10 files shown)


bdraco:
Both addressed in c5cce1f.

1. Core logic now tested: extracted the sentinel translation into a pure `effectiveTablePageSize(pageSize, rowCount)` and added `device-table.test.ts` covering it directly (normal size passthrough; All expands to the row count; the `Math.max(..., 1)` floor so 0 never reaches TanStack on an empty dataset) plus a happy-dom mount asserting a normal size paginates (25 of 30) while All renders all 30 rows on one page with `page-size=0` on the control. On the filter-active point: the effective size is the unfiltered `_rows.length`, which is an upper bound over any filtered subset, so all matching rows still fit page 0. Also verified end-to-end in headless Chrome (30 seeded devices: All shows 30, hides nav, persists across reload).

2. Virtualization: added a decision note to the PR body; it's an explicit opt-in (default stays 25) and rows use a keyed `repeat`, with windowing left as a separate follow-up.

Also named the sentinel `ALL_PAGE_SIZE` and reused it across both files instead of a bare 0.

PR #468, opened 2026-05-30
Offer repeatable triggers in the add-automation wizard

### LINKED ISSUE #1080: [Bug] on_time automation missing from add automation wizard when at least one is already added
### What happened?

Follow-up to #1049 

With the previous changes allowing multiple on_time events, when at least one on_time automation is already defined, the add automations wizard do not list Time -> On Time as an available automation.

Image

### Dashboard version

Lates Beta

### How did you install it?

PyPI

### Where is it running?

Local — Linux

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

esphbot:

The add-automation wizard hid a component trigger once that instance already had an inline `on_*:` handler. Right for once-only triggers, wrong for the ones ESPHome accepts as a list (`time.on_time` and friends), where the user should be able to add a second entry from the picker instead of hand-editing YAML.

The catalog now ships a `repeatable` flag per trigger (companion backend PR), so the wizard reads data instead of a hardcoded id set:

- `_filteredTriggers` keeps a `t.repeatable` trigger offerable after one handler exists.
- `_buildLocation` builds an indexed `component_on` location for it (append index = existing entry count on the instance), mirroring the `interval` path; without the index the backend overwrites the existing block.
- `LIST_SHAPED_TRIGGER_IDS` is deleted. The trigger catalog is backend-generated and already carries derived flags like `is_device_level`, so `repeatable` joins them; any future list trigger is picked up on the next catalog sync with no frontend edit.

Testing: happy-dom tests pin that the picker still offers a repeatable trigger with one present, the location appends at the next index, and non-repeatable triggers stay un-indexed. `npm run lint` (tsc), `npm run test` (1599 pass), and prettier are clean.

- fixes https://github.com/esphome/device-builder/issues/1080
- Requires backend PR esphome/device-builder#1085 (ships the `repeatable` catalog flag); land together.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


bdraco:
@esphbot Rr

src/util/yaml-sections.ts
```diff
@@ -673,6 +673,16 @@ const _DASH_TRIGGER_ENTRY_KEY_RE = new RegExp(
...
 
+/** Catalog trigger ids ESPHome accepts as a repeatable *list* of
+ *  entries — each entry carries its own params (``time.on_time``'s
+ *  cron fields) under one ``on_*:`` key. These stay offerable in the
+ *  add-automation wizard even when one already exists, and the wizard
+ *  appends a new indexed ``component_on`` entry instead of replacing
+ *  the block. Mirrors the body-shape discriminator above
+ *  (``_TRIGGER_ENTRY_KEYS``) and the backend's per-entry
+ *  ``upsert_component_on_entry`` splice path. */
+export const LIST_SHAPED_TRIGGER_IDS: ReadonlySet<string> = new Set(["time.on_time"]);
```
bdraco: hand maintained is correct but not good here

esphbot: Agreed — the hardcoded set is correct behavior but it's a second hand-maintained source of "which triggers are list-shaped," duplicating the `_TRIGGER_ENTRY_KEYS` cron-field list a few lines up. They're conceptually paired but representationally independent (a catalog id here, YAML key names there), so a future repeatable trigger has to be added in both, in two different forms, with nothing keeping them in sync.

The clean fix that fits this repo's lockstep deployment: let the backend own it. `AutomationTrigger` (src/api/types/automations.ts) already carries per-trigger catalog metadata — `applies_to`, `is_device_level`, `config_entries`. Adding a `repeatable: boolean` (or `is_list`) field there in a coordinated backend PR lets `_filteredTriggers` exempt on `t.repeatable` directly, and `add-automation-dialog` already has the live `this._available.triggers`, so it drops `LIST_SHAPED_TRIGGER_IDS` entirely — no frontend-side list to drift. The backend is already the source of truth for the catalog shape, so this is the natural home for the flag.

Would you prefer I keep this PR as the minimal frontend fix and open the backend `repeatable`-flag change as a follow-up coordinated PR (then convert this set to read the flag), or fold both into a single coordinated frontend+backend pair before merging? Happy to go either way.

esphbot: Addressed in the latest revision. The hand-maintained `LIST_SHAPED_TRIGGER_IDS` set is gone — it's no longer anywhere in the tree. Instead, `AutomationTrigger` (src/api/types/automations.ts) now carries a backend-generated `repeatable: boolean` catalog flag, alongside the existing derived metadata (`is_device_level`, `applies_to`, `config_entries`). `_filteredTriggers` exempts on `t.repeatable` directly and `_buildLocation` reads the same flag off the live `this._available.triggers`, so there's no frontend-side list to drift against `_TRIGGER_ENTRY_KEYS` anymore. The backend already owns catalog shape, so this is the natural home for the flag, and it fits the lockstep deployment model — the companion backend PR (esphome/device-builder#1085) ships the flag and lands together with this one. Any future list-shaped component trigger is now picked up on the next catalog sync with no frontend edit.

bdraco:
needs some more rework. doesn't handle a 2nd one well or comments

bdraco:
merge bug is not a regression on this pr. its an existing bug so merging this and fixing in a new pr

PR #485, opened 2026-05-31
Render globals as a repeatable list in the section editor

### LINKED ISSUE #1097: [Bug] Visual editor does not support list of items in globals
### What happened?

The Visual Editor does not support multiple items in globals.

Image

### Dashboard version

Latest Beta

### How did you install it?

PyPI

### Where is it running?

Local — Linux

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

The Visual Editor lost all but one entry when a device had several globals. In ESPHome globals is a list of typed variables, but the section form treated it as a single mapping, so a multi entry block parsed back as empty and saving overwrote every variable.

This renders globals as a repeatable list, reusing the same nested list control that esphome.devices and esphome.areas already use; you can add, edit, and remove each variable in the form and every entry round trips to YAML. The change is gated to globals, so other sections behave exactly as before.

- fixes https://github.com/esphome/device-builder/issues/1097

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/util/section-entry-overrides.ts
+++ b/src/util/section-entry-overrides.ts
@@ -38,6 +38,17 @@ import { makeConfigEntry } from "./config-entry-defaults.js";
  *  the YAML pane. */
 export const MAP_SECTIONS: ReadonlySet<string> = new Set(["substitutions"]);
 
+/** Values-record key the list body is stashed under for a LIST_SECTIONS
+ *  section. Synthetic, not a real YAML key; parse and serialize strip
+ *  it so it never lands in YAML. */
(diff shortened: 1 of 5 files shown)


bdraco:
Closing this. Routing globals through a frontend section override is whack-a-mole; the root cause is that the component catalog does not describe globals as a list, so the section editor cannot know its shape. Reopening the fix at the source rather than special-casing each list-bodied section in the frontend.

PR #489, opened 2026-05-31
Render is_list sections as a repeatable list in the editor

### LINKED ISSUE #1097: [Bug] Visual editor does not support list of items in globals

bdraco:

The Visual Editor lost entries when a `globals:` block had more than one variable. It parsed the list body into an empty record, so the form showed a single item and a save dropped the rest (issue esphome/device-builder#1097).

The backend now flags list-bodied components like globals with a component-level `is_list` (companion PR esphome/device-builder#1108). This reads that flag and renders any `is_list` section as a repeatable keyed list: the flat catalog fields are wrapped in one keyed `nested` + `multi_value` entry, so every item round-trips through the same machinery as `esphome.areas`. Complex bodies that cannot be modelled (font glyphs, lambdas) fall back to the YAML-only notice instead of risking a clobber on save.

This is the frontend half and depends on the catalog flag from the backend PR.

- fixes esphome/device-builder#1097

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


bdraco:
The quoting fixes need to be another PR

PR #513, opened 2026-06-01
Unify the compile / install / logs dialogs on a shared process-terminal base

### LINKED ISSUE #346: Make compiling / installing / logs dialog more consistent
Right now depending on your install method: OTA, WebSerial, Direct Serial; The dialog is different.

- install wireless shows the ota progress
- install using computer running dashboard shows esptool and progress etc
- install using browser serial doesnt show any esptool logs

This should be represented consistently across all install methods.

bdraco:

The compile, install, and logs dialogs were each built on a different base; OTA and server-serial install streamed their esptool output through the command dialog, browser WebSerial install showed none, and the three dialogs each re-implemented the status banner, reset suggestion, toolbar, and log chrome.

This adds a shared `<esphome-process-terminal>` surface (log output, status banner or status card, progress bar, streaming dot, toolbar) on top of `esphome-base-dialog`, and migrates all three dialogs onto it as thin drivers; the command dialog also moves off the raw `wa-dialog`. While here, the browser WebSerial install now streams esptool chip-detect and flash output into the log the same way the backend-driven methods do, so install is consistent across OTA, server-serial, and WebSerial.

Existing flows are preserved (queued overlay, remote-builder sub-line, binary picker, web-download instructions, post-install back-to-install hand-off); the only behavior change is the newly visible WebSerial esptool logs.

- fixes https://github.com/esphome/device-builder-frontend/issues/346

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


bdraco:
Thanks for the review; all three points addressed in the latest push:

1. Firmware install close-animation race: added a `@request-close` handler that flips `_open` to false immediately, with teardown staying in `after-hide` (matches the logs and command dialogs).
2. Logs expand/collapse button accessible name: added `title` + `aria-label` driven by new `logs_expand` / `logs_collapse` keys (en, fr, nl).
3. Unused `toolbar-button` helpers: now adopted in the command and logs toolbars (they replace the inline term-btn markup and command-dialog's local toggle helper), so the module is no longer dead code.

Also unrelated: scrubbed an accidentally committed `node_modules` symlink out of the branch history.

bdraco:
Latest review round addressed:

- Toolbar layout (command + logs): the right-side controls were slotted inside a single wrapper div, collapsing them into one flex item; the wrapper is now `display: contents` so each button promotes to a direct flex item of the terminal toolbar and picks up its gap + per-control wrapping.
- i18n completeness: `logs_expand` / `logs_collapse` (and the new `logs_download`) are now in all five locales (en, fr, nl, hu, zh-CN).
- Logs download button: now has a `title` + `aria-label` (`logs_download`).

From the prior round (already in the branch): firmware install `@request-close` race guard, expand button accessible name, and the term-button helpers are now actually used in the command + logs toolbars. Also did a self-review pass: made the install `cardState` switch exhaustive over `InstallStep`, dropped the empty slotted wrapper divs (slot now sits on the projected content), and added tests for the term-button helpers and the close-race guards.

1721 unit tests pass, `tsc` + production build clean.

bdraco:
Latest review round addressed:

- `renderTermButton` now requires at least one of `label` / `title` (union type), so an icon-only button can't render without an accessible name; `renderTermToggle`'s `title` is likewise required. tsc confirms every call site already supplies one.
- Added `type="button"` to the term buttons (and the logs expand button) to avoid accidental form submission if a toolbar is ever rendered inside a form.
- Marked the decorative streaming dot `aria-hidden="true"`.

tsc + build clean, 1721 tests pass.

PR #523, opened 2026-06-01
Color ESPHome multi-line log continuations in WebSerial logs

bdraco:

WebSerial logs are read straight off the device UART and split on newlines. A multi-line ESPHome log record (a `dump_config` `[C]` block, say) opens its color once on the entry line and only resets on the last; the continuation lines are merely indented, with no header or color of their own. Splitting that on `\n` left every continuation line colorless and prefix-less, so config dumps rendered half-uncolored.

Backend-streamed logs don't hit this: the `esphome logs` CLI already re-applies the prefix + color to each line via aioesphomeapi's `LogParser`. WebSerial bypasses the backend entirely, so it had no equivalent.

This adds `ESPHomeLogParser`, a small TypeScript port of that `LogParser`: a stateful per-line parser that records the entry line's `[…]:` prefix + leading color and re-applies them (with a trailing reset) to continuation lines, so the output matches the backend's per-line shape. Every raw serial line runs through it before timestamping in `streamSerialToDialog`. It is a single shared library so any future client-side raw-log source can reuse it.

- Follow-up to the dialog-unification work in #513 (continuation-line coloring on WebSerial logs).

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/dashboard/actions.ts
+++ b/src/components/dashboard/actions.ts
@@ -5,6 +5,7 @@ import type { ConfiguredDevice } from "../../api/types/devices.js";
 import type { ArchivedDevice, BulkActionResult } from "../../api/types/system.js";
 import type { LocalizeFunc } from "../../common/localize.js";
 import { withBase } from "../../util/base-path.js";
+import { ESPHomeLogParser } from "../../util/esphome-log-parser.js";
 import {
   connectToPort,
   detectChip,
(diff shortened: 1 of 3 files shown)


bdraco:
Thanks for the review — both suggestions addressed:

1. **Empty line clearing carried color.** Fixed: blank lines (empty *or* whitespace-only) are now handled first as pass-throughs that don't touch the carried prefix/color, so a bare line (or a `\n\n` that splits to `""`) mid-block no longer drops coloring for the rest of the block. Added a test asserting color carries across an empty line.
2. **Continuation heuristic divergence.** Added a note at the detection site explaining that keying off leading whitespace (rather than aioesphomeapi's entry-regex non-match) is deliberate: the raw UART stream interleaves non-ESPHome output (esptool / PlatformIO) at column 0, which must pass through as plain lines instead of being re-colored as continuations.

Parser tests now 9/9; lint + build clean.

PR #528, opened 2026-06-01
Offer the OTA or serial choice when opening device logs

### LINKED ISSUE #525: There is no longer a way to get to serial logs except through install
The old builder would let you click logs and if web serial was available and there was at least one serial port candiate it would ask if you want ota or serial and which port

bdraco:

Clicking Logs on an online device jumped straight to wireless logs, so the only way to reach serial logs was to take the device offline or go through install. This shows the existing install method picker in logs mode whenever a serial path is available; browser WebSerial, or serial ports on the server. When neither exists, it opens OTA logs directly as before. This restores the old dashboard behavior where Logs asked whether you wanted wireless or serial.

The picker, its logs mode, and all three logs paths (OTA, server serial, WebSerial) already existed; the only change is the decision about when to show it.

- fixes https://github.com/esphome/device-builder-frontend/issues/525

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/dashboard/install.ts
+++ b/src/components/dashboard/install.ts
@@ -68,7 +68,9 @@ export async function openLogs(
   // Offer the OTA-vs-serial choice whenever a serial path exists — browser
   // WebSerial, or serial ports on the server (mirrors the old dashboard's
   // logs-target behavior; see #525). With no serial option at all, skip the
-  // picker and open OTA logs directly.
+  // picker and open OTA logs directly. The device's online/offline state is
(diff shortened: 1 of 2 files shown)


bdraco:
Both addressed in 42263e9; added a comment that the routing is intentionally state-agnostic (the picker gates its own OTA row on device state), and replaced the silent catch with a console.warn so a real getSerialPorts fault is diagnosable while still falling through to OTA.

PR #530, opened 2026-06-01
Show the log source in the logs dialog header

bdraco:

When you have several log windows open it's easy to lose track of which one is streaming over OTA and which is over serial. This adds the transport as a small monospace chip beside the dialog title, so each window self-identifies: OTA, the serial device path (e.g. /dev/cu.usbserial-110), or Web Serial.

The value comes straight from the dialog's existing state; _port is already "OTA" or the serial path for every open() path, and passive sessions are browser Web Serial, so no callers change. base-dialog gains a generic header-suffix slot (additive; other dialogs slot nothing there and look the same), and logs-dialog supplies the chip. One new localization key, logs_source_web_serial, added to every locale; "Web Serial" is kept verbatim in each since it is a proper API name already used that way across the existing strings.

- N/A

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/base-dialog.ts
+++ b/src/components/base-dialog.ts
@@ -82,17 +82,12 @@ import { centeredMobileDialog } from "../styles/dialog-mobile.js";
  *   plain ``<div class="actions">`` at the end of the
  *   body, and forcing them to migrate to a slotted footer
  *   would balloon the diff for no behaviour change.
- * - ``header-suffix`` slot: inline content rendered right
- *   after the title text in the header (e.g. a status /
- *   source chip). Empty for the common case, so existing
- *   dialogs are visually unchanged; the consumer owns the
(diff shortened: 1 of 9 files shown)


bdraco:
Addressed suggestion #1 (long serial paths overflowing the header on narrow widths) in 25e074c. base-dialog now exposes its header row and title text as the `label-row` / `title-text` parts; logs-dialog flexes that row and ellipsises the device name first, so the source chip stays whole. The chip also gets a last-resort `max-width` clamp plus a `title` tooltip so an extreme path on a tiny screen truncates inside the chip instead of overflowing, with the full value still on hover. Scoped to the logs dialog via those parts, so other dialogs keep their default wrapping title.

Issue #39, opened 2026-05-01
Various styling and alignment issues


marcelveldt:
We need to do a full sweep in styling, alignment etc.
There several issues like elements not aligned properly, style mismatches etc.
Also some places have a lot of padding which maybe can be improved.
Also fix the colors, they don't match the rest of HA, too blue.

tanishqmanuja:
<img width="594" height="163" alt="Image" src="https://github.com/user-attachments/assets/1cc46b13-a0bc-43b8-892b-cf5f6e9221ff" />

swoboda1337:
Also fix the colors, they don't match the rest of HA, too blue.

swoboda1337:
Can see the background color etc do not match HA:
<img width="1056" height="186" alt="Image" src="https://github.com/user-attachments/assets/fb130907-6613-4509-84cc-7cdb91d1aeac" />
<img width="811" height="344" alt="Image" src="https://github.com/user-attachments/assets/0d676574-86a3-477e-bd04-e197766a2c3c" />

bdraco:
All known issues have been addressed

PR #537, opened 2026-06-01
Flow the dashboard on mobile so the table view footer isn't clipped

bdraco:

On mobile, the table view kept the desktop fixed height, internal scroll layout even though the table reflows into stacked cards there; that left dead space above the pagination row and let the fixed footer clip the bottom card.

This makes the dashboard flow in the app's scroll container on phones, the same way the card and YAML views already do, so the page sizes to its content and reserves space for the footer. Desktop is unchanged.

- part of #39

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


bdraco:
wrong solution

PR #540, opened 2026-06-01
Migrate the confirm dialog onto esphome-base-dialog

bdraco:

Part of #39, the dialog consistency cleanup. The confirm dialog used a raw wa-dialog directly while settings already uses the shared esphome-base-dialog, so it skipped the wrapper's busy gate, close-button styling, and event-scope handling.

This moves it onto esphome-base-dialog; the imperative _dialog.open toggle becomes a reactive _open flag wired to the wrapper's request-close and after-hide contract, the public open() and close() API and the confirm, secondary, cancel events are unchanged. The shared modalDialogStyles header, title, body, footer rules now carry a dual selector so one fragment styles both the raw and the wrapped dialogs as the rest migrate. Verified with headless Chrome that the rendered dialog is pixel identical to main; dialog size, header and body padding, title, icon, text, and the Cancel and Confirm buttons all match.

- addresses https://github.com/esphome/device-builder-frontend/issues/39 (dialog consistency)

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/confirm-dialog.ts
+++ b/src/components/confirm-dialog.ts
@@ -1,17 +1,16 @@
 import { consume } from "@lit/context";
 import { mdiAlertOutline } from "@mdi/js";
 import { LitElement, css, html, nothing } from "lit";
-import { customElement, property, query, state } from "lit/decorators.js";
+import { customElement, property, state } from "lit/decorators.js";
 import type { LocalizeFunc } from "../common/localize.js";
 import { localizeContext } from "../context/index.js";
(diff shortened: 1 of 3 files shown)


bdraco:
Addressed the suggestion: added three tests in confirm-dialog.test.ts covering the new reactive paths the migration introduced.

- dismissing without a decision (after-hide on the inner esphome-base-dialog) fires a single cancel
- confirming first (Enter) fires confirm and does not also fire cancel
- request-close flips the reactive open flag to false

Full suite stays green; the Tests box is now ticked.

PR #546, opened 2026-06-02
Migrate the onboarding Wi-Fi dialog onto esphome-base-dialog

bdraco:

Migrates the onboarding Wi-Fi wizard, the last of the four raw `wa-dialog` surfaces, onto the shared `esphome-base-dialog` wrapper so its open/close wiring, busy gate, and close-button styling all come from one place.

This wizard is the only one that pins its actions row with `slot="footer"`, so `esphome-base-dialog` now forwards a footer slot through to the inner `wa-dialog`; dialogs that hide the footer are unaffected. Imperative `dialog.open` becomes a reactive `_open` flag with an `?open` binding, and the save-in-flight veto moves off a bespoke `request-close` `preventDefault` onto base-dialog's `?busy` gate.

Verified pixel-identical before/after via headless Chrome (CDP); content stays at the same gutter and the footer buttons render unchanged. Typing into the SSID field keeps focus and every keystroke, the same input node is reused across re-renders.

- part of #39

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/base-dialog.ts
+++ b/src/components/base-dialog.ts
@@ -1,7 +1,7 @@
 import "@home-assistant/webawesome/dist/components/dialog/dialog.js";
 
-import { LitElement, css, html } from "lit";
-import { customElement, property } from "lit/decorators.js";
+import { LitElement, css, html, nothing } from "lit";
+import { customElement, property, state } from "lit/decorators.js";
 
 import { dialogCloseButtonStyles } from "../styles/dialog-close-button.js";
 import { centeredMobileDialog } from "../styles/dialog-mobile.js";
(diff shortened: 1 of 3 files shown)


bdraco:
Addressed the footer-slot warning. Confirmed in the webawesome source that wa-dialog keys its footer chrome on a direct `[slot="footer"]` child element (`HasSlotController.hasNamedSlot` does `querySelector(":scope > [slot=\"footer\"]")`), not on flattened content, so an always-present forwarding slot would indeed draw an empty footer bar on the footer-less consumers. Fixed at the source in `base-dialog.ts`: the forwarding `<slot name="footer">` is now only rendered when a consumer actually fills the footer (same presence test against our own light DOM), so the four flagged dialogs and every other footer-less consumer render exactly as before; no per-dialog `::part(footer){display:none}` needed.

Verified with headless Chrome: a footer-less base-dialog renders no inner `<footer>` element at all (`hasInnerFooter:false`), while the onboarding wizard still renders its footer bar.

Also added the suggested regression tests for the migrated close / error gating: a failed save keeps the dialog open with an inline error, a successful save closes via the reactive flag and suppresses the session-dismiss, and request-close flips the open flag.

PR #572, opened 2026-06-03
Add schema-backed hover docs to the YAML editor

bdraco:

> **Transitional.** This exists to keep parity with the legacy dashboard's
> Monaco hover. As the structured editor grows to cover more components,
> these tooltips get less valuable — they only fire for what the form
> *can't* document — so we'll likely remove this entirely down the line.
> Scoped narrowly on purpose to minimise what we'd later rip out.

Adds schema-backed **hover documentation** to the YAML editor — but only
for the things the **structured editor can't already document on screen**.

The structured editor renders a form (with field docs) for every component
that has `config_entries`. Hovering those would just duplicate what's
already visible, so the hover is gated to the components it *can't*
form-edit, using the structured editor's own `isYamlOnlySection` so the two
stay in sync.

### What gets a hover vs. what doesn't

A token gets a tooltip only when its resolved component has **no
`config_entries`** (no form), or is a hardcoded YAML-only section
(`packages`, `external_components`).

- **Hover shows** — form-less components: `ethernet`, `spi`, `psram`,
  `lvgl`, `json`, `libretiny`, `packages`, `external_components`, and ~20
  other niche ones, plus bare platform-domain headers (`sensor:`,
  `binary_sensor:`, `light:`, …) that have no top-level form.
- **Hover excluded** — everything form-backed (~206 sections): `esphome`,
  `wifi`, `api`, `logger`, `web_server`, `mqtt`, `captive_portal`,
  `globals`, `substitutions`, … and every `- platform:` item's fields
  (`sensor.dht`, `binary_sensor.gpio`, `ota.esphome`, …), since the
  resolved `domain.platform` component carries `config_entries`.

So hover is off for the bulk of real config (which the form documents) and
on for the form-less components the editor can't represent. For the shown
components, every token still resolves docs (component description, nested
keys, enum values) from the schema bundle (`schema.esphome.io`) with the
catalog as fallback.

**Also fixed:** `fetchBundle` now aliases the `core` bundle to
`esphome.json` (where the core action/condition registries actually live);
a `core.*` registry ref was fetching a nonexistent `core.json` → 404,
breaking schema-backed hover/completion.

- N/A

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/yaml-editor.ts
+++ b/src/components/yaml-editor.ts
@@ -8,10 +8,12 @@ import { basicSetup, EditorView } from "codemirror";
 import { css, html, LitElement } from "lit";
 import { customElement, property, query, state } from "lit/decorators.js";
 import type { ESPHomeAPI } from "../api/esphome-api.js";
-import { apiContext, darkModeContext } from "../context/index.js";
+import { type LocalizeFunc } from "../common/localize.js";
+import { apiContext, darkModeContext, localizeContext } from "../context/index.js";
 import { ESPHOME_YAML_INDENT, esphomeYaml } from "../util/esphome-yaml-lang.js";
 import { createYamlCompletionSource } from "../util/yaml-completion.js";
 import { vscodeDark, vscodeLight } from "../util/yaml-editor-theme.js";
+import { createYamlHoverTooltip } from "../util/yaml-hover.js";
 import { createBackendYamlLinter } from "../util/yaml-lint-backend.js";
 import type { YamlSection } from "../util/yaml-sections.js";
 import {
@@ -53,6 +55,10 @@ export class ESPHomeYamlEditor extends LitElement {
   @consume({ context: apiContext })
   @state()
   private _api?: ESPHomeAPI;
+
+  @consume({ context: localizeContext, subscribe: true })
+  @state()
+  private _localize: LocalizeFunc = (key) => key;
 
   @property() value = "";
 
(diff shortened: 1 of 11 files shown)


bdraco:
With https://github.com/esphome/device-builder-frontend/pull/573 I think the answer is only if the structured editor doesn't display it should we hover. So they still get help for ethernet but not for things the structured editor knows about

bdraco:
Addressed the silent-failure items in 62630ad:

- **Swallowed `loadCatalog` catch** + **unhandled rejection from `resolveHoverTarget`**: consolidated both into a single try/catch around the whole resolve (catalog + schema) that logs via `console.debug` and degrades to no tooltip. The per-gate fail-open catch in `isYamlOnlyComponent` is now redundant, so I dropped it — the optional-chained `config_entries` (the actual ethernet fix) plus this boundary cover it.
- **Negative caching in `getConfigVarDocsAtPath`**: evict the cache entry on rejection (`promise.catch` guarded by identity) so a transient failure retries on the next hover instead of replaying a rejected promise.

Left the `toString().split()` per-hover allocation as-is per your note — it's gated by the (now 500ms) `hoverTime` and mirrors the completion source; happy to switch the line-walkers to iterate the `Text` directly as a follow-up if we want.

bdraco:
Addressed in dda4497: switched the outer hover catch to `console.warn` so a failure is visible at the default console level. As you noted, the catalog/schema fetches already degrade gracefully (and `console.debug` their own misses), so this outer catch only fires on an *unexpected* throw — which is exactly what warrants a visible warning.

Leaving the per-hover `toString().split()` as the noted low-priority follow-up.

PR #574, opened 2026-06-03
Make YAML editor validation errors clearer and better placed

bdraco:

Several improvements to how the YAML editor's backend linter surfaces
ESPHome config errors. Before, a structural error underlined the wrong
lines (and leaked absolute paths), and there was no document-level signal
that the whole config failed to validate.

- **Point block-level errors at the offending key.** ESPHome marks
  errors like `Component not found` / `Platform missing` on the block's
  *value* mapping, so the red underline smeared across the child lines,
  blank lines, and comments instead of the offending `key:`. A multi-line
  range now walks up to its enclosing key and marks just that token.
- **Line-number-as-error-dot instead of a lint-gutter column.** The
  error line's number becomes a red dot (via `gutterLineClass`) rather
  than reserving a separate `lintGutter()` column, so an error never
  widens the gutter or reflows the editor.
- **Document-level errors go in a banner above the editor.** A structural
  error esphome pins on the root `esphome:` block (e.g. "Platform
  missing"), or a parse error with no usable position, surfaces in a
  banner — squiggling the `esphome:` header is noise. Localized errors (a
  component typo, a syntax error on a line) keep their inline squiggle.
  The banner and the squiggle are mutually exclusive.
- **Correct YAML parse-error position.** PyYAML reports the context mark
  (where the block started, often line 1) first and the problem mark
  (where the bad token was found) last; we now use the last `line N,
  column M`, so the squiggle lands on the real line.
- **No path leak.** Absolute filesystem paths embedded in PyYAML /
  ESPHome messages (`"/Users/me/esphome/foo.yaml"`, `"C:\…\foo.yaml"`)
  are collapsed to their basename so the host's directory layout and
  username don't reach the UI.

New unit tests cover the key-retarget walk, the path sanitizer, and the
parse-error position extraction.

- Companion backend fix: esphome/esphome#16777 (points the "Component not found" range at the key). The frontend heuristic handles it for the currently pinned esphome too, so this stands alone.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/device-editor.styles.ts
+++ b/src/components/device/device-editor.styles.ts
@@ -309,6 +309,47 @@ export const deviceEditorStyles = css`
     font-weight: var(--wa-font-weight-bold);
   }
 
+  /* Document-level "configuration invalid" banner above the editor.
+     A flex child of .editor-pane (column + gap), so it sits above the
+     editor body and the body's flex:1 reclaims the rest. */
+  .invalid-banner {
+    flex: 0 0 auto;
+    display: flex;
+    align-items: flex-start;
+    gap: var(--wa-space-s);
+    padding: var(--wa-space-s) var(--wa-space-m);
+    border-radius: var(--wa-border-radius-m);
+    background: var(--wa-color-danger-fill-quiet);
+    border: var(--wa-border-width-s) solid var(--wa-color-danger-60);
+    color: var(--wa-color-danger-text-normal);
+  }
+
+  .invalid-banner-icon {
+    flex: 0 0 auto;
+    font-size: 1.25rem;
+    margin-top: 0.05rem;
+    color: var(--wa-color-danger-60);
+  }
+
+  .invalid-banner-text {
+    display: flex;
+    flex-direction: column;
+    gap: 3px;
+    line-height: 1.4;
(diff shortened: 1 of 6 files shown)


bdraco:
Addressed in bd351b4:

- **Stale banner after device switch** (review suggestion): each lint result is now stamped with the configuration it was computed for; `_onYamlDiagnostics` drops a result whose stamp ≠ the open file, so an in-flight validate for the previous device can no longer flash a stale "invalid" banner. The `updated()` clear stays for the immediate case.
- **Copilot — core-block detection**: replaced `sliceString(from, to) === "esphome"` with `keyAt(doc, from)` (matches `KEY_LINE_RE` on the diagnostic's line), so an expanded/whitespace-padded range can't misclassify a whole-config error.
- **Copilot — empty message**: validation messages now fall back to "Invalid configuration" when they trim to empty, mirroring the YAML path, so the banner never renders a blank line.
- **Copilot — aria-live re-announce**: `_onYamlDiagnostics` shallow-compares the new error list and skips the state update when unchanged, so an identical lint pass no longer re-renders / re-announces the banner.

PR #578, opened 2026-06-03
Make optional enum selects clearable instead of a mid-list "(none)"

bdraco:

Optional enum fields (e.g. a sensor's `device_class`) ship an empty-value
`(none)` option from the catalog, which sorts **mid-list** — so to clear
the field the user had to scroll the dropdown to find `(none)`, or type a
hacky `""`. (Screenshot: `(none)` sitting between `duration` and `energy`.)

Render those selects with a built-in clear (×) instead, and drop the
`(none)` pseudo-option from the list. `wa-select` fires a `change` event
with an empty value when cleared, which the existing change handler already
maps, so no new wiring. A required enum (no empty option) stays
non-clearable with its full option list.

The clearable/visible-options derivation is memoized per `ConfigEntry`
(a stable catalog object) via a `WeakMap`, matching the renderer
memoization pattern in `nested.ts` / `templatable.ts`.

Same "give the user a real clear affordance" intent as #576 (the search-box
×), applied to the structured editor's selects.

- N/A

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/config-entry-renderers/primitives.ts
+++ b/src/components/device/config-entry-renderers/primitives.ts
@@ -23,6 +23,26 @@ import {
   type RenderCtx,
 } from "../config-entry-renderers-shared.js";
 
+// An empty-value "(none)" option marks an optional enum — surface a clear
+// (×) and drop the pseudo-option. Memoized per entry (stable catalog object).
+const _selectOptions = new WeakMap<
+  ConfigEntry,
+  { clearable: boolean; visibleOptions: NonNullable<ConfigEntry["options"]> }
+>();
+
+export function selectOptions(entry: ConfigEntry) {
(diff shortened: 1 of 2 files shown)


bdraco:
Addressed both review nits in 58a753e:

- **Copilot** — `wa-select` isn't an `HTMLSelectElement`; the change handlers now read the value via the repo's `as unknown as { value: string }` pattern (`registry-list.ts`). Fixed both selects in `renderSelectField` for consistency.
- **Kōan** — dropped the `as never` on the `OPTIONAL` / `REQUIRED` test fixtures. `ConfigValueOption` is just `{ label, value }`, which the literals already satisfy, so the cast was unnecessary cargo from `makeEntry`.

PR #583, opened 2026-06-03
Show and expand pin mode scalar shorthand in the structured editor

### LINKED ISSUE #579: Pin mode scalar shorthand (mode: OUTPUT) hidden behind Show advanced and not editable
## Summary

A pin written with the scalar **mode shorthand** (e.g. `mode: OUTPUT`) is hidden in the structured editor unless the user enables "Show advanced", and even then it renders as an empty "Mode" group rather than showing the `OUTPUT` value. A field the user set in YAML should be visible without the advanced toggle, and should reflect its value.

## Reproduction

Device with a gpio switch on an expander pin:

```yaml
pca9554:
  - id: pca9554_hub_out_1
    address: 0x20

switch:
  - platform: gpio
    pin:
      pca9554: pca9554_hub_out_1
      number: 0
      mode: OUTPUT
    name: Open
```

Open the switch section in the structured editor. The pin "Advanced" disclosure does not appear; `mode` only shows after toggling "Show advanced", and then as an empty Mode group (not `OUTPUT`).

## Root cause

The pin `mode` schema entry is type `NESTED` (because `mode` also accepts an object form such as `{input: true, pullup: true}`) and is marked `advanced`. The user's value is the scalar shorthand `"OUTPUT"`. Two gates in `src/components/device/config-entry-render-filter.ts` then hide it:

1. `hasMaterialValue` NESTED branch: `if (!isPlainObject(value)) return false;` so a scalar shorthand reads as "no value set", and the advanced gate drops the field.
2. `filterRenderable` also drops a non-multi_value NESTED entry when no children render; `asRecord("OUTPUT")` is `{}`, so no child fields survive to keep it alive.

Separately, `renderNestedField` (`src/components/device/config-entry-renderers/nested.ts`) has no representation for a scalar value, so even when shown it paints an empty group instead of `OUTPUT`.

## Proposed fix

1. `hasMaterialValue`: treat a defined non-object value on a NESTED entry as material (a scalar shorthand the user set).
2. `filterRenderable`: keep a NESTED entry with a material value even when it has zero renderable children.
3. `renderNestedField` (or the pin renderer): when a NESTED entry holds a scalar shorthand, render it as a read-only "set in YAML, edit there" notice (mirroring the existing `YamlRawValue` multi-value notice) or a scalar fallback, so the value is visible and not silently editable-as-empty.

Needs a small UX decision on the scalar representation (notice vs. inline scalar editor).

## Notes

Pre-existing; not introduced by the field-level-sync work in #573 (the `filterRenderable` / `hasMaterialValue` gates predate it). Found while testing #573.

bdraco:

A pin written in long form with a scalar **mode shorthand** (`mode: OUTPUT`) was mishandled in the structured editor: it read as "no value set", so the advanced gate dropped it, the pin **Advanced** disclosure disappeared entirely, and forcing it visible via "Show advanced" painted an empty **Mode** group that didn't reflect `OUTPUT`.

The mode shorthands (`OUTPUT`, `INPUT_PULLUP`, …) are fixed, platform-independent presets of the same five flags the catalog already models (`input`/`output`/`pullup`/`pulldown`/`open_drain`), per ESPHome's `_set_mode` / `PIN_MODES`. So `mode: OUTPUT` is exactly `mode: {output: true}` and the existing flag checkboxes can model it.

Changes:

- Treat a defined scalar at a NESTED key as a material value, so it stays visible without "Show advanced", and keep such a group even when no child renders for it (`config-entry-render-filter.ts`).
- Expand a known mode shorthand to its flags for display so the checkboxes reflect it; the YAML scalar is preserved until the user toggles a flag, which writes the flag-object form (`config-entry-pin-renderer.ts`, new `src/util/pin-mode.ts`).
- For an unknown shorthand, or any other NESTED scalar, render a read-only "edit in YAML" notice showing the value instead of an empty group (`config-entry-renderers/nested.ts`).

Verified in the browser: with `mode: OUTPUT`, the pin Advanced disclosure opens, the Mode group shows **Output** toggled on, and the YAML still reads `OUTPUT` until a flag is changed.

- fixes https://github.com/esphome/device-builder-frontend/issues/579

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/config-entry-pin-renderer.ts
+++ b/src/components/device/config-entry-pin-renderer.ts
@@ -8,9 +8,10 @@
 import { html, nothing, type TemplateResult } from "lit";
 import type { BoardPin } from "../../api/types/boards.js";
 import type { ConfigEntry } from "../../api/types/config-entries.js";
-import { PinFeature, PinMode } from "../../api/types/config-entries.js";
+import { ConfigEntryType, PinFeature, PinMode } from "../../api/types/config-entries.js";
 import { findUsedPins, sectionEndLine } from "../../util/config-entry-yaml-scan.js";
 import { isPlainObject, isPrimitiveOrNullish } from "../../util/nested-values.js";
+import { expandPinModeShorthand } from "../../util/pin-mode.js";
 import {
   effectiveDisabled,
   fieldKeyAttr,
@@ -19,6 +20,8 @@ import {
   renderStringField,
   type RenderCtx,
 } from "./config-entry-renderers-shared.js";
+import { renderNestedField } from "./config-entry-renderers/nested.js";
+import { renderBooleanField } from "./config-entry-renderers/primitives.js";
 
 /**
  * Parse a pin reference into a GPIO number. Used both for the field's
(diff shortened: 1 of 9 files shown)


bdraco:
Two follow-ups split out of this PR's scope:

- #584 - pin mode flag checkboxes ignore the pin registry's allowed modes (an expander pin can be toggled into combinations ESPHome rejects). Pre-existing; this PR newly exposes the flag toggles for the scalar-shorthand case too. Needs a backend catalog change plus frontend scoping.
- #585 - YAML-cursor navigation can't auto-open a manually-collapsed pin Advanced disclosure.

Issue #549, opened 2026-06-02
Migrate remaining raw wa-dialog dialogs onto esphome-base-dialog


bdraco:
## Migrate remaining dialogs onto `esphome-base-dialog`

PR #548 fixed the close (X) button getting pushed off-screen by a long title, but the fix lives in the shared `esphome-base-dialog` wrapper, so it only reaches dialogs built on that wrapper. CSS `::part` crosses just one shadow boundary, so there is no global rule that can reach a dialog that renders `wa-dialog` directly; each has its own shadow root.

The dialogs below still use `wa-dialog` directly and so do not get the title-ellipsis / close-button-clearance guarantee (or the other shared wrapper behaviour: busy gate, light-dismiss wiring, mobile sizing). Migrating each onto `esphome-base-dialog` brings them in line and closes the gap. `esphome-clone-device-dialog` is the highest priority; it embeds the device name in its title, so it can reproduce the same overflow today.

### Unmigrated dialogs

- [x] `esphome-clone-device-dialog` — `components/clone-device-dialog.ts` (long dynamic title; highest priority)
- [x] `esphome-rename-device-dialog` — `components/rename-device-dialog.ts`
- [x] `esphome-logs-method-dialog` — `components/logs-method-dialog.ts`
- [x] `esphome-yaml-validation-dialog` — `components/yaml-validation-dialog.ts`
- [x] `esphome-unsaved-changes-dialog` — `components/unsaved-changes-dialog.ts`
- [x] `esphome-create-config-dialog` — `components/wizard/create-config-dialog.ts`
- [x] `esphome-labels-filter` — `components/labels/labels-filter.ts`
- [x] `esphome-device-labels-editor` — `components/labels/device-labels-editor.ts`
- [x] `esphome-add-component-dialog` — `components/device/add-component-dialog.ts`
- [x] `esphome-add-script-dialog` — `components/device/add-script-dialog.ts`
- [x] `esphome-add-automation-dialog` — `components/device/add-automation-dialog.ts`
- [x] `esphome-add-api-action-dialog` — `components/device/add-api-action-dialog.ts`
- [x] `esphome-catalog-picker-dialog` — `components/device/automation-editor/catalog-picker-dialog.ts`

Related: #548, #39

PR #589, opened 2026-06-03
Migrate the create-config wizard onto esphome-base-dialog

bdraco:

Migrates `esphome-create-config-dialog` — the new-device wizard (method → board → setup / empty-config / import) — off a directly-rendered `<wa-dialog>` and onto the shared `esphome-base-dialog` wrapper.

Rendering `wa-dialog` directly meant the wizard missed the wrapper's shared behaviour: the busy/light-dismiss gate, the `wa-hide`/`wa-after-hide` wiring (incl. the reopen fix), and the title-ellipsis / close-button-clearance fix (#548).

The conversion uses the established reactive pattern: a state-driven `_open` flag (the dialog already tracked `_open` for the steps' Enter listeners — it's now the single source of truth), a `request-close` handler, and `after-hide` teardown. The public `open()` / `openWithBoard()` / `openAtBoardStep()` methods and the child-step events (`next-step`, `finish-setup`, `create-empty-config`, `import-file`) are unchanged, so the dashboard consumer needs no edits.

### New: `header-prefix` slot on `esphome-base-dialog`

Unlike the other migrated dialogs, the wizard renders a **back arrow to the left of the title** inside the header. The wrapper only had a `header-suffix` slot (content *after* the title), so I added a `header-prefix` slot (symmetric, rendered before the title). It's **empty by default**, so the ~8 already-migrated dialogs are unaffected; only the wizard fills it. The primary-colored header, 40×40 close button, wide (750px) board step, and fullscreen-on-mobile all carry over via the forwarded `::part`s.

Verified in the browser: header look unchanged (blue bar, white title + close X), back arrow appears on non-first steps and steps back correctly, board step is wider, close/Escape/outside-click + reopen all work, and the create flows still land in the device editor.

- part of #549 (next unchecked dialog in the list)

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/base-dialog.ts
+++ b/src/components/base-dialog.ts
@@ -83,6 +83,9 @@ import { centeredMobileDialog } from "../styles/dialog-mobile.js";
  *   dialogs that do use a pinned footer row (e.g. the onboarding
  *   wizard). Only forwarded when a consumer fills it, so footer-less
  *   dialogs render unchanged (see ``willUpdate``).
+ * - ``header-prefix`` slot: inline content before the title
+ *   (e.g. a wizard back button). Empty by default. Symmetric
+ *   with ``header-suffix``.
  * - ``header-suffix`` slot: inline content after the title
  *   (e.g. a status chip). Empty by default, so other dialogs
  *   are unchanged. The row and title are exposed as the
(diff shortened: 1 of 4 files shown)


bdraco:
Thanks — addressed all four in dd70640:

- **Busy gate (Copilot):** bound `_submitting` to `esphome-base-dialog`'s `?busy`, so Escape / outside-click / close can't dismiss while a create or import request is in flight.
- **Back-button a11y (Copilot):** the icon-only header back button now has `aria-label` + `title` via the existing `layout.back` key.
- **header-prefix slot test (Copilot):** added `test/components/base-dialog-header-prefix.test.ts` — pins that the slot renders before the title and stays an empty no-op by default.
- **Open/close contract test (Kōan):** added the `open() / openAtBoardStep() / request-close / close()` reactive `_open` block the sibling #549 migrations carry.

PR #590, opened 2026-06-03
Scope long-form pin Mode flags to the pin's provider

### LINKED ISSUE #584: Pin mode flag checkboxes ignore the pin registry's allowed modes (offer invalid flags for expander pins)
## Summary

The long-form pin **Mode** group always offers the generic esp32 five flags (`input` / `output` / `pullup` / `pulldown` / `open_drain`), regardless of which pin registry the value uses. But each pin registry declares its own allowed modes and a combination validator, so the editor lets users build configs ESPHome rejects.

## Reproduction

A gpio switch/binary_sensor on a pca9554 expander pin, then open the pin Advanced > Mode group and toggle `pullup` / `pulldown` / `open_drain`:

```yaml
switch:
  - platform: gpio
    pin:
      pca9554: pca9554_hub_out_1
      number: 0
      mode:
        output: true
        pullup: true        # rejected
```

ESPHome reports `[pullup] is an invalid option for [mode]` (also pulldown / open_drain). Toggling both `input` and `output` is likewise rejected for pca9554.

## Root cause

ESPHome builds each pin schema with `pins.gpio_base_schema(..., modes=[...], mode_validator=...)`. The pca9554 provider passes `modes=[CONF_INPUT, CONF_OUTPUT]` and a `validate_mode` that requires exactly one of input/output (`esphome/components/pca9554/__init__.py`). Native esp32 allows all five. Our catalog's `_pin_long_form_extras` (device-builder backend, `script/sync_components.py`) attaches the generic esp32 flag set to every pin field and does not reflect the per-registry `modes=` list or validator.

## Proposed fix

- Backend: emit a per-pin-registry allowed-modes map (keyed by registry: pca9554, mcp23xxx, sx1509, native platforms, ...).
- Frontend: detect the pin's registry from the pin value (which expander key is present, or bare GPIO = the device's target platform) and render only the valid flag checkboxes, ideally enforcing the combination validator (e.g. input XOR output for pca9554).

## Notes

Pre-existing since the #430 long-form pin work; the object form already let you toggle all five on an expander pin. #583 (pin mode scalar shorthand) exposes the flag toggles for the scalar-shorthand case too, which surfaced this. Cross-repo: needs a device-builder backend catalog change plus the frontend scoping.

bdraco:

The long-form pin **Mode** checkboxes (`input` / `output` / `pullup` / `pulldown` / `open_drain`) were offered in full for every pin, but an external pin provider allows only a subset: an I2C expander like `pca9554` permits `input` / `output`, a shift register `sn74hc595` only `output`. The editor let users tick flags ESPHome rejects (`[pullup] is an invalid option for [mode]`).

This is the frontend half of #584; the backend (esphome/device-builder#1163) derives a `{provider_key: [allowed_modes]}` map from ESPHome's `PIN_SCHEMA_REGISTRY` and ships it over `components/get_pin_registry_modes`.

## Changes

- `ESPHomeAPI.getPinRegistryModes()` — typed wrapper + payload-shape filter, mirroring `getIntegrationDocs`.
- `src/util/pin-registry-modes-cache.ts` — session-scoped cache (fetched once, shared across forms, subscribe-to-populate; a failed fetch caches `{}` so it doesn't retry-storm).
- `config-entry-form` consumes `apiContext`, kicks the fetch when the context lands, and threads the map onto `RenderCtx.pinRegistryModes` (re-rendering when it populates).
- `config-entry-pin-renderer` detects the provider key present in the pin value (`pca9554`) and narrows the Mode group's flag children to the allowed set before rendering. A native pin (no provider key), an unknown provider, or a missing map keeps every flag — no regression.

## Verification

Unit tests cover the scoping (pca9554 hides pullup; native keeps all; unknown provider / empty map fall back). Verified end-to-end in the browser against the backend branch: with the pca9554 switch the Mode group shows only **Input / Output**, while the native GPIO33 binary_sensor still shows all five.

- fixes https://github.com/esphome/device-builder-frontend/issues/584
- Requires backend esphome/device-builder#1163 (the `components/get_pin_registry_modes` command + artefact); degrades gracefully to "show every flag" without it.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/api/esphome-api.ts
+++ b/src/api/esphome-api.ts
@@ -1381,9 +1381,11 @@ export class ESPHomeAPI {
     }
     const result: Record<string, string[]> = {};
     for (const [key, value] of Object.entries(raw)) {
-      if (typeof key === "string" && Array.isArray(value)) {
-        result[key] = value.filter((m): m is string => typeof m === "string");
-      }
+      if (typeof key !== "string" || !Array.isArray(value)) continue;
+      const flags = value.filter((m): m is string => typeof m === "string");
+      // Omit empty-after-filter providers; an empty allow-list would scope the
+      // Mode group to zero checkboxes instead of falling back to show-all.
+      if (flags.length > 0) result[key] = flags;
     }
     return result;
   }
--- a/src/api/esphome-api.ts
+++ b/src/api/esphome-api.ts
@@ -1364,15 +1364,10 @@ export class ESPHomeAPI {
   }
 
   /**
-   * Map of external pin provider → the long-form `mode` flags it allows
-   * (`pca9554` → `["input", "output"]`). The visual editor scopes the pin
-   * Mode checkboxes against this; a provider absent from the map (or a native
-   * pin, which carries no provider key) shows every flag. Fetched once per
-   * session — the dataset only refreshes with a backend release.
-   *
-   * The WS layer doesn't enforce a shape, so the payload is filtered to the
-   * `{string: string[]}` contract here: a non-object becomes `{}`, and any
-   * non-string flag inside a value array is dropped.
+   * Map of external pin provider → allowed long-form `mode` flags
+   * (`pca9554` → `["input", "output"]`), fetched once per session. Filtered to
+   * the `{string: string[]}` contract: non-object → `{}`, non-string flags
+   * dropped, empty providers omitted.
    */
   async getPinRegistryModes(): Promise<Record<string, string[]>> {
     const raw = await this.sendCommand<unknown>("components/get_pin_registry_modes");
--- a/src/api/esphome-api.ts
+++ b/src/api/esphome-api.ts
@@ -1363,6 +1363,31 @@ export class ESPHomeAPI {
     return result;
   }
 
+  /**
+   * Map of external pin provider → the long-form `mode` flags it allows
+   * (`pca9554` → `["input", "output"]`). The visual editor scopes the pin
+   * Mode checkboxes against this; a provider absent from the map (or a native
+   * pin, which carries no provider key) shows every flag. Fetched once per
+   * session — the dataset only refreshes with a backend release.
+   *
+   * The WS layer doesn't enforce a shape, so the payload is filtered to the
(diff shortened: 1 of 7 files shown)


bdraco:
All review points addressed in 521115d (backend #1163 is now merged):

**Copilot**
1. `getPinRegistryModes()` now omits a provider whose flags filter to empty, so an empty allow-list can't scope the Mode group to zero checkboxes.
2. `providerAllowedModes()` uses an own-property check (`Object.prototype.hasOwnProperty.call`, not `in`) so a pin-value key like `toString` can't match a prototype member, and treats an empty allowed list as no-scoping (show every flag).
3. The cache now isolates each listener call in try/catch + logs.
4. `_resetPinRegistryModesCache()` now clears `_listeners` too (test isolation).
5. Moved the wa-select-sync JSDoc back onto `_syncSelectValues`; `connectedCallback` got its own terse doc.

**Kōan**
- Suggestion #1 (reset clears listeners) and #2 (empty allow-list → show-all) covered by 4 and 2 above; added a unit test for the empty-list fallback.
- Silent-failure (MEDIUM): the failed-fetch `.catch` now `console.warn`s before caching `{}` (kept the empty-cache to avoid a per-render retry-storm, since the form kicks the fetch from `updated()`).

Also opened #592 to extract a shared fetch-once session-blob cache helper as a follow-up (the boilerplate `pin-registry-modes-cache` shares with `automation-catalog-cache`).

bdraco:
Addressed in cc2a4b3: `getPinRegistryModes()` now builds its result with `Object.create(null)` (and the cache's failure-fallback empty map too), so an untrusted `__proto__` / `constructor` key in the WS payload can't pollute the prototype on assignment. Lookups already used `Object.prototype.hasOwnProperty.call`, so the map is prototype-safe end to end.

The other inline threads (own-property check, empty-list omit, listener isolation, reset clears listeners, JSDoc relocation, legacy disallowed-but-present flag visibility + its test, fetch-once) were resolved in the earlier commits 521115d / fc62120 / c393365.

PR #591, opened 2026-06-03
Auto-open the pin Advanced disclosure for YAML-cursor navigation

### LINKED ISSUE #585: YAML-cursor navigation can't auto-open a collapsed pin Advanced (pin:pin-advanced) disclosure
## Summary

YAML-cursor navigation can't auto-open a manually-collapsed pin **Advanced** disclosure, so placing the cursor on a long-form pin field (`mode` / `inverted` / ...) while that disclosure is collapsed can't reveal and scroll to the exact field; it falls back to scrolling the pin container.

## Root cause

`FieldScrollController` opens ancestor groups by dotted path prefix (`"pin"`), but the pin long-form disclosure is keyed `${path.join(".")}:pin-advanced` (`"pin:pin-advanced"`), which the prefix walk never opens.

## Impact

Low today: the disclosure seed-opens by default when the pin carries long-form values, so this only bites a user-collapsed disclosure.

## Proposed fix

Make YAML-cursor navigation able to open the pin Advanced disclosure generically, without hardcoding the renderer's `:pin-advanced` key convention into the shared `FieldScrollController`. Fold it into the pin long-form work.

## Notes

Carried over from the #573 review (see the comment on #579). Related to #583.

bdraco:

YAML-cursor navigation reveals and flashes the form field under the
cursor by opening every ancestor group first. `FieldScrollController`
opened groups by **dotted path prefix** (`"pin"`), but the pin long-form
**Advanced** disclosure (#430) is keyed `${path.join(".")}:pin-advanced`
(`"pin:pin-advanced"`) — a synthetic key the prefix walk never produces.
So clicking the YAML on a long-form pin field (`mode` / `inverted` /
`mode.pullup`) while that disclosure was **manually collapsed** couldn't
reveal the field; it fell back to scrolling the pin container.

Low impact in practice: the disclosure seed-opens by default when the pin
already carries long-form values, so only a user-collapsed disclosure hit
this.

The fix keeps the `:pin-advanced` key convention out of the shared
controller. The disclosure already renders a DOM element even when
collapsed; it now **declares the path prefix it gates** via a
`data-reveal-for` attribute, and its existing `data-field-key` is the open
key. `FieldScrollController` discovers that declaration generically (it
already reads `data-field-key` from the shadow DOM) and opens any
disclosure whose declared prefix is a strict ancestor of the target path —
agnostic to any renderer's keying scheme. A new pure `gatingDisclosureKeys`
helper holds the prefix-match decision (unit-tested without a DOM, mirroring
`advanceScrollGate`).

Composes cleanly: the pin picker itself (`["pin"]`) stays untouched;
`["pin","mode","pullup"]` opens both the dotted `pin.mode` group and the
disclosure in one re-render; `openNested` is idempotent and no-ops in
`requiredOnly` mode.

- fixes esphome/device-builder-frontend#585

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/config-entry-pin-renderer.ts
+++ b/src/components/device/config-entry-pin-renderer.ts
@@ -313,7 +313,11 @@ function renderPinAdvanced(
   };
 
   return html`
-    <div class="pin-advanced" data-field-key="${advancedKey}">
+    <div
+      class="pin-advanced"
+      data-field-key="${advancedKey}"
+      data-reveal-for="${fieldKeyAttr(path)}"
+    >
       <button
         type="button"
         class="pin-advanced-toggle"
--- a/src/components/device/field-scroll-controller.ts
+++ b/src/components/device/field-scroll-controller.ts
@@ -48,6 +48,19 @@ export function advanceScrollGate(
   return { gate: { scrolledKey, lastFocusKey, tries }, scroll };
 }
 
+/** Open keys for disclosures that gate *path* — a disclosure declares the
+ *  path prefix it hides behind; a strict-ancestor prefix means it must open. */
+export function gatingDisclosureKeys(
+  decls: ReadonlyArray<{ prefix: string[]; key: string }>,
+  path: string[]
+): string[] {
+  return decls
+    .filter(
+      (d) => d.prefix.length < path.length && d.prefix.every((k, i) => k === path[i])
+    )
+    .map((d) => d.key);
+}
+
 /** The form surface this helper drives. */
 export interface FieldScrollHost {
   shadowRoot: ShadowRoot | null;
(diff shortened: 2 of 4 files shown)


bdraco:
Both review points addressed in 1ff9625:

- **Copilot (empty `data-reveal-for`)** — valid: `parseFieldKey("")` → `[]`, and `[].every(...)` is vacuously true, so an empty/malformed prefix would gate *every* target. Guarded in the pure `gatingDisclosureKeys` (the single decision point) with `d.prefix.length > 0`, plus a regression test (`ignores an empty-prefix decl instead of gating every path`).
- **Kōan (shadow-root scope)** — `_gatingDecls` intentionally scans only the host shadow root (where pin Advanced renders); went with the documenting comment rather than speculative recursion, noting a future nested-shadow disclosure would need the same recursion `_find` uses.

Full suite: 2017 passing.

Issue #600, opened 2026-06-03
Triage: dialog styles that nearly-match the shared chrome (drift vs intentional)


bdraco:
## Context

While doing the DRY follow-up to #549 (extracting `dialogChromeStyles` / `quietCloseButtonStyles` / `primaryHeaderDialogStyles` into `src/styles/dialog-chrome.ts` and routing the dialogs onto them), a number of dialogs turned out to **nearly** match a shared block but differ in one detail. Those couldn't be folded into the shared fragments without a behavior change, so they were left as-is.

This issue catalogs those near-matches so we can decide, per item, whether the difference is **unintended drift** (→ unify onto the shared fragment) or **genuinely intentional** (→ leave, maybe with a comment explaining why). None of these are bugs today; this is cleanup triage.

### Already consolidated (for reference)
- `primaryHeaderDialogStyles` — `create-config-dialog`, `add-component-dialog`
- `dialogChromeStyles` (neutral header/title/footer, also composed into `modalDialogStyles`) + `quietCloseButtonStyles` — `confirm`, `yaml-validation`, `clone-device`, `rename-device`, `api-key`, `friendly-name`, `adopt`, `device-labels-editor`, `labels-filter`, `pair-build-server`, `feedback`, `bulk-labels`

---

### 1. Process-terminal dialogs share the primary header *bar* but not the rest
`command-dialog`, `firmware-install-dialog`, `logs-dialog` all render a primary-colored `::part(header)` that is identical to `primaryHeaderDialogStyles`' header bar, but each diverges on:
- **title** adds `font-family: <monospace stack>` (the shared primary title has no font-family);
- **close button** — they don't carry the shared 40×40 on-primary `::part(close-button__base)` reset (they rely on the base-dialog default);
- **body** is terminal-specific (`padding: 0; background: var(--term-bg)`).

**Decide:** should these consume `primaryHeaderDialogStyles` for the header bar (overriding title `font-family` + body locally), or is the "terminal dialog" header a deliberately separate family? If the latter, a shared `terminalDialogHeaderStyles` fragment for the three of them may still be worth it (they're identical to *each other*).

### 2. `::part(body)` bottom-padding drift (neutral dialogs)
- `0 var(--wa-space-l) var(--wa-space-l)` — `feedback`, `labels-filter`, `device-labels-editor`
[...]

PR #602, opened 2026-06-03
Consolidate duplicated dialog header/close styles (DRY)

bdraco:

DRY follow-up to the #549 dialog migrations. Auditing the dialogs afterward surfaced three style blocks copied near-verbatim across many components. This hoists them into a single source of truth — `src/styles/dialog-chrome.ts` — and routes every consumer onto it.

### New shared fragments (`src/styles/dialog-chrome.ts`)

- **`primaryHeaderDialogStyles`** — the branded primary-colored header bar + flush 40×40 on-primary close button + `header-prefix` back button. Was duplicated byte-for-byte in **create-config** and **add-component**.
- **`dialogChromeStyles`** — the neutral header padding + title typography + hidden footer. This is exactly what the existing `modalDialogStyles` already encoded, yet ~10 dialogs hand-rolled their own copy. `modalDialogStyles` now **composes** `dialogChromeStyles` (so confirm / yaml-validation are unchanged), and the form/picker dialogs use it directly: **clone-device, rename-device, api-key, friendly-name, adopt, feedback, bulk-labels, pair-build-server, device-labels-editor, labels-filter**.
- **`quietCloseButtonStyles`** — the flat (transparent) close-button reset shared by **clone, rename, device-labels-editor, labels-filter**.

### Why it's safe

Pure refactor: every rule is **relocated unchanged**. Per-dialog deltas (`--width`, body padding, banners) stay local. Within any single dialog the shared fragment's selectors don't overlap the local rules, so the cascade is identical. Verified by `npm run build` + the full suite (**2059 tests**). 17 files, −273 duplicated lines (the 117-line shared module is the only real addition).

### Not force-fit (see #600)

A few dialogs *nearly* match but differ — the process-terminal dialogs' monospace title (command/firmware/logs), and a couple of body-padding variants (e.g. bulk-labels' `--wa-space-m` bottom). Rather than bend them to fit, those are catalogued in #600 for a separate "unintended drift vs intentional" decision.

- follow-up to #549; drift triage tracked in #600

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/adopt-dialog.ts
+++ b/src/components/adopt-dialog.ts
@@ -5,6 +5,7 @@ import type { ESPHomeAPI } from "../api/esphome-api.js";
 import type { AdoptableDevice } from "../api/types/devices.js";
 import type { LocalizeFunc } from "../common/localize.js";
 import { apiContext, localizeContext } from "../context/index.js";
+import { dialogChromeStyles } from "../styles/dialog-chrome.js";
 import { inputStyles } from "../styles/inputs.js";
 import { espHomeStyles } from "../styles/shared.js";
 import { validateDeviceName } from "../util/config-validation.js";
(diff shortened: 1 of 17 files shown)


bdraco:
Thanks both — addressed in 386a840:

**Copilot** (×3, same issue): right, the comments wrongly listed the command dialog as a `primaryHeaderDialogStyles` consumer. It isn't — it keeps its own bespoke header (monospace title + terminal body) and is intentionally left out (tracked in #600). Dropped the command-dialog reference from the JSDoc and both consumer comments, and noted *why* the process-terminal dialogs aren't folded in.

**Kōan** (suggestion #1 — labels-filter scoping/specificity): confirming it's intentional, not incidental. `labels-filter` renders a single `<esphome-base-dialog class="create-dialog">`, so the bare `esphome-base-dialog::part(...)` from the shared fragments lands on exactly that dialog; the inline comment documents the single-dialog assumption. Keeping the shared fragments (vs re-class-scoping) is deliberate — re-scoping would re-introduce the duplication this PR removes, and no `espHomeStyles`/`facetStyles` rule targets `esphome-base-dialog::part(...)`, so the lower specificity has nothing to lose to today.

PR #627, opened 2026-06-04
Align sticky-scroll line numbers with the editor gutter

### LINKED ISSUE #626: Sticky scroll header line numbers are offset from the editor gutter
The YAML editor's sticky scroll (added in #327) works, but the line numbers in the sticky header are horizontally offset from the editor's line-number gutter — they don't line up as one continuous column with the numbers below the header. Per @bdraco this is a CSS alignment issue, not intentional.

### Cause

The sticky row's line-number column (`.cm-esphome-sticky-num` in `src/util/yaml-sticky-theme.ts`) is auto-width (`flex: 0 0 auto`) with `padding: 0 8px` and `text-align: right`, which doesn't match CodeMirror's gutter (`.cm-gutters` / `.cm-lineNumbers .cm-gutterElement`). Sizing/padding `.cm-esphome-sticky-num` to match the gutter should line the columns up.

### Notes

- Low priority / non-blocking.
- Screenshots available in the ESPHome Discord thread.

Builder vs vscode
Image
Image

bdraco:

The sticky scroll header line numbers were offset from the editor gutter; they measured the full `.cm-gutters` width, which includes the fold gutter that `basicSetup` mounts to the right of the line numbers, so the right aligned number landed past the real line number column. This measures the `.cm-lineNumbers` column on its own and pushes the pinned number back onto its right edge, so the sticky numbers line up as one continuous column with the gutter below. The divider under the sticky header is also a touch taller and darker while I was in there.

- fixes https://github.com/esphome/device-builder-frontend/issues/626

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/util/yaml-sticky-render.ts
+++ b/src/util/yaml-sticky-render.ts
@@ -4,6 +4,17 @@ import type { Tree } from "@lezer/common";
 import { highlightTree } from "@lezer/highlight";
 import type { StickyScopeLine } from "./yaml-sticky-scope.js";
 
+/** Right inset that lands the full-gutters-wide num span's glyph on the
+ *  line-number column's right edge: past the fold gutter, plus CM's 3px
+ *  cell inset (gutter cells use padding 0 3px 0 5px). */
+export function stickyNumPaddingRight(
+  gutterWidth: number,
+  lineNumberWidth: number
+): number {
+  if (gutterWidth <= 0 || lineNumberWidth <= 0) return 8; // pre-measure fallback
+  return gutterWidth - lineNumberWidth + 3;
+}
+
 export function createStickyRow(): HTMLDivElement {
   const row = document.createElement("div");
   row.className = "cm-esphome-sticky-line";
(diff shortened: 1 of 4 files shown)


bdraco:
The NUL-byte scrub and `.gitattributes` that made this file's diff binary have moved to their own hygiene PR, #628, which also adds a pygrep pre-commit hook so a NUL can't slip into a source file again. Once #628 merges, this branch rebases onto a clean `main` and the `yaml-sticky-render.ts` diff renders as text. Until then the binary diff here is expected.

bdraco:
Addressed suggestion #1 in f9f69fb. Instead of the hard-coded `3`/`5`, the measure phase now reads the line-number gutter cell's computed `paddingLeft`/`paddingRight` and bundles the gutter geometry into a `GutterMetrics` object; `stickyNumPaddingRight` takes the measured `padRight`, and the row's left inset mirrors the measured `padLeft` (the theme's `5px` is now just the pre-measure fallback). A CodeMirror bump or custom gutter theme that changes the cell padding now self-corrects. Added a test asserting the helper honors the measured inset rather than a constant.

PR #634, opened 2026-06-05
Derive component category labels deterministically instead of translating

### LINKED ISSUE #636: [Bug] Some categories do not have translation keys defined
### What happened?

The Add component wizard have a mix of languages in the categories panel. Translation keys were not provided for some categories in Lokalise

Image

### Dashboard version

Latest Beta

### How did you install it?

PyPI

### Where is it running?

Local — Linux

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

esphbot:

The Add component dialog's category sidebar resolved each label via
`device.component_category_<id>`, falling back to the backend-supplied
English name when a key was missing. Only ~12 of the ~40 categories the
backend emits had keys, so the panel rendered a mix of the user's language
(keyed categories) and English (everything else) — see the screenshot in
the issue (#636).

Per the maintainer call (@bdraco: "rip them out for now so it's all English
since that is what the yaml uses anyways"), this fixes the mixed-language
bug by **removal rather than completion**:

- `filters.ts` now derives each sidebar label deterministically from the
  category id via the shared `categoryChipLabel` helper — the same source
  the component card chip uses — so the sidebar and the card chip render
  identical English for every category (no more `Canbus` vs `CAN Bus`).
- Removes the 10 yaml-mapped `component_category_*` keys from `en.json`.
  The synthetic UI labels `all` / `featured` are kept (they have no yaml
  equivalent).

Categories map directly to YAML keys, so consistent English is correct here;
a per-category i18n table would drift out of step with the chip as new
categories ship from the sync script.

Note: sidebar text shifts plural→singular as a side effect (`Sensors`→`Sensor`),
matching the card chip wording.

- fixes esphome/device-builder-frontend#636

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


jesserockz:
I'm not sure we should translate the component "categories". They directly map to the yaml keys.

bdraco:
@esphbot rr

bdraco:
I have no strong opinions here but half translated is bad. Easy to rip them out now than doing it later so that seems like the right call as its always easier to remove before we go to stable while its still beta

bdraco:
Let's rip them out for now so its all english since that is what the yaml uses anyways

bdraco (changes requested):
PR summary does not reflect the state of this PR

bdraco (dismissed):
Pr title is stale

bdraco:
@esphbot Rr

jesserockz (approved):
I like it, will test now

bdraco (approved):

PR #647, opened 2026-06-05
Round-trip unicode escape sequences in font glyph fields

### LINKED ISSUE #1232: [Bug] Visual editor can not parse glyphs under font->extras with unicode
### What happened?

When adding extra glyphs to a font set, the visual editor fails to parse a list of unicode escape sequences. Also, it does not allow to define an unicode escape sequence because it doesn't wrap the value with quotes.

Image

--------------------------------------------------------------------------------

Image

P.D.: I'll check the glyph key under font because it probably suffers from the same bug. In that case, I could open a new issue.

### Dashboard version

Latest Beta

### How did you install it?

PyPI

### Where is it running?

Local — Linux

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

The visual editor could not round-trip unicode escape sequences in font glyph fields. A flow list such as `glyphs: ["\U000F058F","\U000F0F19"]` loaded as an empty field, and a typed glyph was emitted as a bare unquoted scalar that ESPHome then read character by character.

The form parser now decodes double quoted escapes, a list nested inside a list item serializes as a flow list instead of collapsing to a scalar, and Private Use and control code points are quoted on emit and shown as editable escapes in the input.

- fixes https://github.com/esphome/device-builder/issues/1232

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/config-entry-renderers/lists.ts
+++ b/src/components/device/config-entry-renderers/lists.ts
@@ -1,6 +1,7 @@
 import { html, nothing } from "lit";
 import type { ConfigEntry } from "../../../api/types/config-entries.js";
 import { asMappingList, isPrimitiveOrNullish } from "../../../util/nested-values.js";
+import { escapeForInput, unescapeForInput } from "../../../util/yaml-escape.js";
 import { YamlRawValue } from "../../../util/yaml-serialize.js";
 import {
   effectiveDisabled,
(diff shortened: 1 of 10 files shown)


bdraco:
We have been stepping around adding `eemeli/yaml` as a dep and hand rolling everything but it seems like we might want to give in an do that, but its another large dep

bdraco:
Looks at it again and the conclusion is not to add it because esphome has so many special edge cases it can't parse anyways

PR #648, opened 2026-06-05
Hydrate component bodies for catalog-entry completion

bdraco:

The YAML editor builds its completion catalog from the slim `getComponents` index, which carries no `config_entries` (those hydrate lazily through `components/get_component_bodies`). `resolveAvailableEntries` read `config_entries` straight off that slim entry, got `undefined`, and threw on `.filter` / `.find` / spread for every component known to the index. So nested-key and value completion silently failed under any configured component; only the `fetchComponent` fallback for ids absent from the index ever returned fields.

This resolves the entries through `getComponentBodies` / `fetchComponent` (batched and session-cached) instead of trusting the index for `config_entries`. Nested-key and value completion work again under platform blocks and top-level components.

Found while debugging the missing id autocomplete in the YAML editor (esphome/device-builder#1230); this is the underlying completion breakage, separate from the new id-reference feature.

- fixes <link to issue>

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/util/yaml-completion.ts
+++ b/src/util/yaml-completion.ts
@@ -242,41 +242,55 @@ export async function resolveAvailableEntries(
   platformValue: string | null,
   topLevelKey: string | null
 ): Promise<ConfigEntry[]> {
-  // Special case: cursor is nested under a list-item header
-  // (``- platform: template`` → parentKey="platform"). The form
-  // fields the user wants live on the dotted catalog id
-  // ``<domain>.<platformValue>`` (e.g. ``binary_sensor.template``).
-  // Short-circuit either way — even on a miss, falling through
-  // would call ``fetchComponent(api, "platform")`` which 404s
-  // and poisons the session-scoped cache for the lifetime of
-  // the page (unlikely to matter today, but the failure mode
-  // would be silent if a real ``platform`` component ever
-  // shipped).
+  // The slim ``getComponents`` index carries no ``config_entries``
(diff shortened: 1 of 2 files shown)


src/util/yaml-completion.ts
```diff
@@ -242,41 +242,42 @@ export async function resolveAvailableEntries(
...
       }
-      // Try the dotted lookup — the catalog keys per-platform
-      // entries as ``<domain>.<stem>``.
       if (topLevelKey) {
-        const dotted = catalog.byId.get(`${topLevelKey}.${platformValue}`);
-        if (dotted) {
-          return [...directHit.config_entries, ...dotted.config_entries];
+        const dotted = `${topLevelKey}.${platformValue}`;
+        if (catalog.byId.has(dotted)) {
+          return [...own, ...(await entriesFor(dotted))];
```
bdraco: race here ?

esphbot: No race. `BatchedCache` dedups concurrent fetches via inflight map + microtask flush. Returned arrays are fresh local spreads. No shared state mutated here.

Real cost: sequential awaits. `await entriesFor(parentKey)` flushes its own batch, then the second `entriesFor` starts a new one. Two round trips, not one.

Fix: `Promise.all([entriesFor(parentKey), entriesFor(platformValue)])`. Both keys batch into one `get_component_bodies` call. Closes the interleaving window too.

PR #669, opened 2026-06-06
Give the add-secret dialog's password field an accessible name

bdraco:

Late follow-up to #668. The add-secret dialog wrapped `esphome-password-input` in a `<label>` with an external caption, but that custom element isn't a labelable control, so the inner password input had no accessible name and screen readers announced an unlabeled field. This passes the visible label text through the component's `label` and `placeholder` props, the same way the row renderer already does, so the inner input gets an `aria-label`.

The other late comment (clearing the add error on target change) was on a pre-merge commit; the merged code already recomputes the error conditionally, so nothing to change there.

- follow-up to #668

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


esphbot (approved):
No blocking issues found.

PR #675, opened 2026-06-06
Modernize makeRenderCtx fixture and retire inline RenderCtx literals

### LINKED ISSUE #674: Modernize makeRenderCtx fixture and retire inline RenderCtx literals in renderer tests
## Problem

The shared renderer-test fixture `makeRenderCtx` in `test/components/device/_renderer-fixtures.ts` has drifted from the current `RenderCtx` interface and only compiles because it returns `... as never`:

- It still exposes the old `getPendingNumeric` / `setPendingNumeric`, which are no longer on `RenderCtx`.
- It is missing `getEditingMagnitude` / `setEditingMagnitude` / `clearEditingMagnitude` (and `deviceName` / `pinRegistryModes`).

Because the cast hides these gaps, any test that uses `makeRenderCtx` and exercises a renderer needing the editing-magnitude callbacks (FLOAT_WITH_UNIT, TIME_PERIOD) would hit `undefined` at runtime. That's why several renderer tests hand-roll their own inline `RenderCtx` literals instead of using the fixture:

- `test/components/device/render-string-field-bails.test.ts`
- `test/components/device/render-string-field-secret-picker.test.ts`
- `test/components/device/render-nested-list-field.test.ts`
- `test/components/device/render-map-field.test.ts`

Those inline literals duplicate the same ~25-field ctx stub, so a new `RenderCtx` field has to be added in five places at once (most recently `showAdvanced` / `presentComponents` in #671).

## Proposed fix (small, standalone PR)

1. Bring `makeRenderCtx` up to date with the current `RenderCtx` (add the editing-magnitude callbacks, drop the stale pending-numeric ones) and **remove the `as never` cast** so the fixture is type-checked against `RenderCtx` going forward and can't silently drift again.
2. Migrate the four inline-ctx test files to `makeRenderCtx(values, { overrides })`, keeping each test's specific stubs (e.g. `emitChange` / `renderEntry` spies) as overrides.
3. Run the full suite — the float-with-unit and time-period tests are the ones to watch, since they depend on the editing-magnitude callbacks.

## Why separate

Discovered during the DRY review of #671. Keeping it out of that PR avoids risking the FLOAT_WITH_UNIT / TIME_PERIOD tests on a feature PR; this is pure test-infra cleanup.

bdraco:

The shared `makeRenderCtx` test fixture had drifted from the `RenderCtx` interface and only compiled behind an `as never` cast; it still carried the removed `getPendingNumeric` / `setPendingNumeric` and was missing the editing-magnitude callbacks, so a fixture-using test of a FLOAT_WITH_UNIT or TIME_PERIOD renderer would have hit `undefined` at runtime. This drops the stale fields, adds `getEditingMagnitude` / `setEditingMagnitude` / `clearEditingMagnitude`, and removes the cast so the fixture is type-checked against `RenderCtx` and cannot silently drift again.

Four renderer tests hand-rolled their own ~26-field `RenderCtx` literal to dodge the stale fixture, which meant every new `RenderCtx` field had to be added in five places at once; they now build on the fixture, keeping their spies and the map test's array-rejecting `getAt` as overrides. No production code changes; net 87 fewer lines.

- fixes https://github.com/esphome/device-builder-frontend/issues/674

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/test/components/device/_renderer-fixtures.ts
+++ b/test/components/device/_renderer-fixtures.ts
@@ -78,7 +78,7 @@ export function makeRenderCtx(
   } = {}
 ): RenderCtx {
   return {
-    localize: ((k: string) => k) as never,
+    localize: (k: string) => k,
     disabled: false,
     yaml: "",
     fromLine: 0,
@@ -109,8 +109,9 @@ export function makeRenderCtx(
     renderEntry: vi.fn(),
     getPendingUnit: () => undefined,
     setPendingUnit: vi.fn(),
-    getPendingNumeric: () => undefined,
-    setPendingNumeric: vi.fn(),
+    getEditingMagnitude: () => undefined,
+    setEditingMagnitude: vi.fn(),
+    clearEditingMagnitude: vi.fn(),
     // Stable per-fixture stash owner — tests that exercise the
     // templatable stash WeakMap (literal/lambda recovery) need a
     // single object identity across calls into the renderer. A
@@ -118,7 +119,7 @@ export function makeRenderCtx(
     // production form's "one stashOwner per host element" contract.
     stashOwner: {},
     ...(options.overrides ?? {}),
-  } as never;
+  };
 }
 
 /** Build a minimal ``ConfigEntry`` of *type* with sensible
--- a/test/components/device/_renderer-fixtures.ts
+++ b/test/components/device/_renderer-fixtures.ts
@@ -83,7 +83,7 @@ export function makeRenderCtx(
     yaml: "",
     fromLine: 0,
     sectionKey: "",
-    board: options.board ?? makeTestBoard(),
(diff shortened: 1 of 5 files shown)


esphbot (changes requested):

bdraco:
Fixed in b127832. The fixture's `board` option used `??`, so `board: null` fell back to the default ESP32 stub; the four migrated tests pass `board: null` and want a null board. Now gated on the key's presence: `board: "board" in options ? (options.board ?? null) : makeTestBoard()`, so an explicit null is preserved while an absent option still defaults to the stub. This addresses both koan's Important #1 and Copilot's low-confidence note on the same line. Verified the four migrated tests still pass (those renderers don't read `board`), full suite green (2519).

esphbot (approved):
No blocking issues found.

PR #682, opened 2026-06-07
Split config-entry-form styles to clear the size cap

esphbot:
## What does this implement/fix?

Splits `config-entry-form.styles.ts` (724 lines, over the repo's 500–600 line cap) into two sub-cap stylesheets, with no behaviour change.

- N/A (file-size-cap cleanup)

## How

- Cut the single `configEntryFormStyles` css template at a top-level rule boundary (after `.map-row .map-value-yaml-only`, before the render-error tile).
- Moved the second half into a new sibling `config-entry-form-extra.styles.ts` as `configEntryFormExtraStyles`.
- Composed it immediately after `configEntryFormStyles` in `fieldRendererStyles` (`config-entry-renderers-shared.ts`), so the concatenated rule order — and therefore the cascade — is identical to the original single sheet.
- Result: 368 / 364 lines, both under cap.

A pure CSS split: no logic, no API surface, byte-identical rules.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


bdraco (changes requested):
pr summary is mangled

bdraco (approved):

PR #687, opened 2026-06-07
Hydrate action catalog in script and api-action editors

### LINKED ISSUE #688: script / api-action editors can double-fire _loadAvailable on mount
### Summary

The script editor and the api-action editor can fire `_loadAvailable()` twice on mount, issuing two concurrent `getAvailableAutomations` + body-hydration round trips for the same configuration. Pre-existing and benign (last-writer-wins on identical data), but wasteful and racy. The automation editor already guards against this with a sequence token; the two trigger-less editors do not.

### Where

Both editors trigger the load from two places:

- `script-editor.ts:171` — `connectedCallback` → `_load()` → `_loadAvailable()` (when `configuration` is already set, `script-editor.ts:229`)
- `script-editor.ts:199-200` — `updated()` on a `configuration` change → `_loadAvailable()`
- `api-action-editor.ts:135` and `api-action-editor.ts:161-162` — same shape.

When `configuration` is set before the element connects, `connectedCallback`'s load and the first `updated()` both run, so two loads race to assign `_available`. In the common SPA path `configuration` is committed after connect (so `connectedCallback` skips on the empty default and only `updated()` fires), which is why this hasn't surfaced as a visible bug. But the path is order-dependent, not guaranteed single-fire.

Note: factoring the load through `loadAvailableFor` (the shared helper) widened the await window from one await to two (`getAvailableAutomations` + `hydrateAvailableBodies`), so the concurrent-load window is now larger than before.

### Reference fix

`automation-editor.ts` solves exactly this with a `_loadAvailableSeq` token (`automation-editor.ts:138,392,403,407,434`): bump a counter at the start of each load, and bail / skip assignment when the token is stale. Adopting the same guard in `script-editor.ts` and `api-action-editor.ts` (passing `isStale` to `loadAndHydrateAvailable` and assigning only when still current) would make the two simple editors deterministic and collapse the duplicate round trip. The existing mount tests should add a `getAvailableAutomations` call-count assertion to pin the single-load contract.

### Type

- refactor / maintenance (no user-visible bug today)

### LINKED ISSUE #1286: [Bug] No input field for logger text
### What happened?

When you add a log action, there is no way to specify the text, but the value is required.

Docs:
https://esphome.io/components/logger/#loggerlog-action

Simplified configuration:
```
# Board: Generic ESP32-S3 Board (Generic)
# Definition: definitions/boards/generic-esp32s3/manifest.yaml

esphome:
  name: repeat-test
  friendly_name: Repeat Test

esp32:
  variant: esp32s3
  framework:
    type: esp-idf

logger:

api:
  encryption:
    key: "crP2C6s9CjoW8mS6EaKrt7ond33MT/YCpzJNQnxvyQA="

ota:
  - platform: esphome

wifi:
  ssid: "!secret wifi_ssid"
  password: "!secret wifi_password"

script:
  - id: testscript
    mode: single
    then:
      - logger.log:

```

Screenshot:
Image

### Dashboard version

v0.1.0b100

### How did you install it?

Desktop app

### Where is it running?

Local — Windows

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

Actions added inside a `script:` block showed no input field for their values (e.g. `logger.log:` had nowhere to type the text). The same actions under `interval:` worked.

Cause: the action node skips rendering a form when an action has no `config_entries` (`automation-action-node.ts`), and `config_entries` aren't in the slim catalog; they're hydrated on demand via `/automations/get_bodies`. The automation editor hydrates; the script editor and api-action editor loaded only the slim catalog, so their actions arrived with empty `config_entries` and rendered fieldless. The api-action editor had the identical latent bug (it renders the same action list).

What changed:

- Both editors now hydrate the catalog before rendering action forms.
- The shared load + hydrate + outcome-mapping (and the partial-hydration toast) lives in one place; each editor keeps only a thin assign of its own reactive state.
- The load concurrency guard is structural: a `CatalogLoadController` (`ReactiveController`) owns the sequence token and is the only entry point, so an overlapping load (connectedCallback + updated) can't clobber `_available` or double-fire the toast, and the guard can't be omitted by a future caller. It also invalidates on host disconnect.
- Hydration is scoped to the lists an editor actually renders — the trigger-less editors fetch action + condition bodies only, not trigger bodies.

Verified with unit + happy-dom mount + controller tests, and end-to-end in the browser (the `logger.log` field now renders).

- fixes https://github.com/esphome/device-builder/issues/1286
- closes #688 (the double-fire / concurrency guard, fixed structurally here)
- follow-up: #689 (migrate the automation editor onto the same controller)

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/automation-editor/api-action-editor.ts
+++ b/src/components/device/automation-editor/api-action-editor.ts
@@ -45,7 +45,7 @@ import { registerMdiIcons } from "../../../util/register-icons.js";
 import "./automation-action-list.js";
 import { automationEditorStyles } from "./automation-editor.styles.js";
 import "./callable-params-editor.js";
-import { loadAvailableFor } from "./hydrate-available-bodies.js";
+import { CatalogLoadController } from "./catalog-load-controller.js";
 import { ParseErrorController } from "./parse-error-controller.js";
 import { applyYamlDiff, emptyAutomationTree } from "./serialise.js";
 
@@ -115,6 +115,11 @@ export class ESPHomeApiActionEditor extends LitElement {
   private _applyInFlight = false;
   private _applyDirty = false;
   private _lastSelfWrittenYaml: string | null = null;
+
+  /** Catalog loader; owns the concurrency guard so overlapping loads
+   *  (connectedCallback + updated both reaching ``_loadAvailable``)
+   *  can't clobber ``_available`` or double-fire the toast. */
+  private readonly _catalogLoad = new CatalogLoadController(this);
 
   /** Brief-window dirty flag covering the 200ms debounce gap so the
    *  global save button activates as soon as the user types. */
@@ -327,7 +332,8 @@ export class ESPHomeApiActionEditor extends LitElement {
     // Hydrates config_entries (the slim catalog omits them); without
     // it every action renders fieldless since the node bails on an
     // empty config_entries list.
-    const { available, error } = await loadAvailableFor(
+    this._error = "";
+    const { available, error } = await this._catalogLoad.load(
       this._api,
       this.configuration,
       this._localize
--- a/src/components/device/automation-editor/api-action-editor.ts
+++ b/src/components/device/automation-editor/api-action-editor.ts
@@ -45,6 +45,7 @@ import { registerMdiIcons } from "../../../util/register-icons.js";
 import "./automation-action-list.js";
 import { automationEditorStyles } from "./automation-editor.styles.js";
 import "./callable-params-editor.js";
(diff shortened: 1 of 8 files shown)


esphbot (changes requested):

esphbot (changes requested):

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #690, opened 2026-06-07
Add control-flow action param-rendering regression test (#1285)

bdraco:

Test-only follow-up for #1285 (no input field for a `repeat` action's required `count`). The root cause was the same slim-catalog hydration gap fixed in #687 — once `config_entries` are hydrated, the action node renders `repeat`'s `count`. This adds the missing regression coverage for control-flow actions.

`repeat` (and `if` / `while`) carry both a scalar param and a nested action list. The node renders the params form independently of the nested list, so a control-flow action must show BOTH. The test mounts a `repeat` node and asserts it renders its config-entry-form (the `count` field) alongside its nested `then` action list, guarding against a future change that suppresses params when children exist.

Verified end-to-end in the browser: with `repeat` inside a `script:`, the `count` input now renders.

- regression test for https://github.com/esphome/device-builder/issues/1285 (fixed by #687)

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/test/components/device/automation-editor/automation-action-node-controlflow.test.ts
+++ b/test/components/device/automation-editor/automation-action-node-controlflow.test.ts
@@ -1,0 +1,81 @@
+/**
+ * @vitest-environment happy-dom
+ *
+ * A control-flow action node renders its params form alongside its
+ * nested action list; children must not suppress the params form.
+ */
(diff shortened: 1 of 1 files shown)


esphbot (approved):
No blocking issues found.

PR #691, opened 2026-06-07
Migrate automation-editor onto CatalogLoadController

### LINKED ISSUE #689: Migrate automation-editor onto CatalogLoadController (unify the load concurrency guard)
### Summary

`#687` introduced `CatalogLoadController` (`src/components/device/automation-editor/catalog-load-controller.ts`), a `ReactiveController` that structurally owns the catalog-load concurrency guard (sequence token + host-disconnect invalidation) so a load can't clobber `_available` or double-fire the partial-hydration toast. The two trigger-less editors (`script-editor`, `api-action-editor`) now load exclusively through it.

`automation-editor.ts` solves the same concurrency problem with its own hand-rolled sequence guard (`_loadAvailableSeq` at `automation-editor.ts:138,392,407,434`). It is the only other catalog-form consumer; the add-dialogs correctly use the slim catalog (names only, no hydration). Unifying automation-editor onto the controller would give all three editors one structural guard and delete the bespoke seq dance.

### Why it's not in #687

automation-editor is not a drop-in:

- It hydrates **all** lists (it renders the trigger picker), so `load()` must take a `lists` selector (the controller currently hardcodes `["actions", "conditions"]`).
- It uses `loadAndHydrateAvailable({ onPaint })` to paint the slim catalog early and drop the spinner before hydration finishes, so `load()` must accept and forward an `onPaint` callback — wrapped in the controller's own staleness check so a stale early-paint can't land.
- It owns `_loading` lifecycle around the load.

That's a real refactor of the most complex (and currently well-tested) editor, kept out of the #687 bugfix to isolate risk.

### Proposed work

1. Generalize `CatalogLoadController.load()` to accept `{ lists?, onPaint? }`, forwarding to `loadAndHydrateAvailable` and guarding `onPaint` with the controller's sequence token.
2. Migrate `automation-editor._loadAvailable` to the controller; remove `_loadAvailableSeq`.
3. Keep the existing automation-editor mount tests green (single-load contract, early-paint, partial-hydration toast) and add controller coverage for the `onPaint` staleness wrap.

### Type

- refactor

bdraco:

automation-editor was the last catalog-form editor still rolling its own load-concurrency guard (the hand-rolled `_loadAvailableSeq` token). This moves it onto the shared `CatalogLoadController` from #687, so script-editor, api-action-editor, and automation-editor all share one structural guard.

The controller's `load()` now takes optional `{ lists, onPaint }`; `lists` defaults to actions plus conditions so the trigger-less editors are unchanged, and `onPaint` is wrapped in the controller's sequence token so a superseded early paint can't land. automation-editor passes all three lists since it renders the trigger picker, and keeps owning its own `_loading` lifecycle by driving it off the return shape. As a side effect it now gets host-disconnect invalidation it previously lacked.

Added controller coverage for the onPaint staleness wrap and the lists passthrough; existing automation-editor mount tests stay green.

- fixes https://github.com/esphome/device-builder-frontend/issues/689

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/automation-editor/automation-editor.ts
+++ b/src/components/device/automation-editor/automation-editor.ts
@@ -396,17 +396,13 @@ export class ESPHomeAutomationEditor extends LitElement {
       this.configuration,
       this._localize,
       {
-        // Hydrate all three lists — the editor renders the trigger
-        // picker, unlike the trigger-less editors (actions +
-        // conditions only).
+        // All three lists: this editor renders the trigger picker.
         lists: ["triggers", "actions", "conditions"],
-        // Paint the picker with the slim list AND drop the loading
-        // spinner so the dropdowns mount while hydration runs in the
-        // background; the controller guards this against a superseded
-        // load. The post-hydration ``available`` below carries fresh
-        // array refs so identity-based ``hasChanged`` consumers
-        // (trigger picker, condition tree, action node) re-render with
-        // the hydrated ``config_entries``.
+        // Paint the slim list and drop the spinner so the dropdowns
+        // mount while hydration runs; the controller guards this
+        // against a superseded load. The post-hydration ``available``
+        // below carries fresh array refs so identity-based
+        // ``hasChanged`` consumers re-render with the hydrated bodies.
         onPaint: (painted) => {
           this._available = painted;
           this._loading = false;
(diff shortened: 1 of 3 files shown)


esphbot (approved):
No blocking issues found.

PR #696, opened 2026-06-08
Preserve the !lambda tag through the lambda value sentinel

### LINKED ISSUE #1306: Editing a !lambda automation drops the !lambda tag and flattens multi-line bodies
### Summary

Editing an automation that contains a `!lambda` value corrupts it on round-trip; the `!lambda` YAML tag is dropped and a multi-line lambda body is flattened to one line. Dropping `!lambda` turns the C++ lambda into a plain literal string, which silently breaks the firmware (the source is sent as raw bytes instead of being executed).

### Reproduction

A `number.template` (or any component) with a `set_action:` containing an inline lambda:

```yaml
number:
  - platform: template
    name: Blockade Time
    id: Blockade_Time
    optimistic: true
    set_action:
      - uart.write: !lambda
          std::string blockades = "setInhibit " + str_sprintf("%.0f",x) + "\r\n";
          return std::vector<uint8_t>(blockades.begin(), blockades.end());
```

Parse it via `automations/parse` and re-render via `automations/upsert` (what the editor does on save of an action-list edit).

### Actual

```yaml
      - uart.write: |
          std::string blockades = "setInhibit " + str_sprintf("%.0f",x) + "\r\n"; return std::vector<uint8_t>(blockades.begin(), blockades.end());
```

Two regressions:

1. `!lambda` is replaced with a literal block scalar `|`; the tag is gone, so ESPHome treats the value as a string literal, not a lambda.
2. The two-line body is flattened onto one line.

### Expected

The `!lambda` tag is preserved and the body keeps its original line breaks; round-tripping an unedited automation is a no-op.

### Notes

- Confirmed against the parser/writer: the parsed tree captures the value as `{'data': {'_lambda': '...'}}` with the body already flattened (newlines joined with spaces) at parse time, and the renderer emits `|` rather than `!lambda`.
- Affects any automation action whose value is a `!lambda`, not just `set_action`; reproduced via a component action-field but the lambda emit path is shared.

bdraco:

Companion to esphome/device-builder#1310. The backend now carries a `!lambda` tag through the parse to upsert wire sentinel as an optional `_tag` key; this side preserves it so the user-facing edit flow keeps the tag instead of dropping it.

Three places that previously reconstructed a bare `{_lambda}` and lost the tag: the lambda editor now carries an existing `_tag` forward across body edits, the literal/lambda toggle sets `_tag: "!lambda"` (a templatable value field also accepts a literal, so the lambda form needs the explicit tag), and the YAML serializer emits `!lambda |-` when tagged. Untagged `lambda: |-` stays bare, so the common idiom is unchanged.

- fixes https://github.com/esphome/device-builder/issues/1306

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/api/types/automations.ts
+++ b/src/api/types/automations.ts
@@ -224,10 +224,13 @@ export interface YamlDiff {
  *  ``!lambda |- ...`` block from a literal string. Used wherever a
  *  ``ConfigEntry`` has ``templatable: true`` and the user picked the
  *  lambda branch of the literal/lambda toggle. The backend writer
- *  emits this as a ruamel ``LiteralScalarString`` with ``|-`` style;
- *  the parser inverts. */
+ *  emits an untagged ``_lambda`` as a bare ``|-`` block; ``_tag:
+ *  "!lambda"`` re-emits the explicit tag, which a templatable value
+ *  field (``uart.write:``, ``delay:``) needs to compile as a lambda
+ *  rather than a string literal. Echo it back unchanged on save. */
 export interface LambdaValue {
   _lambda: string;
+  _tag?: "!lambda";
 }
 
 /** Type guard for ``LambdaValue``. */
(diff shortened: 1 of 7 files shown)


esphbot (approved):
No blocking issues found.

PR #701, opened 2026-06-08
Declutter the device navigator and unify section icons

bdraco:

The device navigator left rail gets visually heavy on larger configs, so this is a first pass at calming it down. Rows are now a flat scannable list instead of bordered boxes; the per row chevron only appears on hover or selection; labels are lighter weight with tighter spacing so more fits before scrolling. The section headers gained the same icons the overview pane already uses (cog, chip, automations), so the two surfaces match. The editor header now shows the config filename next to the device name, and clicking "Device navigator" returns you to the device overview where "Change board" lives.

Motivated by Discord feedback that complex configs are hard to navigate; grouping by domain and a search or filter box are natural follow ups.

The section to icon mapping now lives in one module, `section-icons.ts`, consumed by both the navigator and the overview pane so they cannot drift.

- N/A (Discord feedback) https://ptb.discord.com/channels/429907082951524364/1513031639603871956

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/device-board-info.ts
+++ b/src/components/device/device-board-info.ts
@@ -1,10 +1,7 @@
 import { consume } from "@lit/context";
 import {
-  mdiArrowDecisionOutline,
   mdiArrowLeft,
   mdiClose,
-  mdiCogOutline,
-  mdiMemory,
   mdiOpenInNew,
   mdiPartyPopper,
   mdiPlusCircleOutline,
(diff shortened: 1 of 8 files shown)


esphbot (changes requested):

bdraco:
Addressed both review points in 9dd7133; the navigator title now nests the button inside the `<h2>` (heading owns the button, button holds an icon plus a `<span>`) so the markup is valid and the heading semantics are preserved, and `.editor-header-title` drops `flex-shrink: 0` and gains `min-width: 0` so its ellipsis engages, with the filename span set to `flex-shrink: 2` so it yields first when the header is tight.

esphbot (changes requested):

bdraco:
Addressed the label-in-name warning in bd9d67a; dropped the button `aria-label` so the visible "Device navigator" span text is the accessible name (WCAG 2.5.3), keeping `title="Back to device overview"` as the supplementary hint.

esphbot (approved):
No blocking issues found.

bdraco:
Applied the touch suggestion in 562e456; wrapped the hide-until-hover chevron in `@media (hover: hover)` so on touch devices (no hover) the per-row chevron stays visible, while pointer devices keep the decluttered reveal-on-hover behavior.

esphbot (approved):
No blocking issues found.

PR #702, opened 2026-06-08
Honor BACKEND_PORT in the dev server proxy

bdraco:

The dev server proxy target was hardcoded to `localhost:6052`, so testing a frontend against a backend on any other port meant editing `rspack.cjs` by hand. This honors a `BACKEND_PORT` env var (defaulting to 6052) for the `/ws`, `/api`, `/boards`, and legacy REST proxies, mirroring how `dev-server.cjs` already honors `PORT` for the frontend port. With both, two checkouts can run side by side; `PORT=5175 BACKEND_PORT=6060 npm run dev` points a second dev server at a second backend without touching tracked files. The startup banner now prints the resolved target instead of the literal 6052.

- N/A (dev tooling)

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/build-scripts/dev-server.cjs
+++ b/build-scripts/dev-server.cjs
@@ -1,4 +1,4 @@
-const { createRspackConfig } = require("./rspack.cjs");
+const { createRspackConfig, BACKEND_PORT } = require("./rspack.cjs");
 const { RspackDevServer } = require("@rspack/dev-server");
 const rspack = require("@rspack/core");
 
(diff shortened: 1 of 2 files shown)


bdraco:
Addressed in 072ce31; BACKEND_PORT is now parsed and validated once in rspack.cjs (parseInt plus Number.isFinite, falling back to 6052), the proxy targets interpolate that resolved value, and dev-server.cjs imports the same constant so the banner and the proxy can no longer diverge.

esphbot (approved):
No blocking issues found.

PR #703, opened 2026-06-08
Split dashboard styles to keep files under the size cap

esphbot:
## What
Split `dashboard/styles.ts` (1018 lines) into two CSS consts across two files to bring both under the 500-600 line cap.

## Why
`src/components/dashboard/styles.ts` had grown to 1018 lines — well over the repo's file-size cap (README → Code structure policies). It's a pure styles file with no logic, so the cleanest cut is to lift the overflow rules into a sibling file and keep composing them in the same `static styles` array.

## How
- Moved the lower half of the rules into a new `dashboard/styles-extra.ts` exporting `dashboardStylesExtra`.
- `dashboard.ts` composes `[espHomeStyles, inputStyles, dashboardStyles, dashboardStylesExtra]` — `dashboardStylesExtra` immediately follows `dashboardStyles`, so the concatenated sheet (and therefore the cascade order) is byte-identical to before.
- Split at a top-level rule boundary; verified the joined rule bodies diff-clean against the pre-split file. No visual or behavioural change.

## Types of changes

## Testing
- `npx tsc --noEmit` clean for all touched files (the lone pre-existing `@replit/codemirror-indentation-markers` error in `yaml-editor.ts` is unrelated install drift; CI installs cleanly).
- `npx vitest run test/components/dashboard/` → 67 passed.
- `prettier --write` on touched files.

Refactor only — no behaviour change.


bdraco (changes requested):
styles-extra is a bad name. This split is poorly designed

bdraco:
@esphbot rr

bdraco:
This still needs to be split by concern not extra

bdraco:
@esphbot rr

bdraco:
Superseded by #726, which redesigns the split by dashboard surface (styles.ts shell + toolbar; device-grid-styles.ts; yaml-mode-styles.ts; skeleton-styles.ts) rather than the styles-extra half-cut here. Closing in favor of that.

PR #706, opened 2026-06-08
Flag and offer to create missing !secret references in the structured editor

bdraco:

When a field references a secret that is not defined in secrets.yaml, the structured editor used to render it just like a valid one, masked dots and all, so there was no sign anything was wrong and no way to fix it; the YAML code view flagged it, but the visual editor did not.

The secret picker now detects when the referenced key is absent from the loaded secret list and flags it with a danger styled trigger and an alert icon. In place of the misleading masked value, it shows a short warning plus an inline field to create the secret; saving writes the value into secrets.yaml, the shared key cache refreshes, and the warning clears on its own, restoring the normal value reveal.

This is frontend only; it reuses the existing config/get_secrets list and the secrets.yaml write path, so no backend change is needed.

- N/A

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


esphbot (changes requested):

bdraco:
Thanks for the review; addressed in b413f43.

Important #1 (linked path leaves the warning stuck): the create/migrate flow now refreshes the secret keys cache on any successful ensure, not just when a write happened; ensureSecretInYaml still only fires secrets-saved on a real write, so the helper refreshes directly to cover the already-exists case on a stale cache. Both the create and migrate paths self heal now.

A11y: the inline value field is now an esphome-password-input (masked by default, reveal toggle, accessible name via its label prop), the warning carries role="alert", and the picker trigger gets a distinct aria-label in the missing state.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

esphbot (changes requested):

esphbot (changes requested):

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #707, opened 2026-06-08
Open Archived Devices from any route, including the editor

### LINKED ISSUE #1320: [Bug] “Archived Devices” cannot be accessed when the editor is active
### What happened?

When a device is open in the editor, you can still access all menu items from the hamburger menu except for “Archived Devices”. The menu item isn't grayed out, and there isn't an error message or anything like that; it just doesn't do anything.

Maybe a helpful message could be displayed here?

### Dashboard version

v0.1.0b104

### How did you install it?

Desktop app

### Where is it running?

Local — Windows

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

From the hamburger menu, "Archived Devices" did nothing while a device was open in the editor; every other item worked. The dialog was rendered inside the dashboard page and opened via a window event only the dashboard listened for, so once the editor route swapped the dashboard out, the trigger and the dialog were both gone.

This moves the archived-devices dialog to the app root and opens it with a bubbling `open-archived-devices` event, the same pattern the Settings, Firmware Jobs, and Feedback dialogs already use, so it works on every route. The dialog is now self-contained; it runs unarchive and delete (behind its own nested confirm) directly instead of delegating to the dashboard, so the dead dashboard wiring is removed.

- fixes https://github.com/esphome/device-builder/issues/1320

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/app-shell.ts
+++ b/src/components/app-shell.ts
@@ -80,6 +80,8 @@ import {
 } from "./app-shell/settings-actions.js";
 
 import "../pages/dashboard.js";
+import "./archived-devices-dialog.js";
+import type { ESPHomeArchivedDevicesDialog } from "./archived-devices-dialog.js";
 import "./command-palette.js";
 import "./esphome-layout.js";
 import "./esphome-login.js";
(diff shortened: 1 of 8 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
I think we should just hide it on non-main page.

PR #711, opened 2026-06-08
Use the normal UI font for terminal dialog titles and buttons

### LINKED ISSUE #1327: [Bug] Dialog titles and buttons using a monospace font
### What happened?

Is there a reason why the title bar and buttons in this dialog box are in a monospace font? It makes sense for the logs, of course, but I find it distracting elsewhere - it doesn't fit with the rest of the ESPHome Device Builder design.

Image

### Dashboard version

v0.1.0b104

### How did you install it?

Desktop app

### Where is it running?

Local — Windows

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

The Logs and Install dialogs rendered their title bar and the terminal toolbar buttons in a monospace font. Monospace is right for the log output, but on the title and buttons it looked out of place against the rest of the UI. This keeps monospace for the log content only; the title and the toolbar buttons now inherit the normal UI body font.

While here, the monospace font stack was duplicated across five rules; it is now a single `--term-mono-font` token in the shared terminal tokens, referenced by the status banner, queued overlay, sub-lines, and the logs transport chip.

- fixes https://github.com/esphome/device-builder/issues/1327

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/command-dialog/styles.ts
+++ b/src/components/command-dialog/styles.ts
@@ -27,7 +27,6 @@ export const commandDialogStyles = css`
     color: var(--esphome-on-primary);
     font-size: var(--wa-font-size-s);
     font-weight: var(--wa-font-weight-bold);
-    font-family: "SF Mono", "Fira Code", "Fira Mono", "Cascadia Code", monospace;
   }
   esphome-base-dialog::part(body) {
     padding: 0;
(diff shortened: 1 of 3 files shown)


esphbot (approved):
No blocking issues found.

PR #717, opened 2026-06-08
Render a lambda delay value in the editor instead of 0 Seconds

### LINKED ISSUE #1335: [Bug] The editor does not recognize a lambda as the value of a delay action
### What happened?

Maybe related to #1333, I'm not sure.

The editor does not recognize a lambda as the value of a delay action.

Simplified configuration:
```
substitutions:
  devicename: wf-tuersprechstelle
  friendly_devicename: Wf Türsprechstelle
  system_status_led_id: "id_status_led"
  system_status_led_brightness: "20%"

esphome:
  name: ${devicename}
  friendly_name: ${friendly_devicename}

esp32:
  board: esp32-s3-devkitm-1
  framework:
    type: esp-idf
  flash_size: 4MB

wifi:
  ssid: !secret wifi_ssid
  password: !secret wifi_password
  min_auth_mode: WPA2
  enable_btm: True
  enable_rrm: True
  output_power: 8.5

  # Enable fallback hotspot (captive portal) in case wifi connection fails
  ap:
    ssid: "${friendly_devicename} AP"
    password: !secret ap_wifi_password

captive_portal:

psram:
  mode: quad
  speed: 80MHz

# Enable logging
logger:

# Enable Home Assistant API
api:
  id: id_api
  encryption:
    key: !secret api_enc_key 

ota:
  - platform: esphome
    password: !secret api_ota_password

switch:

  - platform: gpio
    id: wf_tuerklingel_klingelsignal
    name: "Klingeln"
    disabled_by_default: true
    pin:
      number: 2
      inverted: true

number:

  - platform: template
    name: "Klingelsignal Dauer"
    id: id_klingelsignal_dauer
    optimistic: true
    device_class: duration
    mode: BOX
    unit_of_measurement: s
    min_value: 1
    max_value: 10
    step: 0.5
    initial_value: 2
    restore_value: true

script:

  - id: scr_klingelsignal_intern
    mode: single
    then:
      - repeat:
          count: 3
          then:
            - switch.turn_on: wf_tuerklingel_klingelsignal
            - delay: !lambda "return (int)(id(id_klingelsignal_dauer).state * 1000 / 3);"
            - switch.turn_off: wf_tuerklingel_klingelsignal
```

Screenshot:
Image

### Dashboard version

v0.1.0b105

### How did you install it?

Desktop app

### Where is it running?

Local — Windows

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

The Delay automation action is templatable, so `delay: !lambda "return ...;"` is valid, but the Visual Editor rendered it as `Value: 0` / `Unit: Seconds` and dropped the lambda the moment you touched the field. The bespoke delay renderer only ever drew the number + unit widget; a lambda arrives under `params.id` as a sentinel object, so the string-only shortcut path fell through to the empty default.

This adds a literal/lambda toggle to the delay renderer, matching the toggle the catalog-driven templatable fields already use. When the value is a lambda it shows the C++ editor with the body and round-trips the `!lambda` tag; a plain `delay: 2s` still renders as value + unit. Switching forms stashes the other side's body so an accidental toggle does not discard your work. Backend already parses and re-emits the tag, so this is editor-only.

- fixes https://github.com/esphome/device-builder/issues/1335

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/automation-editor/automation-action-node.ts
+++ b/src/components/device/automation-editor/automation-action-node.ts
@@ -31,7 +31,9 @@ import type {
   AvailableComponentInstance,
   AvailableScript,
   ConditionNode,
+  LambdaValue,
 } from "../../../api/types/automations.js";
+import { isLambdaValue } from "../../../api/types/automations.js";
 import type { BoardCatalogEntry } from "../../../api/types/boards.js";
 import type { LocalizeFunc } from "../../../common/localize.js";
 import { localizeContext } from "../../../context/index.js";
@@ -43,6 +45,8 @@ import { registerMdiIcons } from "../../../util/register-icons.js";
 import { renderAdvancedToggle } from "../advanced-toggle.js";
 import "../config-entry-form.js";
 import type { ConfigEntryValueChange } from "../config-entry-form.js";
+import "../config-entry-renderers/lambda-editor.js";
+import { lambdaBodyOf } from "../config-entry-renderers/lambda.js";
 import "./automation-condition-tree.js";
 import { automationEditorStyles } from "./automation-editor.styles.js";
 import "./catalog-picker-dialog.js";
@@ -143,6 +147,11 @@ export class ESPHomeAutomationActionNode extends LitElement {
   /** "Show advanced settings" gate for the action params form. */
   @state() private _showAdvanced = false;
 
+  /** Last lambda body typed for the Delay action, kept so flipping the
+   *  literal/lambda toggle back and forth doesn't discard the user's
+   *  C++ before they return to it. Reset when the action kind changes. */
+  @state() private _delayLambdaStash = "";
+
   static styles = [espHomeStyles, inputStyles, automationEditorStyles];
 
   /**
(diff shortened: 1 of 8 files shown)


esphbot (changes requested):

bdraco:
Thanks, both good catches.

Important #1 (literal lost on a literal to lambda to literal round trip): fixed in d126246. The stash is now two-sided; toggling to lambda captures the current value + unit via `_readDelay()` and toggling back restores it through `_writeDelay(value, unit)`, mirroring `templatable.ts`. Added a `seconds: "5"` to lambda to literal test that pins the restore.

Suggestion #2 (YamlRawValue block lambda): not reachable for delay here, so I left `_delayLambda()` on `isLambdaValue`. Automation action params come from the backend decompose (`controllers/automations/_decompose.py`), which renders a block `!lambda |-` to the `{_lambda}` sentinel and a tagged scalar to `{_lambda, _tag}`; it never emits a `YamlRawValue`. `YamlRawValue` is a config-entry-form raw-YAML construct, not an automation param shape, so `lambdaBodyOf`'s extra branch is just defensive breadth and can't be hit from `_renderDelayLambda` (its arg is already a `LambdaValue`).

While here I also pulled the literal/lambda toggle markup and CSS into a shared `literal-lambda-toggle.ts` so this renderer and the templatable field wrapper stop duplicating both.

esphbot (approved):
No blocking issues found.

PR #720, opened 2026-06-08
Render UNKNOWN config fields as YAML-only

### LINKED ISSUE #1328: [Bug] Invalid editor for ntc sensor
### What happened?

When adding a ntc sensor, the wizard marks calibration as a required field, but the editor only allows to enter a number. After entering a number just to proceed, an error is immediately raised.

Image

Image

Note: Just a head-up, calibration is marked as not supported when it is defined in the YAML editor, something to take into account when designing a solution

Image

### Dashboard version

Latest Beta

### How did you install it?

PyPI

### Where is it running?

Local — Linux

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

The backend emits `type: unknown` for a field it cannot model as a single control (a mapping or list union like the ntc sensor's `calibration`). The form dispatch sent `unknown` to the string input default, so the wizard showed a broken required text box. This routes `ConfigEntryType.UNKNOWN` to the existing YAML-only notice so the user edits it in the YAML pane, and skips its validation so a required one does not block the wizard with an error that cannot be cleared in the form.

- fixes esphome/device-builder#1328

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/config-entry-form.ts
+++ b/src/components/device/config-entry-form.ts
@@ -51,7 +51,11 @@ import {
   filterRenderable,
   renderFilterOptions,
 } from "./config-entry-render-filter.js";
-import { fieldKeyAttr, parseFieldKey } from "./config-entry-renderers-shared.js";
+import {
+  fieldKeyAttr,
+  parseFieldKey,
+  renderYamlOnlyField,
+} from "./config-entry-renderers-shared.js";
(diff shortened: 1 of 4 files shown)


esphbot (changes requested):

esphbot (approved):
No blocking issues found.

PR #722, opened 2026-06-08
Write secrets through the atomic config/set_secret command

### LINKED ISSUE #1334: Per-key secret writes race on the shared secrets.yaml (lost updates); add atomic config/set_secret
## Summary

Writing a single secret from the dashboard is a **whole-file read-modify-write of the shared `secrets.yaml`**, with no concurrency control. Two overlapping writers each read the same base text and each write back their own full version, so the last write wins and the other's change is silently lost.

Found while building the inline secret create/edit UX in the frontend: esphome/device-builder-frontend#706.

## The race

The frontend's secret-write helpers (`ensureSecretInYaml` / `setSecretInYaml`) do:

```
content = await api.getConfig("secrets.yaml")     // devices/get_config  (read whole file)
// …append or replace the one key in JS…
await api.updateConfig("secrets.yaml", content)    // devices/update_config (write whole file)
```

Interleaved writers (two browser tabs, the device editor racing the `/secrets` editor, or two quick saves) lose updates:

1. Tab A reads `secrets.yaml` v1, edits key `X`.
2. Tab B reads `secrets.yaml` v1, edits key `Y`.
3. A writes `v1 + X` → v2.
4. B writes `v1 + Y` → overwrites v2; **`X` is lost**.

`secrets.yaml` is a single shared file (one per config dir), so any per-key write through the generic whole-file path is exposed. The frontend's per-component `_busy` flag only serializes within one field instance — it doesn't coordinate across components, tabs, or clients.

## Proposed fix — `config/set_secret` under a lock

The backend already has the safe single-key primitive in `helpers/secrets_state.py`:
`_replace_or_append_secret` (line-based, inline-comment-preserving) + `atomic_write_file`, used today by onboarding's `set_wifi_credentials` (`write_wifi_secrets`).

- Add a WS command **`config/set_secret(key, value)`** that performs the read-modify-write **under an `asyncio.Lock`**, reusing `_replace_or_append_secret`. The backend runs a single event loop, so the lock makes the whole read-modify-write atomic and eliminates the interleave above. (A `config/delete_secret` could follow the same shape if needed.)
- Frontend: switch `ensureSecretInYaml` / `setSecretInYaml` to call `config/set_secret` instead of `devices/get_config` + `devices/update_config`, dropping the whole-file round-trip. (`created` vs overwrite can be surfaced in the response if the linked/created toast distinction is still wanted.)

Residual: two *separate* backend processes sharing one config tree (e.g. two addon instances) would still race — rare; an OS-level file lock around the write would close that too if we decide it matters.

## Repro

1. Open the same device's secret-eligible field in two tabs (or the device editor + `/secrets`).
2. In tab A, set secret `foo`; in tab B (without reloading), set secret `bar`.
3. Whichever saves second wins; the other key is missing from `secrets.yaml`.

bdraco:

`ensureSecretInYaml` / `setSecretInYaml` wrote one secret by reading the whole `secrets.yaml` over `devices/get_config`, mutating one key in JS, and writing the whole file back over `devices/update_config`. Two overlapping per-key writes (two tabs, the device editor racing `/secrets`, two quick saves) each read the same base and wrote back their own full version, so the last write silently dropped the other key.

Both helpers now go through the backend's atomic `config/set_secret(key, value, overwrite)` command, which does the read-modify-write under a lock; `ensure` passes `overwrite=false` so the create-if-absent existence check happens server-side instead of as a racy check-then-set here, and keeps the `{created}` result that drives the created-vs-linked toast. Drops the now-dead JS YAML mutation helpers, and masks the secret `value` in the debug `[SENDING]` log. The whole-file `/secrets` editor still uses `getConfig`/`updateConfig`; only per-key writes move.

Needs the companion backend change that adds the command (it ships prebuilt in the wheel).

- fixes https://github.com/esphome/device-builder/issues/1334

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/api/esphome-api.ts
+++ b/src/api/esphome-api.ts
@@ -100,6 +100,7 @@ function redactForLog(payload: unknown): unknown {
   // all of them should have their secrets redacted before logging.
   const containsSecret =
     (command !== null && command.startsWith("auth")) ||
+    command === "config/set_secret" ||
     "token" in obj ||
     "password" in obj ||
     (result !== null && ("token" in result || "password" in result));
@@ -112,6 +113,10 @@ function redactForLog(payload: unknown): unknown {
     const safeArgs: Record<string, unknown> = { ...args };
     if ("token" in safeArgs) safeArgs.token = "<redacted>";
     if ("password" in safeArgs) safeArgs.password = "<redacted>";
+    // config/set_secret carries the secret in ``value``.
+    if (command === "config/set_secret" && "value" in safeArgs) {
+      safeArgs.value = "<redacted>";
+    }
     clone.args = safeArgs;
   }
   if (result !== null) {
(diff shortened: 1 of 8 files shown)


esphbot (approved):
No blocking issues found.

PR #724, opened 2026-06-08
Re-lint the open YAML editor on a secrets save

### LINKED ISSUE #1332: YAML editor lint stays stale after secrets.yaml (or any included file) write
## Summary

After a secret is written while the YAML editor is open (inline create/edit from the secret picker, or the secrets editor), the code editor keeps showing the stale `Secret '<name>' not defined in "<file>"` squiggle on the `!secret` line. It only clears once the 60s validate cache expires or the device YAML is edited (which changes the cache key). The structured editor already self-heals; the YAML lint does not.

Found while building the missing-`!secret` UX in the frontend: esphome/device-builder-frontend#706.

## Root cause

`EditorController.validate_yaml` caches the result per session keyed **only on the device-YAML content hash** (`controllers/editor.py`, `_CachedValidation` / `_VALIDATE_CACHE_TTL = 60.0`):

```python
content_hash = fnv1a_32(content.encode("utf-8"))
cached = session.cached
if cached is not None and cached.is_fresh_for(content_hash):
    return cached.result
```

Writing `secrets.yaml` (or any `!include` / `external_components` source) does not change the device YAML, so the hash is unchanged and the cached `{validation_errors: [...]}` is returned for up to 60s. The warm `esphome vscode --ace` subprocess would read the fresh `secrets.yaml` if re-run, but the cache short-circuits it. So even forcing a client re-lint returns the stale result.

This is a latent correctness bug beyond secrets: editing any externally-referenced file (`!include`d fragment, `packages:`, `external_components`) leaves dependent device validations stale for the TTL.

## Proposed fix (coordinated, two repos)

**Backend (this repo):** invalidate the editor session validate cache when a config file is written. Cleanest hook is the write path that persists `secrets.yaml` / device YAML (the `editor`/config write command behind `updateConfig`): clear `session.cached` for all `_EditorSession`s (a referenced file can affect any open device's validation). Add an `EditorController.invalidate_cache()` and call it on any config-dir write. Keep it coarse (clear all) — validations are cheap to recompute and correctness beats a 60s stale window.

**Frontend (device-builder-frontend):** on the `secrets-saved` window event (already dispatched by the secret write path), force the open editor to re-lint (`forceLinting(view)` from `@codemirror/lint`) so the now-invalidated result is recomputed promptly instead of on the next keystroke.

## Repro

1. Open a device whose YAML has `api: { encryption: { key: !secret foo } }` where `foo` is absent from `secrets.yaml`.
2. The code editor squiggles `Secret 'foo' not defined`.
3. Create `foo` via the secret picker's inline create (or add it in the secrets editor) — the structured editor clears.
4. The code editor keeps the squiggle for ~60s despite `foo` now being defined.

bdraco:

After a secret is written while the YAML editor is open (secret picker inline create, secrets editor), the code editor kept the stale `Secret '<name>' not defined` squiggle until the lint cache expired. A secrets.yaml write doesn't change the editor doc, so the lint plugin has nothing scheduled and `forceLinting(view)` alone is a no-op. This adds a `relintEffect` plus a `needsRefresh` hook on the backend linter; on the `secrets-saved` window event the editor dispatches that effect and forces a re-lint, so the now-invalidated backend result is recomputed promptly instead of on the next keystroke.

Pairs with the backend cache invalidation that makes the recomputed result fresh.

- fixes esphome/device-builder#1332

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/yaml-editor.ts
+++ b/src/components/yaml-editor.ts
@@ -1,6 +1,7 @@
 import { autocompletion } from "@codemirror/autocomplete";
 import { indentWithTab, undoDepth } from "@codemirror/commands";
 import { indentUnit } from "@codemirror/language";
+import { forceLinting } from "@codemirror/lint";
 import { EditorState, StateEffect, StateField } from "@codemirror/state";
 import { Decoration, keymap, type DecorationSet } from "@codemirror/view";
 import { consume } from "@lit/context";
@@ -30,6 +31,7 @@ import { createYamlHoverTooltip } from "../util/yaml-hover.js";
 import {
   createBackendYamlLinter,
   lintErrorLineGutter,
+  relintEffect,
 } from "../util/yaml-lint-backend.js";
 import type { YamlSection } from "../util/yaml-sections.js";
 import {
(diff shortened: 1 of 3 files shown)


esphbot (approved):
No blocking issues found.

PR #726, opened 2026-06-08
Split dashboard styles into four concern files

bdraco:
## What
Split `dashboard/styles.ts` (1018 lines) into four concern-named files, each a coherent dashboard surface, all composed in order into the dashboard host's `static styles`.

## Why
`src/components/dashboard/styles.ts` was well over the file-size cap. This supersedes #703, which cut the file mechanically in half into a `styles-extra.ts`; per the review on that PR, the name was poor and the split was arbitrary. This repo already organizes dashboard styles by surface (`table-styles.ts`, `table-cell-styles.ts`, `skeletons.ts`), so the split follows that convention.

## How
- `styles.ts` -> `dashboardStyles`: shell layout, search toolbar, view toggle, empty-search state.
- `device-grid-styles.ts` -> `deviceGridStyles`: discovered and configured device grids, plus the create-device affordances (add card, table create button, FAB).
- `yaml-mode-styles.ts` -> `yamlModeStyles`: the YAML search mode hit list.
- `skeleton-styles.ts` -> `skeletonStyles`: card and table loading skeletons.
- `dashboard.ts` composes `[espHomeStyles, inputStyles, dashboardStyles, deviceGridStyles, yamlModeStyles, skeletonStyles]`.

Every rule stays in the dashboard host's shadow DOM; nothing moved to other components.

## Behaviour-free, verified
Concern grouping reorders sections relative to the original single file, so I verified the cascade is unaffected:
- The rule text is identical; the set of rule blocks across the four files is a byte-exact multiset of the original (only order changed, nothing edited or dropped).
- No selector appears in more than one file, so source-order between files cannot change any cascade outcome (order only decides between equal-specificity rules targeting the same element).
- `npx vitest run` 2730 passed; `npm run lint` and `npm run build` clean; each file is comfortably under the cap (412 / 355 / 163 / 103).
- Browser smoke test: the dashboard renders with the six composed stylesheets and computed styles for elements from each new file resolve correctly.

## Types of changes

## Testing
- `npx tsc --noEmit` clean.
- `npx vitest run` -> 2730 passed.
- `npm run build` clean.
- `prettier` clean on touched files.

Supersedes #703.


esphbot (approved):
No blocking issues found.

bdraco:
@esphbot addressed in 5c82968. Moved the `.device-count` / `.device-count-row` / `.table-device-count-row` block out of `yaml-mode-styles.ts` and into `styles.ts` under a "Device count row" divider in the toolbar concern, where its consumers (`render-toolbar.ts`, `render-content.ts`) live; they were only grouped with the YAML hit list because of their contiguous position in the original file. Also rebased onto main to pick up #724. Re-verified: every meaningful CSS line is still a byte-exact multiset of the original, no selector spans two files, 2732 tests pass, all four files under the cap (448 / 355 / 129 / 103).

esphbot (approved):
No blocking issues found.

PR #731, opened 2026-06-09
Map navigator glyphs for ~250 top-level domains, aligned to the docs icons

bdraco:

Follow-up to the navigator subgroup glyphs. The previous pass left most top-level component domains on the neutral `shape-outline`, so common ones a user actually configures (`runtime_stats`, `psram`, `usb_host`, `usb_uart`, `zwave_proxy`, and many more) still showed the generic shape.

This fills `DOMAIN_ICON` for ~250 top-level domains, grouped into families: the rest of the bluetooth family, board platforms, USB, mesh radios (zwave / zigbee / thread), power / memory / lifecycle, sub-GHz and IR radios, NFC / RFID readers, presence radars, touch / keypad, audio / voice, LED drivers, ADC / DAC, IO and bus expanders, energy meters, and graphics.

Icon picks follow the esphome-docs per-component SVG, which is the same image the component **header** shows (the catalog `image_url`), so the navigator subgroup glyph now matches the header for these components. Two deliberate departures from the docs:

- Platform domains (`sensor`, `light`, `switch`, …) keep their semantic glyph (gauge, lightbulb, …) instead of the docs' generic `folder-open` placeholder.
- A few specific glyphs are kept over a weaker docs pick, e.g. `zwave_proxy` stays the distinct `z-wave` icon rather than the docs' generic `server-network`.

Platform components (e.g. `esp32_rmt_led_strip`, `dht`) still inherit their domain's glyph and need no entry. Only `demo`, `script`, and a couple of obscure components fall through to the neutral shape.

- n/a (reported via screenshots: common components showing the generic shape in the navigator)

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


esphbot (approved):
No blocking issues found.

PR #728, opened 2026-06-09
Fill navigator subgroup glyphs for the bluetooth family and other top-level domains

bdraco:
# What does this implement/fix?

The navigator groups Components (and Core) by domain and shows a per-domain glyph on each subgroup header. The glyph comes from `DOMAIN_ICON` in `navigator-row-icons.ts`, which didn't list `bluetooth_proxy`, so it fell back to the neutral `shape-outline` even though the mapped sibling `esp32_ble_tracker` and the component detail header (catalog `bluetooth.svg`) both show Bluetooth; the nav and header disagreed for the same component.

Auditing the map against the catalog's top-level component domains showed the same gap for the rest of the Bluetooth family and a set of other common top-level keys. This fills them in:

- Bluetooth family: `bluetooth_proxy`, `ble_client`, `ble_nus`, `esp32_ble_beacon`, `esp32_ble_server`
- Board platforms (head the Core list on every device): `esp32`, `esp8266`, `rp2040`, `bk72xx`, `rtl87xx`, `libretiny`, `nrf52`, `host`
- Common components: `mqtt`, `i2s_audio`/`microphone`/`micro_wake_word`/`voice_assistant`, `improv_serial`/`esp32_improv`, `remote_receiver`/`remote_transmitter`, `deep_sleep`, `status_led`, `wireguard`, `http_request`, `lvgl`, `font`/`image`/`online_image`/`animation`
- Top-level keys that look like platforms but own their YAML block: `esp32_camera`/`esp32_camera_web_server`, `syslog`, `modbus`/`modbus_controller`/`modbus_server`

Platform components (e.g. `esp32_rmt_led_strip`, `dht`, `gpio`) keep grouping under their already-mapped domain (`light`/`sensor`/...), so they inherit the domain glyph. The niche long tail stays on the neutral `shape-outline` by design.

**Related issue or feature (if applicable):**

- n/a (reported via screenshots: Bluetooth Proxy navigator glyph showed a generic shape)

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [ ] New feature (non-breaking change which adds functionality) — `new-feature`
- [ ] Enhancement to an existing feature — `enhancement`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [ ] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `docs`
- [ ] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
[...]

PR #732, opened 2026-06-09
Show a leading domain glyph on ungrouped navigator rows

bdraco:

Polishes the device YAML navigator so flat (ungrouped) rows are as scannable as the grouped Components subgroups, plus several follow-on fixes from screenshot feedback. Follow-up to #728 / #731.

- **Leading domain glyph on ungrouped rows.** `renderNavRow` takes a `showIcon` flag: Core and the flat Automations list get a leading `iconForDomain` glyph; grouped Component rows stay glyph-free since their subgroup header already carries it. The trailing chevron moved to its own class so hover-declutter stays scoped to it.
- **Collapse single-of-a-kind domains.** A Components domain with only one component rendered a collapsible header guarding one row; those now render as a flat row with the domain glyph and component name, glyph aligned in the same column as the multi-item subgroup headers. Multi-item domains keep their header; while filtering, a lone match keeps its header for domain context.
- **ESPHome logo for the core row.** The esphome core row shared esp32's chip glyph; it now uses a monochrome `currentColor` ESPHome mark (new `esphome-mono.svg`) so it mutes to the same tone as the other glyphs while staying distinguishable.
- **Domain hover tooltip.** Row glyphs get a `title` naming the domain, the only domain cue on a flat or automation row.
- **Automation label rework.** Component-bound and action-field automation rows lead with the trigger/action (so `turn_on_action` vs `turn_off_action` stay distinct), with the instance on line 2; core-name trimming drops the redundant "Component" / "Configuration" suffix.
- **Scalar parsing.** `readInstanceScalar` strips a quote-aware inline comment, and treats a comment-only value (`name: # c`) as empty.
- Trim the navigator row left inset, and add the `script` domain icon mapping.

- n/a (follow-up to #728 / #731; reported via screenshots of the Core configuration list)

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


esphbot (approved):
No blocking issues found.

bdraco:
On the implicit `clock-outline` registration suggestion: resolved in 2e5ad36 by dropping `iconForRowKey` and `AUTOMATION_ICONS` entirely. Automation rows now resolve through `iconForDomain(item.parentKey ?? item.key)` against the single `DOMAIN_ICON` map, so the returned name and its registration come from the same source by construction:

```
registerMdiIcons(Object.fromEntries([...Object.values(DOMAIN_ICON), FALLBACK]));
// ...
return (DOMAIN_ICON[domain] ?? FALLBACK)[0];
```

`script` is now a real `DOMAIN_ICON` entry, and interval keeps `clock-outline` from the existing `interval` entry — there's no longer a resolver that can return a name outside the registered set, so the silent-unregister failure mode the suggestion flagged can't happen.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

bdraco:
Thanks, both addressed:
- `readInstanceScalar` now treats a comment-only value (`name: # c`) as empty (b83591f). Kept the fix scoped to the instance-scalar reader rather than `splitInlineComment`, since the shared splitter feeds the serializer's comment-preserving round-trip.
- Updated the PR description to cover the singleton-domain collapse and the other follow-on changes (logo, tooltip, padding, automation labels) so it no longer reads as render+style only.

esphbot (approved):
No blocking issues found.

bdraco:
Fixed the orphaned JSDoc in b41d606: `renderNavRow`'s doc now sits on its own function, with `renderRowGlyph`'s doc above `renderRowGlyph`.

esphbot (approved):
No blocking issues found.

PR #733, opened 2026-06-09
Fix navigator section lock-out after YAML cursor reveal

bdraco:

Fixes a regression from #729 (navigator "reveal selected"): after clicking a line in the YAML editor, the navigator would expand the cursor's section but then refuse to let you open any other section; the clicked section snapped back open on every render.

The reveal controller re-dispatched `section-reveal` on every host update until it confirmed a scroll latch. When the selected row never became visible, a component inside a collapsed subgroup, or the hidden second navigator instance that the page also renders, that latch never set, so the next render, including the one triggered by the user clicking a different section header, re-forced the cursor's section open. The user was locked out of toggling anything else.

The fix makes the `section-reveal` dispatch one-shot per selected line via a separate `_revealedLine` latch, so a manual re-close of that section no longer re-triggers the forced open. The deferred scroll retry is left intact, so a row that becomes scrollable later (collapsed subgroup expanded) still scrolls into view.

Verified empirically with Chrome/CDP against a real device: pre-fix, clicking Core after the cursor revealed Automations did nothing (Automations stayed force-open); post-fix, Core opens and stays open across re-renders.

- fixes <link to issue>

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/navigator-reveal-controller.ts
+++ b/src/components/device/navigator-reveal-controller.ts
@@ -59,6 +59,13 @@ export class NavigatorRevealController implements ReactiveController {
       this._revealedLine = null;
       return;
     }
+    // The selection genuinely moved to a different line: drop the stale latch
+    // so a line whose reveal never scroll-latched (collapsed subgroup) can
+    // reveal again when the user clicks back to it. Same-line re-renders keep
+    // the latch, which is what prevents the toggle lock-out / snap-back.
+    if (this._revealedLine !== null && this._revealedLine !== selectedLine) {
+      this._revealedLine = null;
+    }
     if (selectedLine === this._scrolledLine) return;
     const index = sectionIndexForLine(buckets, selectedLine);
     if (index === -1) {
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
Addressed the suggestion in 55533da: the `_revealedLine` latch now clears when the selected line genuinely changes, so returning to a line whose reveal never scroll-latched (collapsed subgroup) reveals it again; same-line re-renders keep the latch, which is what prevents the lock-out and snap-back. Added a test pinning re-reveal after the selection moves away and returns.

esphbot (approved):
No blocking issues found.

bdraco:
Good catch narrowing it to the `index === -1` path. That exact path is already closed by 55533da: the latch reset runs at the top of `hostUpdated` (before the `index === -1` early return), so an intervening unscoped line clears `_revealedLine` and returning to a still-closed section A re-reveals. This is your suggested "clear it whenever selectedLine changes" option. Added 0395c4e, a focused controller test driving the `index === -1` intervening case directly; it fails before the reset and passes after.

esphbot (approved):
No blocking issues found.

PR #737, opened 2026-06-09
Move first-paint state writes to willUpdate

bdraco:

On first load of the device editor, four components wrote reactive state from inside `updated()`. That callback runs after the render commits, so on first paint (when every initialized property is in `changedProperties`) each write scheduled a second, wasted render; Lit logged a `change-in-update` warning for each. Moving the first-paint writes into `willUpdate` folds them into the in-progress render, so both the warnings and the extra passes are gone. The DOM-dependent work stays in `updated()`, since `device-board-info` still drives its child editors through `@query` refs that exist only after render.

Verified with a headless before/after: the four warnings (`command-palette`, `device-editor`, `device-board-info`, `device-section-config`) go from one each to zero.

- N/A; console-only warning, no filed issue

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/command-palette.ts
+++ b/src/components/command-palette.ts
@@ -324,9 +324,6 @@ export class ESPHomeCommandPalette extends LitElement {
 
   protected willUpdate(changed: Map<string, unknown>) {
     if (changed.has("_open")) this._escape.set(this._open);
-  }
-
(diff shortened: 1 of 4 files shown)


esphbot (approved):
No blocking issues found.

PR #738, opened 2026-06-09
Give core navigator rows specific glyphs instead of the generic cog

bdraco:

The Core configuration list leaned on two generic glyphs; a cog for substitutions, packages, globals, external components, json, byte buffers and the hash helpers, and a chip for every 32 bit board platform. This gives each a meaningful icon so a long list stays scannable by shape. Substitutions now show code braces, packages a box, globals a variable, external components a puzzle, json the code json glyph, byte and split buffers a code array, sha256 and the hmac helpers a pound box, and the esp32, esp8266 and libretiny family a 32 bit cpu. Preferences also moves off the memory glyph, which stays with psram, to a save settings glyph.

- N/A

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


esphbot (approved):
No blocking issues found.

PR #739, opened 2026-06-09
Let the dependency-detour banner flow as inline text

bdraco:

The "Adding a dependency. We'll bring you back to X after." banner in the add-component dialog rendered poorly when the sentence was wider than the dialog. `.return-banner` was `display: flex`, so the prefix text, the bold component name, and the suffix text each became a separate flex item; instead of wrapping as one sentence, flex-shrink split them into three mis-aligned columns. Dropping the flex (and the `align-items` / `gap` that only applied to it) makes the banner a plain block, so the sentence wraps normally with the name inline. The sibling `.bundle-banner`, which needs flex for its leading icon, is unchanged.

- N/A; reported from the dashboard, no filed issue

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/add-component-dialog.styles.ts
+++ b/src/components/device/add-component-dialog.styles.ts
@@ -19,9 +19,6 @@ export const addComponentDialogStyles = css`
      "add a dependency" mid-way through adding another component.
      Tells them we'll bring them back to the original after. */
   .return-banner {
-    display: flex;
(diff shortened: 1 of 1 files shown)


esphbot (approved):
No blocking issues found.

PR #745, opened 2026-06-10
Hide the green lock for all encrypted devices in compact views

### LINKED ISSUE #1369: [Bug] Green lock icon not displayed sometimes in device list
### What happened?

The green locker that indicates "api traffic encrypted" doesn't appear for some devices, while looking at the details it shows its encrypted: 

Image

looking to "cortina-ar" details we can see the icon. Status should match

Image

### Dashboard version

main branch - cb76150

### How did you install it?

From source (script/setup)

### Where is it running?

Local — Linux

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

The green encryption lock showed in the device detail drawer but not in the device list for confirmed-encrypted devices, so the two views disagreed. The list and the drawer were deriving the lock through separate code paths, and the list only hid the lock for mDNS-confirmed devices; YAML-encrypted-but-unconfirmed ones still showed it, which read as random.

Now the compact views (table rows, cards) hide the lock for every encrypted (active) device, so the list shows only attention states (plaintext, pending, mismatch). The drawer still shows the full encrypted badge for single-device inspection, and it now reads from the same VISUALS source the list uses so the two can't drift apart again.

- fixes https://github.com/esphome/device-builder/issues/1369

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/dashboard/device-drawer-content/render-sections.ts
+++ b/src/components/dashboard/device-drawer-content/render-sections.ts
@@ -1,6 +1,10 @@
 import { html, nothing, type TemplateResult } from "lit";
 import type { ConfiguredDevice } from "../../../api/types/devices.js";
 import type { LocalizeFunc } from "../../../common/localize.js";
+import {
+  getEncryptionVisual,
+  type EncryptionState,
+} from "../../../util/encryption-state.js";
 import { formatFileSize } from "../../../util/format-file-size.js";
 import { splitIntegrations } from "../../../util/integration-split.js";
 import { buildWebUiUrl } from "../../../util/web-ui-url.js";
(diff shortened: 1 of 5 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
Tightened `badgeClass` to a literal union; it's a CSS class, so that matches the `cssClass` precedent and lets tsc catch a class typo. Left `labelKey` (and the adjacent `tooltipKey`) as `string` on purpose; those are i18n keys, and the whole codebase types localization keys as plain string (`LocalizeFunc` takes `key: string`), so a union there would be the inconsistent one.

esphbot (approved):
No blocking issues found.

PR #747, opened 2026-06-10
Collapse secrets editor split view into a Form/YAML toggle

### LINKED ISSUE #136: Changes to secrets file editor
- [ ] What is this different view ?? - change the view back to the intended design https://github.com/esphome/device-builder-frontend/pull/747
- [ ] Add secret button shouldn't not be dashed https://github.com/esphome/device-builder-frontend/pull/747

bdraco:

Drops the side by side split mode from the secrets editor. The structured form and the YAML editor are now a simple two way toggle, since there is no reason to view both at once. The Add secret button used a dashed outline; it now matches the other filled buttons in the UI.

The YAML editor stays reachable so users can always fall back to editing raw YAML.

There are a few concerns we need to think about.

- We need to make secrets easy to do per device so users can easily follow https://esphome.io/guides/security_best_practices/ which is what the structured editor does. We want them to have a separate encryption key per device, and unique passwords per device so a singled compromised device does not lead to all devices being compromised.

- We will have some very unhappy users if they can't fallback to edits in YAML so ripping out YAML completely will generate some major community push back.

- fixes https://github.com/esphome/backlog/issues/136

Button adjusted
Screenshot 2026-06-10 at 10 43 57 AM

Split view dropped as there is no reason they need to see both at once
Screenshot 2026-06-10 at 10 44 03 AM

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/secrets/secrets-structured-editor.styles.ts
+++ b/src/components/secrets/secrets-structured-editor.styles.ts
@@ -6,9 +6,11 @@ export const secretsStructuredEditorStyles = css`
     height: 100%;
     box-sizing: border-box;
     overflow-y: auto;
-    /* This element is the single scroll container for the form pane, so
-       the clearance for the floating Save button has to live here — the
-       wrapping pane never scrolls. */
+    /* Reserve the gutter so the scrollbar clears the row controls. */
+    scrollbar-gutter: stable;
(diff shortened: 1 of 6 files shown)


stvncode (approved):
This better thanks 🙏

PR #749, opened 2026-06-10
Tokenize the shared action-button padding

bdraco:

The primary action button padding `8px 18px` was repeated as a literal in the shared `.btn` rules and in every dialog's action row. This adds an `--esphome-button-padding` token and points all of them at it, so the value lives in one place. The 8px reuses `--wa-space-xs`; the 18px has no spacing-scale token, so it stays inline inside the token.

No visual change at the default root font size; the computed padding matches the prior 8px 18px literal. The 8px reuses --wa-space-xs, so under a non-default root font size the vertical axis scales while the 18px stays fixed.

- n/a

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/adopt-dialog.ts
+++ b/src/components/adopt-dialog.ts
@@ -171,7 +171,7 @@ export class ESPHomeAdoptDialog extends LitElement {
       }
 
       .btn {
-        padding: 8px 18px;
+        padding: var(--esphome-button-padding);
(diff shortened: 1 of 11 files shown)


esphbot (approved):
No blocking issues found.

PR #752, opened 2026-06-10
Group the select-multiple toggle with the device count

### LINKED ISSUE #141: select multiple is not aligned in card view, it's in a weird place in table view
split from https://github.com/esphome/backlog/issues/135

bdraco:

The select-multiple toggle used space-between, so it floated against the far edge of the count row; in card view it sat unanchored opposite the search box, in table view it split across the width away from the count. It is now grouped next to the device count, left-aligned under the search box, so the placement reads the same in card and table views and on mobile.

While reviewing, the toolbar rows also had uneven vertical spacing on mobile; the column gap was 2px while the control-row wrap gap was larger, so the search-to-Filters gap did not match the Filters-to-count gap. Both now share one token, so every stacked toolbar row has an even gap.

Trade-off: grouping the toggle on the left is slightly harder to thumb-reach on mobile than the old far-right spot, but in practice it is minor; the previous layout was worse when the controls reflowed, so it stays grouped for the cleaner, consistent look.

- fixes esphome/backlog#141

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

## Screenshots

Screenshot 2026-06-10 at 11 39 04 AM
Screenshot 2026-06-10 at 11 39 14 AM

### DIFF
--- a/src/components/dashboard/render-toolbar.ts
+++ b/src/components/dashboard/render-toolbar.ts
@@ -150,11 +150,12 @@ export function renderToolbar(
   total: number
 ): TemplateResult {
   // Layout: [search] [view-toggle] [facets…]
-  //         [X devices]                 [Select multiple]
+  //         [X devices] [Select multiple]
   // Select-multiple sits paired with the device-count on its own
   // row — both reference the device list ("operate on these N
-  // devices") so semantically they belong together. Frees the
-  // toolbar-row above for filter-related controls only.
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

stvncode (approved):
Good idea 👍

PR #753, opened 2026-06-10
Add a shared hover-ring token; use it for feedback modal rows

### LINKED ISSUE #137: Modal colors
Color are not good. Need some love

bdraco:
> [!WARNING]
> I wasn't really sure on this one on how big the scope should be so I left it at this. It likely needs more feedback. Or it might be something better for Steven to address directly since he is going to know better what is needed here

The feedback dialog rows hovered to a washed out tint that barely stood out from the default rows. Rather than special-case one dialog, this adds a shared `--esphome-hover-ring` token (a primary ring plus the focus-ring glow, the same affordance as focusing the toolbar search box) and uses it for the feedback rows; both row icons also turn primary on hover. The token bundles the ring and glow into one box-shadow, so other hover states across the app can migrate to it incrementally and the look stays in one place.

Scope note: `--esphome-tint` is currently the app-wide hover convention (~50 sites). This PR establishes the new token and applies it to the feedback dialog (the sample in the issue); broader rollout is intended as follow-ups pending design sign-off, not part of this PR.

While here, fixed the command palette mode-toggle hover, which set `background: var(--esphome-surface-hover)`, a token that does not exist, so the hover had no background; it now uses `--esphome-tint`.

Verified in light and dark mode.

- fixes https://github.com/esphome/backlog/issues/137

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

## Screenshots
Screenshot 2026-06-10 at 12 11 08 PM


stvncode (approved):
Change a bit the style. Thanks 🙏

PR #755, opened 2026-06-10
Fix select-mode checkbox alignment in card view and stacked table

bdraco:

In select mode the checkbox was misaligned in both device views:

- **Card view**: the 22px checkbox top-aligned against the two-line name+config block instead of centering on the title line (~2px high). The checkbox glyph now gets a flex box sized to the title's line height (`--wa-font-size-m` × `--wa-line-height-normal`), so it centers exactly on the device name while the header keeps `align-items: flex-start` (status badge stays top-anchored).
- **Table stacked/mobile layout**: `td.select-col` rendered as its own full-width row above the card title, visually detached from everything. It now pins to the card's top-left corner, mirroring the kebab pinned top-right, with the title cell reserving room on the left exactly like the kebab reservation on the right.

Verified empirically (headless Chrome at 1100px and 390px): card checkbox center now matches the title-line center to the pixel; stacked cards show checkbox top-left / kebab top-right flanking the title; desktop table is untouched (rules are scoped to the stacked-layout media query and `:host([select-mode])`).

- part of esphome/backlog#135 (item 3: select-multiple checkbox misaligned)

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/dashboard/table-styles.ts
+++ b/src/components/dashboard/table-styles.ts
@@ -465,13 +465,26 @@ export const tableLayoutStyles = css`
       padding: 0;
     }
 
-    /* Select checkbox sits at the top of the card. */
+    /* Select checkbox pins to the card's top-left corner, mirroring
+       the kebab at top-right, instead of stacking as its own
+       full-width row above the title. */
     td.select-col {
-      order: -2;
+      position: absolute;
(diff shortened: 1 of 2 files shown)


stvncode (approved):
lgtm ✅

PR #757, opened 2026-06-10
Rename Address to Hostname and link each drawer address row to its own host

### LINKED ISSUE #741: Web link icon should be on address, and rename address to hostname
### What happened?

The icon for accessing the web UI is located next to the IP address field, but it actually accesses the mDNS name (in this case: http://wf-tuersprechstelle.local/).

Wouldn't it make more sense for the icon to be located next to the address field? Or perhaps there should be an icon next to each of the two fields, allowing users to access the device via either the IP address or the mDNS name?

Maybe the address field should be labeled hostname?

Image

### Dashboard version

v0.1.0b108

### How did you install it?

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

Renames the device "Address" field to "Hostname" in the drawer and the device
table column, and gives each address row in the drawer its own open Web UI link.
The hostname row opens `http://<name>.local`; each IP row opens that IP.

Previously only the IP row carried the link, and it was built from the mDNS name,
so the icon next to the IP address actually opened the hostname; that is what the
issue reported. The localization key is renamed too (`drawer_address` to
`drawer_hostname`, `table_col_address` to `table_col_hostname`) so translations
don't carry a stale `address` name for a "Hostname" string.

- fixes https://github.com/esphome/device-builder-frontend/issues/741

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #759, opened 2026-06-10
Extract a shared visit Web UI link renderer

bdraco:

DRY follow-up to #757; no behavior change.

The "open Web UI" anchor was hand-rolled in five places (drawer hostname/IP rows,
table action cell, device card, discovered-device card, table row menu), each
repeating the `target="_blank"` + `rel="noopener noreferrer"` security pair, the
`action_visit_web_ui` aria/title, and the `open-in-new` glyph. This extracts one
`renderVisitWebUiLink` helper (`src/util/visit-web-ui-link.ts`) that every site
now calls, so the security pair can't drift from a new call site.

It also renames the drawer value renderer that #757 made serve the hostname row;
`renderIpValue` becomes `renderAddressValue` and its `.ip-value` / `.ip-value-text`
/ `.ip-visit-link` classes become `.address-*`, since they are no longer IP only.

Each anchor renders the same DOM as before; the renamed and new tests pin the
shapes, and a browser check confirmed the drawer is unchanged.

- n/a; follows up #757

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


esphbot (approved):
No blocking issues found.

PR #772, opened 2026-06-11
Use the shared FakeHost test helpers everywhere

bdraco:

Follow-up to #771, which introduced the shared `test/_fake-host.ts` helper but deliberately migrated only its own new call site to keep the feature diff small.

This moves the remaining duplicated ReactiveControllerHost stubs onto the shared helpers:

- The three local `class FakeHost` copies (`escape-controller`, `serial-ports-poll-controller`, `yaml-search-controller` tests) now import the shared class — it's a superset of all of them (the `updates` counter is simply unused where not asserted).
- The three verbatim-identical `vi.fn()`-based `fakeHost()` arrow stubs (`session-blob-cache-controller`, `trigger-catalog-controller`, `parse-error-controller` tests) move to a new spy-based `fakeHost` export in the same helper, keeping the existing call sites and assertion styles unchanged.

Intentionally-minimal one-off stubs stay local (`enter-controller`'s `{ addController() {} }`, `catalog-load-controller`'s one-liner, `component-name-resolver-controller`'s lifecycle-capturing stub) — they aren't the shared shape and exist for local reasons.

Net −52 lines, behavior-free: no test bodies or assertions changed.

- follow-up to #771

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/test/_fake-host.ts
+++ b/test/_fake-host.ts
@@ -1,4 +1,5 @@
 import type { ReactiveController, ReactiveControllerHost } from "lit";
+import { vi } from "vitest";
 
 /** Minimal ReactiveControllerHost for controller unit tests. */
 export class FakeHost implements ReactiveControllerHost {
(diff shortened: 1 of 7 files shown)


esphbot (approved):
No blocking issues found.

PR #773, opened 2026-06-11
Keep ${var} substitutions editable in the structured editor

### LINKED ISSUE #1391: [Bug] Substitutions are unexpectedly resolved in the structured editor so they can't be editted in place
### What happened?

Here's another thing where I'm not sure if it's a bug or expected behavior. It's an edge case. Please see screenshots below.

Simplified configuration:
```
substitutions:
  current_res: "0.05"
  voltage_div: "720"

sensor:
  - platform: hlw8012
    sel_pin:
      number: GPIO12
      inverted: True
    cf_pin: GPIO04
    cf1_pin: GPIO05
    current_resistor: ${current_res}
    voltage_divider: ${voltage_div}

    current:
      name: "${plug_name} - Ampere"
      unit_of_measurement: A
      accuracy_decimals: 3
      icon: mdi:flash-outline

    voltage:
      name: "${plug_name} - Volt"
      unit_of_measurement: V
      accuracy_decimals: 1
      icon: mdi:flash-outline

    power:
      name: "${plug_name} - Watt"
      unit_of_measurement: W
      id: "${plug_name}_Wattage"
      icon: mdi:flash-outline

    change_mode_every: 4
    update_interval: 3s

```

Screenshots:

Substitutions resolved in structured editor:
Image

Pressed "arrow up" on "Current Resistor" Value:
Image

Substitution was replaced and unrelated "Voltage Divider" value triggered error message "Must be a number"

### Dashboard version

v0.1.0b110

### How did you install it?

Desktop app

### Where is it running?

Local — Windows

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

Substitutions on typed fields were lost in the structured editor. A value like `current_resistor: ${current_res}` landed in a number/float/select/switch widget that can't hold the token, so the field showed the catalog default instead of the substitution, and the first interaction (an arrow-up, a toggle) overwrote `${current_res}` with a literal value in the YAML. Worse, once any field changed, full-form validation flagged every other untouched `${var}` numeric field red with "Must be a number", since `Number("${voltage_div}")` is `NaN`.

This routes any field whose value is a `${var}` / `$var` reference through the existing editable text renderer (the same path string fields already use), so the literal token round-trips, shows its resolved value as a hint, and is editable in place; `SECURE_STRING` keeps its masked input. Validation now treats a substitution reference as valid, since it resolves at build time and can't be checked here. Both are single guards at the form's leaf-dispatch and the validator.

- fixes esphome/device-builder#1391

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/config-entry-form.ts
+++ b/src/components/device/config-entry-form.ts
@@ -47,7 +47,10 @@ import {
 } from "../../util/pin-registry-modes-cache.js";
 import { registerMdiIcons } from "../../util/register-icons.js";
 import { SessionBlobCacheController } from "../../util/session-blob-cache-controller.js";
-import { parseSubstitutions } from "../../util/substitutions.js";
+import {
+  hasSubstitutionReference,
+  parseSubstitutions,
+} from "../../util/substitutions.js";
 import {
   _isStructuralType,
   filterRenderable,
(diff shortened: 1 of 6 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #778, opened 2026-06-12
Parse list items whose first field is a nested mapping

bdraco:

`parseYamlSectionValues` dropped every field of a `multi_conf` list item when the item's first field was a nested mapping. It read the item's child indent from the nested grandchildren instead of the item's field column, so it hoisted the grandchildren to the item level and lost the nested field plus all the sibling keys; the structured editor then rendered an empty form. This fixes the child indent detection for an inline dash key, and captures a dash line key whose value is a nested block. A small shared helper now does the nested mapping read for both the inline and the in loop paths so they stay in agreement.

This surfaced while testing the `font.file` structured editor (esphome/device-builder#1389), but it is an independent parser bug that already affected any component whose first list item field is a nested mapping; reordering the fields avoided it.

Tested with new read cases (gfonts, local, web, 4 space indent, sibling before and after) and round trip cases (edit a sibling, edit a nested field, unchanged save) that parse then write back through `updateSectionInYaml`; the full suite (2942 tests) and `npm run lint` pass, and it was verified live in the editor.

- relates to esphome/device-builder#1389

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/util/yaml-section-lexer.ts
+++ b/src/util/yaml-section-lexer.ts
@@ -125,6 +125,15 @@ export const parseBlockScalarHeader = (raw: string): BlockScalarHeader | null =>
 export const LIST_ITEM_BARE_DASH_RE = /^\s+-\s*$/;
 
 /**
+ * Capture the ``<indent>-<spaces>`` prefix up to the first non-space of a
+ * list item's inline key (``  - file:`` → ``"  - "``). Its length is the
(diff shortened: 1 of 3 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #779, opened 2026-06-12
Route a commented dash-line key to its nested mapping

bdraco:

Follow up to #778. When a list item's first field is a nested mapping and its dash line carries a comment (`- file:  # gfonts source`), the inline key reader kept the comment as the scalar value, so the nested block below was dropped. The reader now treats a comment-only remainder as empty (reusing `isCommentLine`) so the value routes to the nested-mapping path; the standalone comment is dropped, matching what the dash-line rewrite already does. A scalar with a trailing comment (`- file: gfonts://Roboto  # cached`) still parses to the value with the comment stripped.

This was the non-blocking edge Copilot flagged on #778.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/util/yaml-section-reader.ts
+++ b/src/util/yaml-section-reader.ts
@@ -407,17 +407,21 @@ export function parseSectionCore(
   if (isListItem) {
     const firstMatch = lines[startIdx].match(LIST_ITEM_INLINE_KEY_RE);
     if (firstMatch) {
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #782, opened 2026-06-12
Render the command palette in a wa-dialog so it stacks above open dialogs

### LINKED ISSUE #1402: [Bug] CTRL+K dialog opens behind the current dialog
### What happened?

If another dialog is open (Settings, Firmware Tasks, ...) and you press CTRL+K, the CTRL+K dialog opens _behind_ the current dialog and cannot be interacted with. I suspect that's not what was intended.

Image

Image

### Dashboard version

v0.1.0b112

### How did you install it?

Desktop app

### Where is it running?

Local — Windows

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

Ctrl+K opened the command palette behind an already open dialog; Settings and Firmware Tasks are wa-dialogs shown via showModal, which puts them in the browser top layer, and no z-index can put a plain fixed overlay above that. The palette is now a wa-dialog itself, so it opens in the top layer above the current dialog and joins the dismissible stack; Escape closes the palette first and leaves the dialog underneath open, backdrop click and focus handling come from wa-dialog's light-dismiss and autofocus instead of the custom backdrop and EscapeController.

- fixes https://github.com/esphome/device-builder/issues/1402

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/command-palette.styles.ts
+++ b/src/components/command-palette.styles.ts
@@ -1,62 +1,29 @@
 import { css } from "lit";
 
 export const commandPaletteStyles = css`
-  :host {
-    position: fixed;
-    inset: 0;
-    /* Above WA component overlays (wa-popup is 899, wa-select is 900).
-       Note: top-layer popovers are above any z-index — open() also
-       force-closes them so this matters mainly for non-popover layers. */
-    z-index: 10000;
-    pointer-events: none;
+  wa-dialog {
+    --width: min(640px, calc(100vw - var(--wa-space-l)));
+    --spacing: 0;
+    --backdrop-filter: blur(2px);
+    --wa-color-overlay-modal: rgba(0, 0, 0, 0.45);
+    --show-duration: 140ms;
+    --hide-duration: 120ms;
   }
 
-  :host([hidden]) {
-    display: none;
-  }
-
-  .backdrop {
-    position: fixed;
-    inset: 0;
-    background: rgba(0, 0, 0, 0.45);
-    backdrop-filter: blur(2px);
-    pointer-events: auto;
-    animation: fade-in 0.12s ease-out;
-  }
-
-  .dialog {
-    position: fixed;
-    top: 18%;
-    left: 50%;
-    transform: translate(-50%, 0);
-    width: min(640px, calc(100vw - var(--wa-space-l)));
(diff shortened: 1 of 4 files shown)


esphbot (changes requested):

bdraco:
Addressed in 0cd3f63:

1. The dialog now has an accessible name; the header is hidden with a `::part(header)` style instead of `without-header`, so wa-dialog's `label` still resolves through `aria-labelledby`, and the search input got an explicit `aria-label`.
2. Reopen focus verified headlessly with puppeteer; the input regains focus on the first open and on later reopens, since wa-dialog runs its `[autofocus]` query on every `show()`.
3. Closed-state rendering is now gated on a flag that clears on `wa-after-hide`; a closed palette renders nothing again, the content only stays mounted through the hide animation.
4. Added `test/components/command-palette-after-hide.test.ts` pinning the flag sync, the descendant `wa-after-hide` guard, and that `close()` keeps content for the hide animation. Top layer stacking and Esc ordering were also verified against the issue repro, palette closes first, the dialog under it stays open.

esphbot (changes requested):

bdraco:
Addressed in 17db7fa: state now syncs on the initiating `wa-hide` (so Esc and light-dismiss clear the yaml search debounce immediately, no animation window), and `wa-after-hide` only unmounts the content, guarded with `!_open` so a reopen mid animation keeps it. Tests updated to pin the split contract, including the reopen-during-animation case and the descendant event guard on both handlers.

esphbot (approved):
No blocking issues found.

bdraco:
Restored an explicit post render focus in `open()` in 18df84b, kept alongside the `autofocus` attribute. For the record, wa-dialog does not rely on native parse time autofocus; its `show()` runs `querySelector("[autofocus]").focus()` in a rAF on every `showModal()`, and reopen focus was verified headlessly across multiple opens. The explicit call is belt and braces against a future webawesome change, and the attribute keeps `showModal()`'s native focus steps landing on the input rather than the dialog container.

esphbot (approved):
No blocking issues found.

PR #783, opened 2026-06-12
Add depends_on_value_any set-valued visibility gate

### LINKED ISSUE #1397: Ethernet lacks a structured editor
Since https://github.com/esphome/bluetooth-proxies does use ethernet and people do change these, it would be nice to have an editor there since we do change fields (`clk` recently) and it would make it easier for users to adapt to new schema

bdraco:

Adds a set-valued visibility gate to config entries: `depends_on_value_any` shows an entry only when the dependency's value is in the list. `isEntryVisible` checks it after the existing scalar gates, so rendering, validation and serialization all honor it through the one shared predicate.

This is the frontend half of polymorphic structured editors for components whose whole CONFIG_SCHEMA is a `cv.typed_schema` union (ethernet, spi, usb_uart, i2s_audio and friends). The backend companion PR (esphome/device-builder#1409) emits one entry per unique field, gated on the set of variants that carry it; picking a different type in the editor swaps the visible fields. Verified live against the companion backend: ethernet renders a structured editor, switching the type from LAN8720 to W5500 swaps the RMII pin fields for the SPI ones.

- fixes esphome/device-builder#1397

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/api/types/config-entries.ts
+++ b/src/api/types/config-entries.ts
@@ -141,6 +141,8 @@ export interface ConfigEntry {
   depends_on_value: ConfigPrimitive | null;
   /** Show only when dependency value does NOT equal this. */
   depends_on_value_not: ConfigPrimitive | null;
+  /** Show only when dependency value is in this list. */
+  depends_on_value_any: ConfigPrimitive[] | null;
   /**
    * Hide this entry unless the named component is configured on the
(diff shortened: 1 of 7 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #789, opened 2026-06-12
Read zero-indented LIST_SECTIONS bodies against the header indent

### LINKED ISSUE #788: Column 0 compact sequence under a LIST_SECTIONS singleton reads empty and save replaces the block
Follow up documented in #786, pre existing but more visible now that column 0 platform lists work in the editor (esphome/device-builder#1403).

A `LIST_SECTIONS` singleton whose whole value is a column 0 compact sequence reads as empty in the values layer, and saving after adding a row replaces the entire block:

```yaml
globals:
- id: x
  type: int
```

Read: `_detectSectionChildIndent` falls back to the 2 space indent (the column 0 dash is not deeper than the header's floor of 0), then `isChildListItemLine(lines[peek], "  ")` fails its lead vs parent indent check (`yaml-section-lexer.ts` ~321), so the `LIST_SECTIONS` branch is skipped and `values.globals` is absent; the editor shows zero globals.

Write: a no touch open is safe (`!Array.isArray(undefined)` bails at `yaml-section-values.ts` ~131), but once the user adds one row the draft flushes with `values.globals` as an array, the `LIST_SECTIONS` branch splices the full `findSectionRange` span, and every pre existing column 0 global is silently replaced by the single new row. That is data loss in the open editor, add row, save flow.

**Required coupling, do not fix the read path alone.** Routing column 0 sequences into `parseListBlock` requires loosening `LIST_ITEM_BARE_DASH_RE` and `LIST_ITEM_DICT_KEY_RE` (`yaml-section-lexer.ts` ~127, ~157) in the same change; otherwise a column 0 `- id: x` misses `LIST_ITEM_DICT_KEY_RE`, the block classifies as a scalar list, and the items round trip back into the YAML as quoted strings on save, which is type corruption rather than emptiness.

A cheaper interim hardening is also acceptable as a first PR: have the read flag "section has content the parse did not claim" and make the `LIST_SECTIONS` splice bail in that case, converting the data loss into a no op save.

bdraco:

A `LIST_SECTIONS` singleton whose whole value is a zero indented sequence (`globals:` with column 0 dashes) read as empty in the values layer, the editor showed no global variables, and saving after adding a row replaced the entire block with just the new row. The reader's list bodied peek and `parseListBlock` call now use the section header's own indent instead of the detected child indent, matching `parseListBlock`'s documented contract; the child indent fallback of two spaces was what rejected column 0 dashes. The coupled complexity regexes `LIST_ITEM_BARE_DASH_RE` and `LIST_ITEM_DICT_KEY_RE` are loosened to `\s*` in the same change, without that a column 0 `- id:` classifies the block as a scalar list and the items round trip as quoted strings. The loosenings only activate for header indent zero blocks; every other path terminates on those lines before the complexity checks run, and the full suite confirms no behavior change elsewhere.

Verified in the browser against a column 0 `globals:` config: the editor now renders both variables as structured rows, and the add row round trip is pinned by unit tests.

- fixes #788

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/util/yaml-section-lexer.ts
+++ b/src/util/yaml-section-lexer.ts
@@ -122,8 +122,8 @@ export const parseBlockScalarHeader = (raw: string): BlockScalarHeader | null =>
  * dropped from the values dict, and the user's empty row vanishes
  * on save. Marking it complex routes the block through
  * ``collectBlockListMappings`` which already treats bare dashes
- * as ``{}`` placeholders. Indent is ``\s*`` — a zero-indented
- * sequence puts dashes at column 0, like ``LIST_ITEM_START_RE``.
+ * as ``{}`` placeholders. Indent is ``\s*`` for
+ * ``LIST_ITEM_START_RE``'s column-0 reason.
  */
 export const LIST_ITEM_BARE_DASH_RE = /^\s*-\s*$/;
 
@@ -153,8 +153,8 @@ export const LIST_ITEM_INLINE_KEY_PREFIX_RE = /^(\s*-\s*)\S/;
  *
  * Allows zero trailing whitespace after the colon (header-only
  * line) AND content after it (`- lambda: |-`); both forms are
- * complex. Indent is ``\s*`` — a zero-indented sequence puts
- * dashes at column 0, like ``LIST_ITEM_START_RE``.
+ * complex. Indent is ``\s*`` for ``LIST_ITEM_START_RE``'s
+ * column-0 reason.
  */
 export const LIST_ITEM_DICT_KEY_RE = /^\s*-\s+[a-zA-Z_][\w.]*:(?:\s|$)/;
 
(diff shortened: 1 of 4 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
Confirmed, and the radius is tighter than the suggestion assumes. `keyIndent=""` reaches `_scanValueBlock` only through the `LIST_SECTIONS` branch: of its four call sites, the in item lambda path passes `dashIndent + ESPHOME_YAML_INDENT` (always nonzero), the nested key paths pass `childIndent` / `indent` (matched at two or more spaces), and `parseListBlock`'s pass through is the parent indent, which is only ever `""` from the `LIST_SECTIONS` branch this PR touched.

A non member top level key with a column 0 dash body never reaches the loosened regexes at all; the main loop's `isChildListItemLine(peek, childIndent)` rejects it (0 < 2) and `readNestedMappingAfter` bails on the same comparison, so those sections keep their pre PR behavior byte for byte. That shape (`external_components:` with column 0 items) is the separately tracked values layer gap, not something this PR changes.

PR #797, opened 2026-06-12
Render the Delay unit picker with wa-select

bdraco:

In the automation editor's Delay action form the Value input and the Unit picker rendered at different heights; the unit picker was a native select, so it missed both the shared input min-height and the wa-select chrome that every other value plus unit pair uses.

The unit picker is now a wa-select like the time period and interval fields, and the local input/select chrome overrides are dropped so both controls take the shared form-control styling; verified in the browser that the pair renders equal height and that changing the unit still round-trips to the YAML.

- none, reported during testing of the visual editor

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


esphbot (changes requested):

esphbot (changes requested):

esphbot (approved):
No blocking issues found.

PR #800, opened 2026-06-12
Render remote build submit dialog pickers with wa-select

### LINKED ISSUE #799: Remote build submit dialog selects render with browser default chrome
The remote build submit job dialog (`src/components/remote-build-job-dialog.ts`, the `#rb-config` and `#rb-target` selects) uses native `<select>` elements with no styling rule, so both render with browser default chrome inside an otherwise design system styled dialog. The component already includes `inputStyles`, which only covers native inputs and `wa-select`; a bare `<select>` matches neither.

Same root cause as the Delay unit picker fixed in #797; converting both to `wa-select`/`wa-option` picks up the shared form control chrome. No adjacent text input in this view, so it is a chrome inconsistency rather than a side by side height mismatch.

bdraco:

The remote build submit job dialog rendered its configuration and target pickers as native selects, so both showed browser default chrome inside an otherwise design system styled dialog; the component already pulls in inputStyles, which only covers native inputs and wa-select, so a bare select matched neither.

Both pickers are now wa-select with wa-option, picking up the shared form control chrome like every other picker in the app; the labels use aria-labelledby since a native label for does not reliably associate with a custom element. Verified in the browser that both render at the shared form control height, and a unit test pins the labelled wa-select shape and the seeded selection.

- fixes https://github.com/esphome/device-builder-frontend/issues/799

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/remote-build-job-dialog.ts
+++ b/src/components/remote-build-job-dialog.ts
@@ -23,6 +23,9 @@ import { isTerminalJobStatus } from "../util/firmware-job-status.js";
 import { renderErrorBanner } from "../util/render-error.js";
 import { remoteBuildJobDialogStyles } from "./remote-build-job-dialog.styles.js";
 
+import "@home-assistant/webawesome/dist/components/option/option.js";
+import "@home-assistant/webawesome/dist/components/select/select.js";
+
 import "./ansi-log.js";
 import "./base-dialog.js";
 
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #802, opened 2026-06-12
Render the secret target picker with wa-select

### LINKED ISSUE #798: Add Secret dialog target select renders unstyled beside the styled name input
The Add Secret dialog's target select (`src/components/secrets/secrets-structured-editor.ts`, the `.add-select` element) is a native `<select>` with no styling rule anywhere, so it renders with browser default chrome. It sits directly above the secret name text input in the same form, which gets the shared `inputStyles` chrome (`--wa-form-control-height`, raised background), so the two controls have mismatched height and styling.

Same root cause as the Delay unit picker fixed in #797: `inputStyles` harmonizes native inputs and `wa-select` to the same form control height, but a bare `<select>` matches neither rule. Fix is the same one element swap to `wa-select`/`wa-option`.

bdraco:

The Add Secret dialog's target picker was a native select, so it carried browser default arrow and focus chrome and hand rolled its own height/border/background in the styles file rather than going through the wa-select convention the rest of the app uses; it sat directly above the secret name text input, which gets the shared inputStyles chrome.

The target picker is now a wa-select with wa-option, picking up the shared inputStyles combobox chrome keyed on the form control height token, so it matches the name input beside it; the hand rolled `.add-select` chrome block is dropped, and the label uses aria-labelledby since a native label for does not reliably associate with a custom element. A unit test pins the labelled wa-select shape, and the existing target switch tests still pass.

- fixes https://github.com/esphome/device-builder-frontend/issues/798

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


esphbot (changes requested):

esphbot (approved):
No blocking issues found.

PR #803, opened 2026-06-12
Style remote build dialog action buttons via shared styles

### LINKED ISSUE #801: Remote build dialog action buttons render with browser-default chrome
The remote build job dialog's action row buttons (`src/components/remote-build-job-dialog.ts`, the `.btn-secondary` Close/Back and `.btn-primary` Build remotely buttons) render with raw browser default chrome; there is no style rule for those classes in `remote-build-job-dialog.styles.ts`, and the styles file's own comment acknowledges it: "The other two buttons in this dialog (.btn-primary / .btn-secondary) inherit browser-default chrome."

The codebase already has a shared module for exactly this, `src/styles/dialog-action-buttons.ts` (`dialogActionButtonStyles`), which provides `.btn`, `.btn--cancel`, and `.btn--primary` for dialogs that render their own inline action row. The fix is to import that fragment into the dialog's `static styles` and rename the six button usages from `.btn-secondary` / `.btn-primary` to `.btn .btn--cancel` / `.btn .btn--primary`, matching the convention the other inline action-row dialogs use.

Surfaced while fixing the unstyled selects in #799 / #800; same dialog, separate concern.

bdraco:

The remote build job dialog rendered its action row buttons (Close / Back and Build remotely) with the `.btn-secondary` / `.btn-primary` classes, which had no style rule, so they showed raw browser default chrome inside an otherwise design system styled dialog; the styles file's own comment even acknowledged it.

The dialog now pulls in the shared `dialogActionButtonStyles` and the buttons use `.btn .btn--cancel` / `.btn .btn--primary`, matching the inline action row convention the other dialogs use. The per row destructive Cancel button moves to `.btn .btn--danger`, keeping just the danger tint in the local styles over the shared base, per that module's destructive variant contract, which also drops the duplicated base shape it used to hand roll. Verified in the browser that Close picks up the neutral surface chrome and Build remotely the primary tint at equal height.

- fixes https://github.com/esphome/device-builder-frontend/issues/801

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/remote-build-job-dialog.styles.ts
+++ b/src/components/remote-build-job-dialog.styles.ts
@@ -27,29 +27,22 @@ export const remoteBuildJobDialogStyles = css`
     margin-top: var(--wa-space-xs);
   }
 
-  /* Destructive variant for the per-row Cancel-this-job
-     button. Mirrors confirm-dialog's destructive btn tint
-     without dragging the whole confirm-dialog style block in.
-     The other two buttons in this dialog (.btn-primary /
-     .btn-secondary) inherit browser-default chrome. */
-  .btn-danger {
+  /* Destructive variant for the per-row Cancel-this-job button —
+     just the danger tint over the shared .btn base from
+     dialogActionButtonStyles, per that module's "destructive
+     variants stay in the consumer's local styles" contract. */
+  .btn--danger {
     background: var(--esphome-error);
     color: var(--esphome-on-primary);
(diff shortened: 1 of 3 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #805, opened 2026-06-13
Pre-fill required nested sensors and tidy the add-component layout

### LINKED ISSUE #1423: Adding ags10 fails because TVOC doesn't get pre-filled
the TVOC required field isn't create with a name by default. 

Also the required fields should be shown at the top, and the optional ones under it. 

The optional ones should not be expanded by default because it overloads the screen

Image

Image

bdraco:

Adding a component with a required nested entity reading (ags10's `tvoc`) failed with "invalid_args: Missing required field: TVOC". The group serialized only once it held a value, so an untouched Add sent no `tvoc` and the backend rejected it.

Seed the reading's name from its label (the same value the optional enable toggle already uses) so a bare Add produces a valid sensor. While here, float required fields above optional ones in the add dialog, and stop force-expanding every nested group; optional readings now start collapsed so the form does not overload the screen.

- fixes https://github.com/esphome/device-builder/issues/1423

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/add-component-form.ts
+++ b/src/components/device/add-component-form.ts
@@ -227,6 +227,18 @@ export class ESPHomeAddComponentForm extends LitElement {
     for (const entry of entries) {
       if (entry.type === ConfigEntryType.NESTED) {
         const sub = this._seedDefaults(entry.config_entries ?? [], seedAll);
+        // A required entity sub-reading (ags10's tvoc) serializes only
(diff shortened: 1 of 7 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #807, opened 2026-06-13
Pre-fill dep-added buses from the requester's bus constraints

### LINKED ISSUE #1425: Adding an I2C component with a frequency constraint (ags10) yields an invalid bus frequency
### What happened?

Adding a component that constrains its I²C bus frequency (e.g. `ags10`) reuses the existing `i2c:` bus but never sets a compatible `frequency:`, so the generated config fails ESPHome's final validation.

Repro:
1. Config with a plain `i2c:` bus (no `frequency:`, so it defaults above 15kHz).
2. Add component → sensor → AGS10 → Add.
3. Editor squiggle on `i2c:`: `Component ags10 cannot be used with a frequency of over 15khz for the I2C bus.`

The add itself produces structurally valid YAML (`sensor: - platform: ags10\n  i2c_id: i2c_1\n  tvoc:\n    name: TVOC`); the incompatibility only surfaces at compile-time, because the dashboard has no awareness of the bus-frequency constraint.

### Root cause

`ags10` (and other slow I²C devices) declare the limit via a final-validation against the bus, not a regular schema field:

```python
# esphome/components/ags10/sensor.py
FINAL_VALIDATE_SCHEMA = i2c.final_validate_device_schema("ags10", max_frequency="15khz")
```

`i2c.final_validate_device_schema` (esphome/components/i2c/__init__.py:297) injects a `cv.Required(frequency)` Range onto the *hub* schema. The component catalog is built from the per-component schema and doesn't carry these cross-component (device → bus) constraints, so neither the add-component dependency flow nor the form knows the existing bus needs `frequency: 15kHz`.

### Possible directions

- Sync-script: introspect `FINAL_VALIDATE_SCHEMA` to capture min/max bus frequency (and timeout) constraints into the catalog entry.
- Add-component flow: when the chosen component constrains a bus it's attaching to, either set a compatible `frequency:` on the bus or warn the user before Add.

Separate from the `ags10` prefill fix (esphome/device-builder-frontend#805), which is what surfaced this while testing.

### Dashboard version

dev (main)

bdraco:

A bus added through the dependency detour came up with defaults the requesting component rejects at final validation: ags10 needs i2c at or below 15kHz, most uart devices pin an exact baud rate, and many require specific pins.

The detour now reads the requester's `bus_constraints` from the catalog (added in esphome/device-builder#1429), prefills exact-match values and a clamped frequency on the bus form, and marks constraint-required pins as required so submit waits for a pin choice. Verified against the live backend: ags10 yields an i2c form at 15kHz, dfplayer yields a uart form at 9600 baud with tx_pin required, mcp9600 leaves the in-range default alone.

Depends on esphome/device-builder#1429 for the catalog field; without it the detour behaves exactly as today.

- fixes https://github.com/esphome/device-builder/issues/1425

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/api/types/components.ts
+++ b/src/api/types/components.ts
@@ -108,6 +108,11 @@ export interface ComponentCatalogEntry {
   /** Interfaces this component can be referenced *as* beyond its own domain
    *  (an `adc` sensor provides `voltage_sampler`). */
   provides?: string[];
+  /** Requirements this component imposes on the bus it attaches to, keyed
+   *  by bus id ('i2c' / 'spi' / 'uart'): exact-match values (baud_rate,
+   *  parity, ...), range bounds (min/max_frequency in Hz) and required
+   *  pins (require_tx / require_mosi / ...). Drives dep-add prefill. */
+  bus_constraints?: Record<string, Record<string, unknown>>;
   /** The component's configuration schema. May contain `nested` entries
    *  (`type === "nested"`) whose `config_entries` recurse. */
   config_entries: ConfigEntry[];
--- a/src/components/device/add-component-dialog-dep-nav.ts
+++ b/src/components/device/add-component-dialog-dep-nav.ts
@@ -1,5 +1,9 @@
 import type { ESPHomeAPI } from "../../api/index.js";
 import type { ComponentCatalogEntry } from "../../api/types/components.js";
+import {
+  busConstraintPrefill,
+  type BusPrefill,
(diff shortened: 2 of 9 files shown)


esphbot (changes requested):

esphbot (approved):
No blocking issues found.

PR #813, opened 2026-06-13
Surface a select field's schema default in the box and menu

bdraco:

A select in the structured editor rendered blank when its field was
omitted from YAML, even when the schema carried a default — and opening
the menu activated the *first* option (wa-select's default behavior with
no committed value) rather than the actual default. So the field looked
unset and the wrong option looked selected (the `spi` `type` discriminator
was the reported case: it defaults to `single`).

This surfaces the default in two places:

- **In the box:** the default's label shows as a muted placeholder. Muted
  reads as "this is the default"; a solid value means an explicit pick —
  that contrast is the default-vs-selected signal, so no separate hint
  line is needed.
- **In the menu:** the default option gets a quiet "default" second line
  (same shape as the pin menu's notes), so it stays identifiable even
  though wa-select still keyboard-activates the first option.

Pairs with the backend recovering `cv.typed_schema` discriminator
defaults (esphome/device-builder#1437), so `spi` `type` shows a muted
`single`. Also helps every ordinary `cv.Optional(default=...)` enum that
already carries a default (e.g. `sensor.aht10`'s `variant`).

- Companion to esphome/device-builder#1437

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/config-entry-renderers/primitives.ts
+++ b/src/components/device/config-entry-renderers/primitives.ts
@@ -461,8 +461,9 @@ export function renderSelectField(entry: ConfigEntry, path: string[], ctx: Rende
   // still flags as selected.
   const valueLower = value.toLowerCase();
   const defaultStr = entry.default_value != null ? String(entry.default_value) : "";
+  const defaultLower = defaultStr.toLowerCase();
   const defaultOption = entry.options?.find(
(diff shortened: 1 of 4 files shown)


esphbot (approved):
No blocking issues found.

PR #817, opened 2026-06-13
Resolve aliased pin values and grey a pin field's default

bdraco:

A pin field whose value is a named alias rendered with nothing selected. On ESP8266 a `uart: { rx_pin: RX }` (where `RX` is GPIO3) showed a blank RX Pin dropdown; `parsePinGpio` only understands `GPIOn` / integer / `P0.x` forms, so the value fell through as the raw string `RX` and matched no option.

When the value isn't a `GPIOn` form, the pin renderer now resolves it against the board's per-pin `aliases` (the catalog carries `RX`/`TX`/`D1`/... on each pin) before falling back to the string; the resolved GPIO drives both the selected option and the re-add of a filtered-out active pin, so an aliased value stays visible. `BoardPin` gains the optional `aliases` field.

It also greys a pin field's schema default in the box when the field is unset, matching the select renderer; an alias-named default (i2c `sda: SDA`) resolves to its GPIO so the box shows the real pin (GPIO4). A field with no default (uart `tx_pin`) stays blank.

Needs the companion backend that adds the aliases to the catalog and resolves a config's `board:` to its exact entry; fixes esphome/device-builder#1442.

- Reported for the structured editor; no separate tracking issue.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/api/types/boards.ts
+++ b/src/api/types/boards.ts
@@ -29,6 +29,9 @@ export interface BoardPin {
   available: boolean | null;
   occupied_by: string | null;
   notes: string | null;
+  /** Named forms a config may refer to this pin by (``RX``, ``D1``); the
+   *  catalog omits the key when there are none. */
+  aliases?: string[];
 }
 
 /**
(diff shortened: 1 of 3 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #824, opened 2026-06-14
Show auto-resolved default for single-candidate ID reference fields

bdraco:

An ID reference field such as **Ld2410 ID** renders an empty dropdown when it has no value, even when the config holds exactly one matching component that ESPHome resolves to automatically. This shows that sole match as a greyed default in the closed select and tags its option, mirroring how GPIO and enum fields display their default; the value stays omitted, so ESPHome keeps resolving it. When a top level `packages:` block or a `<<:` merge could pull in another match, the hint is suppressed, since the in editor scan cannot see merged files.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


esphbot (approved):
No blocking issues found.

PR #825, opened 2026-06-14
Preselect the current value when opening the options combobox

bdraco:

Opening a pre-filled options combobox (the ESP8266/Libretiny **Board** picker is the obvious one) left the list scrolled to the top with nothing highlighted, so a committed value far down the list looked unselected. On open, the current value is now set as the active option and scrolled into view; arrow keys move relative to it, and a free-text value that isn't in the list stays unhighlighted.

Note: the first ArrowDown after opening now moves from the current value rather than always jumping to the top option; that matches the preselection and is the expected combobox behavior.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/options-combobox.ts
+++ b/src/components/options-combobox.ts
@@ -166,7 +166,12 @@ export class ESPHomeOptionsCombobox extends LitElement {
     this._committed = this.value;
     this._query = this.value;
     this._dirty = false;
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

PR #829, opened 2026-06-14
Coerce numeric combo box values to their declared type on emit

bdraco:

The `allow_custom_value` combo box (`esphome-options-combobox`) always emits a string, so a numeric field rendered as a combo box wrote a quoted value like `baud_rate: "9600"` into the YAML instead of a number. This coerces the picked / typed value back to the entry's declared type before emitting, so INTEGER and FLOAT fields stay numeric.

The coercion is extracted into a shared `coerceValueToEntryType` helper and reused by both the combo box and the existing suggestion select (which had the same inline logic). INTEGER goes through `coerceIntFieldValue`, so a decimal above 2^53 stays a string for 64-bit precision and a `0x…` literal isn't truncated; nothing new is reinvented.

This pairs with esphome/device-builder#1451, which turns the UART `baud_rate` field into a common-rate combo box.

- No tracking issue.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


esphbot (changes requested):

PR #835, opened 2026-06-14
Resolve featured-add id references to the existing bus

bdraco:

Adding a featured component whose id reference (like the Apollo Starter Kit battery monitor's `i2c_id`) carries a backend manifest preset would write that literal verbatim; on a config whose bus has a different id (e.g. `i2c_1`), the result was `i2c_id: i2c_bus`, a dangling reference that won't compile.

Now an unlocked id-reference field is seeded with the matching component already in the config: exactly one matching bus is filled in, and anything ambiguous (no bus, two or more buses, or a `packages:` merge that could hide one) is left unset so the existing missing-dependency detour or the picker takes over. A stale preset literal is never written. Locked references (a deliberately pinned bundled `output`) keep their literal.

The resolution rule is shared with the id-reference picker via `resolveSoleCandidate`. The two agree for same-domain bus refs (i2c/spi/uart); a cross-domain interface ref finds no same-domain candidate at seed time and defers to the picker, which resolves it with its async interface providers.

Note on scope: this seeding sits after the `required` gate, so it also resolves *required* reference fields on regular (non-featured) components, pre-selecting a single existing bus there too. That matches the picker's own single-candidate default and is the intended behaviour, just broader than the "featured-add" framing.

- fixes <link to issue>

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/add-component-form.ts
+++ b/src/components/device/add-component-form.ts
@@ -16,7 +16,7 @@ import { seedBoardPinDefaults } from "../../util/board-pin-defaults.js";
 import { ComponentNameResolverController } from "../../util/component-name-resolver-controller.js";
 import {
   findReferenceCandidates,
-  yamlHasMergedSources,
+  resolveSoleCandidate,
 } from "../../util/config-entry-yaml-scan.js";
 import { validateEntries, type ValidationError } from "../../util/config-validation.js";
 import {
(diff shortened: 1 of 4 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #839, opened 2026-06-14
Surface the backend reason in the rename-failure toast

bdraco:
## Description

Renaming a device could fail with a bare toast **"Failed to rename "<name>""** and no reason. `executeRename` was the only device action using a bare `catch {}`, so it discarded the rejected `devices/rename` command's error. The siblings (`executeClone`, `executeFriendlyName`) already bind the error and surface `getErrorMessage(err)` as a `{reason}`; this brings rename in line.

The most common trigger is a synchronous collision: a rename queues a compile + OTA job that writes `<new-name>.yaml` before the OTA step, so a retry while that job is still in flight rejects with `invalid_args: A device named <new-name>.yaml already exists` — a reason the user never saw.

Reproduced live over `/ws` and verified end-to-end in a headless browser; the toast now reads:

> Failed to rename "rename_test": invalid_args: A device named rename-test.yaml already exists

This is the error-reporting half. A follow-up will address the underlying job-failure path (a failed rename job currently reports `exit_code=1` with `error=None`, and the command-dialog banner shows a generic "Rename failed.").

related issue https://github.com/esphome/device-builder/issues/1465

## Types of changes

## Testing

- New `test/components/dashboard/actions-ui.test.ts` asserts the rename-failure toast carries the backend reason.
- `npm run lint` + full vitest (3219 tests) pass; prettier clean.

### DIFF
--- a/src/components/dashboard/actions-ui.ts
+++ b/src/components/dashboard/actions-ui.ts
@@ -93,10 +93,15 @@ export async function executeRename(
   let response: Awaited<ReturnType<ESPHomeAPI["renameDevice"]>>;
   try {
     response = await host._api.renameDevice(device.configuration, newName);
-  } catch {
-    toast.error(host._localize("dashboard.action_rename_failed", { name: device.name }), {
-      richColors: true,
-    });
+  } catch (err) {
+    const reason = getErrorMessage(err);
+    toast.error(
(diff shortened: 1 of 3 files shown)


esphbot (approved):
No blocking issues found.

PR #849, opened 2026-06-15
Split numeric field renderers out of config-entry primitives

esphbot:

Split the numeric/magnitude field renderers out of
`config-entry-renderers/primitives.ts` into a sibling
`primitives-numeric.ts`, bringing both files under the 600-line cap.

`primitives.ts` had regrown to 624 lines, over the repo's file-size cap.
The file is a collection of independent field-renderer functions, so it
splits cleanly by field shape — the same pattern already used by the
sibling `lists.ts` / `nested.ts` / `registry-list.ts` modules in this
folder.

- Moved the numeric cluster — `renderNumberField` (and its private
  `renderIntField` / `renderHexIntField` / `hexDisplayOrFallback`
  helpers), `renderTimePeriodField`, and `renderFloatWithUnitField` —
  into `primitives-numeric.ts`.
- `primitives.ts` re-exports the three public renderers from the new
  module, so every consumer (the `config-entry-renderers` barrel,
  `config-entry-pin-renderer`, and the test suite) keeps importing from
  the same path — zero consumer churn.
- The numeric cluster shares no private helpers with the remaining
  selection/boolean/text renderers, so there's no circular import.
- Result: `primitives.ts` 298 lines, `primitives-numeric.ts` 342 lines.
  No behaviour change; public surface unchanged.

- N/A

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


bdraco (changes requested):
You must use the pr template, do not mangle it

bdraco:
@esphbot rr

bdraco (approved):

PR #857, opened 2026-06-15
Release the device's active slot when a stopped install compile cascades

### LINKED ISSUE #1482: UI state of device not released on build break
### What happened?

When running a build for a device the UI "locks" the state of the device (twirl on the status).
If you kill the build mid way, the UI never "releases" the state of the device - you need to reload the page to interact with the device again.

### Dashboard version

v0.1.0b118

### How did you install it?

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell

```

### Anything else?

This happens both with local and remote builds.

bdraco:

Stopping a compile that is part of an install left the device card stuck on "Installing..." until a page reload.

An install queues a compile plus a dependent upload. Hitting Stop cancels the compile, which cascades a cancel onto the still-queued upload. The compile's `job_cancelled` arrives while the upload is still non-terminal, so `terminateJob` mistook it for a live successor (a re-compile supersede), dropped it from history, and returned early without releasing the card's active slot; the upload's later cancel did not match the slot either, so it stayed latched on the cancelled compile. A reload re-derived state from the snapshot (both terminal) and cleared it, which is why reloading "fixed" it.

The fix releases the active slot whenever the terminating job is the latched one, before the supersede branch can return early. A freshly-queued follow-up keeps its slot since its job_id will not match. Added a regression test replaying the exact queued-upload, queued-compile, started-compile, cancelled-compile, cancelled-upload sequence captured from the live backend, plus a guard that a real live successor's slot is preserved.

- fixes https://github.com/esphome/device-builder/issues/1482

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


esphbot (approved):
No blocking issues found.

PR #859, opened 2026-06-16
Send empty wifi credentials when reusing secrets in the wizard

### LINKED ISSUE #1490: Default device YAML generates quoted "!secret" tags which are treated as literal strings
### The problem

When creating a new device from a board manifest (tested with ESP32-C3 
Super Mini), the generated default YAML wraps !secret tags in quotes:

  wifi:
    ssid: "!secret wifi_ssid"
    password: "!secret wifi_password"

This causes the device to literally attempt connecting to a network 
named "!secret wifi_ssid" instead of resolving the secret from 
secrets.yaml. The log shows:

  Hidden candidate '!secret wifi_ssid' at index 0
  Disconnected reason='Probe Request Unsuccessful'

The correct syntax is:

  wifi:
    ssid: !secret wifi_ssid
    password: !secret wifi_password

The !secret YAML tag must not be quoted. Quoting it converts it from 
a YAML directive into a plain string. There is no error or warning 
shown to the user — the device compiles and flashes successfully but 
silently fails to connect to WiFi.

Note: ESPHome 2026.4.5 added a workaround on the bundle side 
(esphome#16271), but the root cause — the template generating quoted 
!secret tags — remains unfixed in Device Builder.

### Which version of ESPHome Device Builder has the issue?

0.1.0b110

### What platform are you using?

Windows

### Anything in the logs that might be useful for us?

```txt
Affected manifest: definitions/boards/esp32-c3-supermini/manifest.yaml
ESPHome version: 2026.5.3
```

### Additional information

_No response_

bdraco:

When a device is created with an existing secrets.yaml, the wizard reused the wifi secret by sending the literal string "!secret wifi_ssid" as the ssid value; the backend treats a non-empty ssid as a real credential and quotes it, so the generated YAML became ssid: "!secret wifi_ssid", a plain string that silently fails to join wifi.

The wizard now sends empty ssid/psk in that case; the backend already emits the proper unquoted !secret tags when those fields are empty, see https://github.com/esphome/device-builder/blob/5238bfadd3421631327850bac9343ef7ee7a2762/esphome_device_builder/helpers/device_yaml/_generation.py#L240-L248. This covers both the secrets-present path (wifi screen skipped) and the kept-the-prefilled-value path on the wifi screen; a typed wifi value still passes through unchanged.

- fixes https://github.com/esphome/device-builder/issues/1490

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/wizard/wizard-step-setup.ts
+++ b/src/components/wizard/wizard-step-setup.ts
@@ -365,7 +365,8 @@ export class ESPHomeWizardStepSetup extends LitElement {
   private _onNext() {
     if (this._stage === "name") {
       if (this._hasSecretWifi) {
-        this._finish("!secret wifi_ssid", "!secret wifi_password");
+        // Empty ssid/psk tells the backend to emit unquoted !secret tags.
+        this._finish("", "");
         return;
       }
       if (this._secretWifiSsid && !this._wifiSsid) {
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

PR #861, opened 2026-06-16
Split RP2040 and RP2350 into per-chip picker filters and badges

bdraco:

The board picker treated RP2040 and RP2350 as one chip: a single "RP2040 / RP2350" filter chip, and every card badged "RP2040" even for RP2350 boards (Pico 2 / Pico 2 W, Plasma2350W, NULA RP2350). We should have modeled this the way ESP32 already models its chip variants (a per-chip discriminator on the shared platform); we did not, so users could not tell the two chips apart in the picker.

This splits the one chip into separate "RP2040" and "RP2350" filters (mirroring the per-variant ESP32 chips) and badges each card with its real chip. ESPHome lumps both under the `rp2040` platform, so the backend now carries the chip series as `esphome.mcu` and filters on it (companion backend PR below); the filter passes `mcu` through and the badge prefers `mcu` over `variant`/`platform`.

The `mcu` field is generic: it can be reused to split the libretiny families (e.g. bk72xx chips) the same way if we want chip-level filters there in the future.

- Companion backend PR: esphome/device-builder#1499

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/api/esphome-api.ts
+++ b/src/api/esphome-api.ts
@@ -1320,6 +1320,7 @@ export class ESPHomeAPI {
     query?: string;
     platform?: string;
     variant?: string;
+    mcu?: string;
     tag?: string;
     offset?: number;
     limit?: number;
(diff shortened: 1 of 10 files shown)


esphbot (approved):
No blocking issues found.

PR #862, opened 2026-06-16
Make lvgl YAML-only and stop hydrating its 14 MB body

bdraco:

Editing an `lvgl:` block in the Visual Editor is broken and unsafe. lvgl is a recursive widget tree (`pages:` -> `widgets:` -> widgets that contain `widgets:`); the catalog can't express the recursion, so the structured form can't round-trip it (`pages:` renders an empty list even when the YAML has pages) and editing in the form silently drops config. Add `lvgl` to `YAML_ONLY_SECTIONS` so the section renders the existing "edit in YAML" notice and round-trips losslessly through the YAML pane.

lvgl's per-id catalog body is also ~14 MB, and it was being fetched just to show a name or classify the section. Source YAML-only sections' metadata from the slim index instead of the body:

- `loadConfig` skips the body fetch for YAML-only sections and builds the header from the slim catalog (`entries: []` drives the notice).
- The navigator and the name-resolver controller resolve display names from the slim catalog (new `getLoadedCatalog`) and skip the body fetch for YAML-only sections.
- `yaml-hover` and autocomplete short-circuit on YAML-only sections so hovering or typing under `lvgl:` never hydrates the body.

The lvgl body is never requested. Companion backend PR fixes the bare `lvgl` catalog entry's borrowed "LVGL Number" name so the slim title reads "LVGL".

- fixes the lvgl Visual Editor breakage on the 2026.6 beta catalog

## Types of changes

## Checklist

- [x] Tests added/updated (`test/components/device/yaml-only-section.test.ts`).
- [x] `npm run lint` (tsc) and `npm test` (3363) pass.
- [x] Prettier-formatted.

### DIFF
--- a/src/components/device/device-navigator.ts
+++ b/src/components/device/device-navigator.ts
@@ -24,6 +24,7 @@ import {
   subscribeComponentCache,
 } from "../../util/component-name-cache.js";
 import { registerMdiIcons } from "../../util/register-icons.js";
+import { loadCatalog } from "../../util/yaml-completion-catalog.js";
 import {
   type YamlSection,
   categorizeSections,
@@ -40,6 +41,7 @@ import { type NavRow, resolveBucketLabels } from "./navigator-labels.js";
 import { type NavAction, renderNavSection } from "./navigator-render.js";
 import { NavigatorRevealController } from "./navigator-reveal-controller.js";
 import { navItemMatches } from "./navigator-search-match.js";
+import { YAML_ONLY_SECTIONS } from "./yaml-only-sections.js";
 
 import "@home-assistant/webawesome/dist/components/icon/icon.js";
 import "./add-automation-dialog.js";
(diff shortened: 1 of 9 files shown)


esphbot (changes requested):

bdraco:
Thanks — addressed in b257d2b:

**🟡 #1 (no re-render when loadCatalog resolves):** fixed. `loadCatalog` now `.then(() => requestUpdate())` in both the navigator and `ComponentNameResolverController.kickoff`, so a YAML-only section's friendly name replaces the raw key as soon as the slim index lands instead of waiting for an unrelated re-render.

**🟡 silent failure / 🟢 #2 (redundant catch in loadConfig):** dropped the `.catch(() => null)`. `loadCatalog` degrades to an empty index internally (logged) and never rejects, so the catch was dead; a genuine throw now reaches `loadConfig`'s existing try/catch. For a YAML-only section a miss/failure still renders the edit-in-YAML notice (with a raw-id header), which is the correct degradation since the user edits YAML there anyway.

**🟢 #1 (untested guard):** added a `resolveAvailableEntries` test pinning that a YAML-only `topLevelKey` returns `[]` and never calls `getComponentBodies` — the load-bearing "never hydrate the body" assertion.

**Copilot (add-component flow):** good catch. The companion backend PR #1510 suppresses lvgl's per-id body to ~140 bytes (config_entries dropped, since it's YAML-only), so the add-component hydration is no longer a 14 MB fetch. Making the add-component dialog itself YAML-only-aware (insert the scaffold, no structured form) is a separate UX follow-up; I scoped the description to navigate/edit/hover/section-open + the backend suppression.

esphbot (changes requested):

bdraco:
Closing — superseded by the backend body-suppression in esphome/device-builder#1510.

Once #1510 ships lvgl's catalog body with `config_entries: []`, the frontend renders it YAML-only automatically: `isYamlOnlySection` already returns true for a zero-entry section, so no hardcoded `lvgl` entry, slim-index plumbing, or per-consumer body-fetch guards are needed. The body is ~140 bytes (no 14 MB fetch), and the navigator/header show "LVGL" from the corrected name in #1506. Letting the empty body drive the behavior is the cleaner design — no frontend coupling duplicating what the catalog already encodes.

PR #866, opened 2026-06-17
Return to the method chooser when cancelling empty-config

bdraco:

Two related fixes to the new-device wizard.

**Cancel routing.** "Create empty configuration" (an advanced option on the "How would you like to create your project?" chooser) goes straight to the empty-config step; no board is picked on that path. Its Cancel button dispatched `next-step` with `"board"`, so cancelling dropped the user on the board picker. Dispatch `"method"` instead, so Cancel returns to the method chooser, matching the dialog's header back arrow which already maps `empty-config` to `method`.

**Advanced disclosure persistence.** Once Cancel returns to the method chooser, the user lands on a freshly re-mounted method step (the dialog swaps step elements via a switch), so the "Advanced" section they had opened was collapsed again. The open state is per-session, not per-mount, so it now lives on the dialog: `_advancedOpen` is owned by `create-config-dialog`, passed down as `.advancedOpen`, toggled via a `toggle-advanced` event the step emits, and reset on dialog open so a fresh wizard still starts collapsed.

Tests: empty-config Cancel routing; the method step renders advanced options from the prop and signals via the event rather than flipping local state; and a dialog-level round trip (open, toggle advanced, into empty-config, back) that pins the disclosure stays open, plus reset-on-reopen.

- fixes <link to issue>

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/wizard/create-config-dialog.ts
+++ b/src/components/wizard/create-config-dialog.ts
@@ -62,6 +62,12 @@ export class ESPHomeCreateConfigDialog extends LitElement implements ImportFlowH
   @state()
   private _step: WizardStep = "method";
 
+  // Owned here, not in the method step, so the "Advanced" disclosure stays
+  // open when the user navigates into an advanced option (empty-config /
+  // import) and back — the step element is unmounted across that transition.
+  @state()
+  private _advancedOpen = false;
+
   // Drives the step components' Enter listeners: the steps stay mounted in
   // the wa-dialog while it's merely hidden (light-dismiss / Escape / close),
   // so they must deactivate on hide, not just on unmount.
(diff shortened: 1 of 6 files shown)


esphbot (changes requested):

bdraco:
Addressed both Kōan items in `6608e8b`:

**Important #1 (description):** rewrote the PR description so it covers both changes; the Cancel routing fix and the advanced-disclosure persistence (and why the latter is a necessary consequence of the former, since Cancel now lands you back on a re-mounted method step).

**Suggestion #2 (persistence coverage):** added a dialog-level test in `create-config-dialog.test.ts` that pins the actual promise: open, toggle Advanced, navigate into empty-config (method element unmounts), navigate back to method, assert the re-mounted step still gets `advancedOpen=true`; plus a reset-on-reopen case. This fails if a future change drops the `.advancedOpen` binding or resets the flag on `next-step`.

3378 tests pass; lint + prettier clean.

esphbot (approved):
No blocking issues found.

PR #867, opened 2026-06-17
Fix misaligned Paired Senders Remove button

bdraco:
# What does this implement/fix?

The Remove control on each Paired Senders row was a text button reusing the shared `.peer-remove` class, which is sized as a 32px icon square; the text overflowed the box, so the hover background sat offset behind the text and clipped at the panel edge. This makes it the same delete icon button the offloader pairing row uses; a neutral bordered square at rest that tints red on hover, since removing a sender is destructive. The shared `.peer-remove` now carries that destructive icon-button chrome, so the offloader trash button is unchanged.

**Related issue or feature (if applicable):**

Found during manual testing of the Build server settings.

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [ ] New feature (non-breaking change which adds functionality) — `new-feature`
- [ ] Enhancement to an existing feature — `enhancement`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [ ] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `docs`
- [ ] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Checklist

- [x] The code change is tested and works locally.
- [x] `npm run lint` passes.
- [x] `npm run test` passes.
- [ ] Tests have been added to verify that the new code works (where applicable).

PR #868, opened 2026-06-17
Drop redundant .btn-unpair; .peer-remove is the single destructive icon button

bdraco:

Follow-up to #867. That PR promoted the shared `.peer-remove` into the destructive icon button, which left it restating, byte-for-byte, the chrome `.btn-unpair` already defined in `offload-styles.ts`; the same look lived in two stylesheets and could drift. `.btn-unpair` was only ever used on the offloader trash button, which already carries `.peer-remove`, so this drops the redundant class and its rules and leaves `.peer-remove` as the single source. No visual change; the trash button, the Edit pencil, and the Paired Senders Remove all render exactly as before.

Follow-up to #867 (addresses the duplication noted in review).

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/settings-dialog/build-offload-pairing-row.ts
+++ b/src/components/settings-dialog/build-offload-pairing-row.ts
@@ -147,7 +147,7 @@ export function renderPairingRow(
           : nothing}
         <button
           type="button"
-          class="peer-remove btn-unpair"
+          class="peer-remove"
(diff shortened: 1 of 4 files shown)


esphbot (approved):
No blocking issues found.

PR #870, opened 2026-06-17
Reconnect Web Serial logs with a fresh port after a flash reset

bdraco:

After flashing a native-USB ESP32 (C3 / S3 / C6) over Web Serial, the post-install "Logs" viewer failed to attach on **Chrome** ("Failed to reopen the serial port for logs…") and clicking Start did nothing useful, while **Firefox** showed the full boot log.

Root cause: the install→logs handoff reopened the same `SerialPort` esptool used for flashing. The post-flash reset re-enumerates the USB device. On Chrome the cached handle then stays dead (`open()` throws `NetworkError` indefinitely), whereas `navigator.serial.getPorts()` returns a fresh, openable handle for the same authorized device. Firefox reopens the cached handle directly, which is why it worked there.

Fix: when a closed port is reopened for logs, resolve the **live** port from `getPorts()` (matching the device's USB vendor/product, falling back to the cached handle for Firefox and non-re-enumerating UART bridges) and open that, retrying every 200 ms across the re-enumeration window. The logs now reconnect automatically on both browsers with no picker. A user-gesture reconnect via `requestPort()` on "Start" remains the last resort if the device is truly gone (a picker needs a click; `getPorts()` does not). The reopen-failure message now names the port (USB vendor:product).

Confirmed on hardware: on Chrome, `getPorts()` returns one port that opens after the C3 re-enumerates, while the cached handle's `open()` throws at the same time.

- fixes <link to issue>

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/dashboard/actions-ui.ts
+++ b/src/components/dashboard/actions-ui.ts
@@ -211,7 +211,6 @@ export async function openLogsWithMethod(
       return;
     }
     if (!serialPort) return; // User dismissed the port picker.
-    const port = serialPort;
     host._logsDialog.configuration = device.configuration;
     host._logsDialog.name = device.friendly_name || device.name;
     // Reconnect (the dialog's "click Start to reconnect") re-acquires a fresh
@@ -222,7 +221,7 @@ export async function openLogsWithMethod(
     // attach toasts the reopen-retry failure itself; cover any other rejection
     // so it can't escape this fire-and-forget call as an unhandled rejection.
     try {
-      await attachSerialLogStream(port, host._logsDialog, host._localize);
+      await attachSerialLogStream(serialPort, host._logsDialog, host._localize);
     } catch {
       toast.error(host._localize("dashboard.logs_web_serial_open_failed"), {
         richColors: true,
--- a/src/components/dashboard/actions-ui.ts
+++ b/src/components/dashboard/actions-ui.ts
@@ -10,7 +10,11 @@ import type { ESPHomePageDashboard } from "../../pages/dashboard.js";
 import { getErrorMessage } from "../../util/error-message.js";
 import { firmwareJobDisplayName } from "../../util/firmware-job-display.js";
 import { clearJustCreated } from "../../util/just-created.js";
-import { attachSerialLogStream } from "../../util/post-install-logs.js";
+import {
+  attachSerialLogStream,
+  reconnectWebSerialLogs,
+  requestAndOpenSerialPort,
+} from "../../util/post-install-logs.js";
 
 export async function executeFriendlyName(
   host: ESPHomePageDashboard,
@@ -195,18 +199,9 @@ export async function openLogsWithMethod(
       });
       return;
     }
-    let serialPort: SerialPort;
+    let serialPort: SerialPort | null;
     try {
-      serialPort = await (
-        navigator as unknown as {
-          serial: { requestPort: () => Promise<SerialPort> };
-        }
-      ).serial.requestPort();
-    } catch {
-      return; // User dismissed the port picker.
-    }
-    try {
-      await serialPort.open({ baudRate: 115200 });
+      serialPort = await requestAndOpenSerialPort();
     } catch {
       // The user picked a port but it couldn't open (claimed by another tab,
       // driver error); unlike a picker dismissal this needs feedback.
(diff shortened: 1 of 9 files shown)


bdraco:
Thanks — addressed the Copilot comments in `bd6356f`:

- **requestPort cancel vs real failure** (comments on L6/L38): `requestAndOpenSerialPort` now reuses `isPortPickerCancel()` from `web-serial.ts` — it returns `null` only for a genuine dismissal (`NotFoundError`) and rethrows anything else, so a SecurityError/InvalidStateError surfaces instead of being swallowed. Added a test for the rethrow path.
- **`matchesDevice` undefined match** (L153): now requires both `usbVendorId` and `usbProductId` to be present, so two non-USB ports don't match as the same device via `undefined === undefined`.
- **`formatSerialPortLabel` false "USB"** (L25): the no-USB-ids fallback is now a neutral "the serial port" rather than "USB serial device".
- **"any other granted port" fallback** (L191): already removed in `09138f5` (a `/simplify` pass) — `openLiveSerialPort` only considers same-device-VID:PID handles, then the cached one, never an unrelated granted port.

Full suite + lint + prettier green.

esphbot (changes requested):

bdraco:
Thanks Kōan. This review was on `09138f5`; two of the three items were already fixed in `bd6356f`, and the last two are now in `e067bcd`:

- **Warning #1 (cancel vs real failure):** already addressed in `bd6356f` — `requestAndOpenSerialPort` uses `isPortPickerCancel()` and rethrows non-cancel `requestPort` errors (with a test for the rethrow path).
- **Suggestion #1 (matchesDevice):** the `undefined === undefined` guard landed in `bd6356f`; `e067bcd` adds the comment that VID:PID can't disambiguate two identical boards (Web Serial exposes no per-device serial; the post-install flow is single-device anyway).
- **Suggestion #2 (lost breadcrumb):** restored in `e067bcd` — `openLiveSerialPort` keeps the last `open()` error and `console.error`s it once when the re-enumeration window elapses, so a genuinely flaky board still leaves the DOMException name/message in the console. Added a test asserting the breadcrumb fires.

Also confirming your note: Copilot's "any other granted port" concern doesn't apply — candidates are strictly the same-VID:PID granted handles plus the cached one.

Full suite + lint + prettier green.

esphbot (approved):
No blocking issues found.

bdraco:
Addressed the latest Copilot + Kōan notes in `ba52183`:

- **Fail fast on non-recoverable opens** (Copilot, and Kōan suggestion #2): `openLiveSerialPort` now bails immediately on a non-`NetworkError` open failure (logging it), instead of retrying the whole ~8s window and then showing the misleading "still restarting" message. `NetworkError` (device gone mid-re-enumeration) and the "already open" `InvalidStateError` are still tolerated. Split the timeout test into a fast-bail case and a NetworkError-retry case.
- **`abortSerialReconnect` guard** (Copilot): now only acts while `reconnecting`, never on a live `serial` session — so it can't drop to `dead` and leak an open reader/port.
- **Fallback label** (Kōan suggestion #1): no-USB-id ports now read "unknown device" rather than "the serial port", so the message isn't "...for logs (the serial port)."

Full suite + lint + prettier green.

esphbot (approved):
No blocking issues found.

bdraco:
Addressed the latest test-hygiene notes (Copilot + Kōan, both on `ba52183`) in `44f7899`:

- **`withRequestPort` / `withGetPorts` restore** (Copilot L67/L77, Kōan suggestion #1): they now record whether `navigator.serial` existed beforehand and `delete` the injected property on restore when it didn't (happy-dom has none), instead of leaving it present with value `undefined` — so a leaked `"serial" in navigator` can't make later tests order-dependent. Factored into a shared `restoreSerial` helper.
- **Stale file header** (Copilot L7): updated to describe the actual behavior — the auto reopen prefers `navigator.serial.getPorts()` (no picker) and only the user-gesture "Start" reconnect re-prompts via `requestPort()`.

Both are test-only; no production change. Kōan flagged the PR merge-ready otherwise, and confirmed Copilot's earlier L202 "any other granted port" concern doesn't apply to the final code. Full suite + lint + prettier green.

esphbot (approved):
No blocking issues found.

PR #871, opened 2026-06-17
Resume USB device setup when actioned from another route

bdraco:

The global "USB device connected" toast's "Set it up" action did nothing when the user was inside the device editor or the secrets page; it only worked on the main dashboard. The toast is mounted app-wide, but its action dispatches a `esphome-serial-setup` window event whose only listener, and the create-config wizard it opens, live on the dashboard page; the router unmounts the dashboard on every other route, so off the dashboard the event had no handler and the dialog was not in the DOM.

Now the action branches on the current route; on the dashboard it dispatches as before, and elsewhere it stashes the live `SerialPort` and navigates home, where the dashboard consumes it on mount and resumes the same detect-and-open-wizard flow. The editor and secrets pages install a leave guard that can veto the navigation; in that case the stash is cleared so a stale port cannot fire on a later, unrelated dashboard mount.

Found during manual testing; clicking "Set it up" from inside the editor.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/app-shell.ts
+++ b/src/components/app-shell.ts
@@ -52,7 +52,13 @@ import {
   versionContext,
 } from "../context/index.js";
 import { espHomeStyles } from "../styles/shared.js";
+import { stripBase } from "../util/base-path.js";
 import { isExpert } from "../util/experience.js";
+import { navigate } from "../util/navigation.js";
+import {
+  consumePendingSerialSetup,
+  markPendingSerialSetup,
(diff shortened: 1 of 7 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #872, opened 2026-06-17
Fix Web Serial flashing of ESP8266 (flash at 0x0, boot-mode reset)

### LINKED ISSUE #1529: Flashing ESP8266 over web serial fails
### The problem

I wanted to flash an ESP8266 on Windows using the beta backend; the flashing process works, but the ESP becomes unresponsive afterwards. However, if I download the file and flash it via web.esphome.io or the classic backend, it works fine. I am using the standard configuration yaml, and the same issue occurs with the Device Builder beta in Home Assistant.

### Which version of ESPHome Device Builder has the issue?

ESPHome Device Builder v0.1.0b116 ESPHome 2026.5.3

### What platform are you using?

Windows

### Anything in the logs that might be useful for us?

```txt

```

### Additional information

_No response_

bdraco:

Flashing an ESP8266 over Web Serial completed but the chip never booted; downloading the binary and flashing via web.esphome.io or the classic backend worked. Two root causes, both ESP8266/UART-bridge specific:

**1. Wrong flash address (primary).** The install flow defaulted to `0x10000` (the ESP32 app partition) and only used `0x0` when a binary's filename contained `"factory"`. ESP8266 ships a single complete `firmware.bin` (PlatformIO's `esp8266_copy_factory_bin` output) with no `"factory"` variant, so it was written at `0x10000` — leaving the boot address empty. The serial log confirmed it: `Wrote 455760 bytes ... at 0x10000`. New `pickFlashTarget` flashes a merged `*factory*` image or any **esp82xx** (ESP8266 / ESP8285) `firmware.bin` at `0x0`; an ESP32 app-only image stays at `0x10000`. `chipName` is esptool-js's *description* (`loader.main()` returns `getChipDescription()` → `ESP8266EX` / `ESP8285`, not `ESP8266`), so it's normalized via `chipNameToVariant` and matched on the `esp82` family.

**2. Wrong post-flash reset.** The classic/UART-bridge branch used esptool-js's `ClassicReset`, which is the *enter-bootloader* sequence (drives GPIO0 low as it releases EN), leaving the chip in the serial bootloader. Replaced with `classicHardReset` — an EN pulse with GPIO0 released — matching esptool's `--after hard-reset`.

Both match the legacy dashboard (`esphome/dashboard`): its web-serial flasher always wrote a complete image at **`address: 0`** for every chip, and its `resetSerialDevice` is the same RTS pulse. The watchdog-reset path (ESP32-S2/S3/C2/C3) and native USB-Serial/JTAG path are unchanged.

- fixes https://github.com/esphome/device-builder/issues/1529

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/firmware-install-dialog/install-flow.ts
+++ b/src/components/firmware-install-dialog/install-flow.ts
@@ -34,6 +34,33 @@ export function compileFailureDetail(err: unknown): string {
   return err instanceof Error ? err.message.trim() : String(err ?? "").trim();
 }
 
+/**
+ * Choose which binary to flash over Web Serial and its flash offset.
+ *
+ * ESP8266 / ESP8285 flash a single complete ``firmware.bin`` at 0x0; ESP32 uses
+ * a merged ``*.factory.bin`` (bootloader + partitions + app) at 0x0, falling
+ * back to the app image at 0x10000. Flashing an ESP8266 image at 0x10000 leaves
+ * the boot address empty so the chip never boots (#1529). Returns ``null`` when
+ * there's no binary to flash.
+ *
+ * ``chipName`` is esptool-js's chip *description* (``loader.main()`` returns
+ * ``getChipDescription()`` — e.g. ``ESP8266EX`` / ``ESP8285``, not ``ESP8266``),
+ * so normalize it via ``chipNameToVariant`` and match the ``esp82`` family.
+ */
+export function pickFlashTarget(
+  chipName: string,
+  binaries: FirmwareBinary[]
+): { binary: FirmwareBinary; address: number } | null {
+  const factory = binaries.find((b) => b.file.includes("factory"));
+  const binary = factory ?? binaries[0];
+  if (!binary) return null;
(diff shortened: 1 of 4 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

bdraco:
Addressed the test-hygiene notes (Copilot + Kōan) in `78ef95f`:

- **Vacuous DTR assertion** (both): added a positive `expect(transport.setDTR).toHaveBeenCalledWith(false)` alongside the existing `every(...)` check, so the test fails if the classic reset ever stops releasing GPIO0 entirely.
- **Dead eslint-disable** (Kōan): removed — the repo has no eslint and the file uses no `any` (just `as unknown as`).

Skipped the optional factory-detection-by-`type`-tag suggestion: the filename-substring approach mirrors the legacy flasher and the backend tag values aren't verified here, so it stays as-is.

3396 tests pass; lint + prettier clean.

PR #873, opened 2026-06-17
Hide browser Web Serial install for non-ESP platforms (RP2040 / RP2350, nrf52, libretiny)

bdraco:

The install dialog offered the **browser Web Serial** flash option for an RP2350/RP2040 device — the user saw the RP2350B in the browser's serial-port picker, and flashing failed.

Browser Web Serial uses esptool-js, which speaks only the esptool ROM protocol. It can't flash:
- **RP2040 / RP2350** — enter the bootloader via a BOOTSEL / 1200-baud touch and flash by UF2 mass-storage copy;
- **nrf52** — UF2 / serial-DFU;
- **libretiny** (bk72xx / rtl87xx / ln882x) — ltchiptool's own serial protocol.

These flash over serial only via the **backend** (`esphome run` / the dialog's server-serial method), never the browser. So the Web Serial row was dead for them.

This gates the Web Serial row on the platform: the existing ESP-only check (used for the web.esphome.io download row) is generalized into `_isEsptoolPlatform` (`esp32*` / `esp82*` — covers esp8266 and esp8285), and Web Serial renders only for those. Server-serial stays available for non-ESP platforms, including on localhost where it would otherwise collapse into the now-hidden Web Serial row. OTA (when online) and the UF2 download (advanced section) are unaffected.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/install-method-dialog.ts
+++ b/src/components/install-method-dialog.ts
@@ -112,14 +112,19 @@ export class ESPHomeInstallMethodDialog extends LitElement {
   }
 
   /**
-   * web.esphome.io / esp-web-tools only supports ESP32 (all variants)
-   * and ESP8266. UF2 platforms (RP2040, nrf52, libretiny) ship a
-   * different binary format that the browser flasher can't handle, so
-   * we hide the option entirely for those.
+   * ESP32 (all variants) and ESP8266 / ESP8285 — the only chips the browser
+   * flashers can handle. Both Web Serial (esptool-js) and web.esphome.io /
+   * esp-web-tools speak the esptool ROM protocol. Non-ESP targets can't be
+   * flashed from the browser: RP2040 / RP2350 and nrf52 use a BOOTSEL /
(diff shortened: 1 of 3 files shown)


esphbot (changes requested):

bdraco:
fail closed is the right shape here. we only support esp chips for serial flash

bdraco:
Re Kōan warning #1 (empty/unknown `target_platform`): confirmed intentional — fail-closed is the desired behavior. We shouldn't offer a browser flasher that won't work for an unknown platform; server-serial and OTA still render, and `target_platform` is reliably populated for configured devices, so this only hides Web Serial transiently before metadata loads.

Locked it in with a comment on `_isEsptoolPlatform` in `2b31791`.

Copilot generated no comments. 3419 tests pass; lint + prettier clean.

esphbot (approved):
No blocking issues found.

PR #875, opened 2026-06-17
Warn to keep the window visible during a Web Serial flash

bdraco:

Navigating away from the tab/window during a browser (Web Serial) flash sometimes fails the flash. Browsers throttle `setTimeout`/`setInterval` to ≥1s in hidden tabs, which disrupts esptool-js's timing-sensitive serial writes — and there's no JS API to opt a normal page out of that throttling, so it can't be prevented technically. The correct mitigation is a user warning, which is what the legacy dashboard did ("Keep this page visible to prevent slow down", shown during the writing phase).

This shows a "Keep this window visible …" note in the install dialog's status subtext during the `flashing` step. That step is the **Web Serial** path only — OTA / server-serial flash on the backend and aren't affected by browser visibility — so the warning is correctly scoped (it doesn't appear for those install methods).

A Screen Wake Lock was considered and rejected: it only prevents display sleep, not hidden-tab timer throttling (the actual failure), so it wouldn't fix the "navigated away" case.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/firmware-install-dialog/renderers.ts
+++ b/src/components/firmware-install-dialog/renderers.ts
@@ -134,6 +134,9 @@ export function cardStatusDetail(host: ESPHomeFirmwareInstallDialog): string {
   }
   if (host._step === "download-ready") return downloadReadyDetail(host);
   if (host._step === "error") return host._errorMessage;
+  // Hidden tabs throttle timers, which can stall the Web Serial write and fail
+  // the flash; there's no API to opt out, so warn the user to stay on the page.
+  if (host._step === "flashing") return host._localize("firmware.flashing_keep_visible");
   return "";
 }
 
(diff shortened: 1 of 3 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
Addressed Kōan suggestion #1 in `cde74de`: swapped the bogus `"resetting"` step (not a real `InstallStep`) for genuine non-flashing steps (`connecting` / `queued` / `downloading`), and added a comment noting the restart phase keeps `_step === "flashing"` (only the status message changes) so the warning correctly persists there. Copilot had no comments. 3421 tests pass, lint + prettier clean.

PR #877, opened 2026-06-17
Keep the serial logs viewer responsive under a flood

bdraco:

Flashing a verbose device over Web Serial (an ESP8266 with `logger: baud_rate: 19200`) could fill the logs viewer with pages of fragmented garbage and lock up the browser tab until the device was reset. Two distinct problems:

1. **Resilience.** The serial path appended one line at a time (`dialog._lines = [...dialog._lines, line]`), each triggering a full re-render of an unbounded list, so a fast stream froze the tab. This batches appends into one render per animation frame (mirroring the existing `command-dialog` batcher) and caps the retained buffer to the newest 5000 lines. The view now stays responsive regardless of input rate.

2. **Garbage source.** An ESP8266 prints its boot/reset banner at a fixed 74880 baud; read at the app's configured baud it mis-samples into mojibake (U+FFFD replacement chars, control bytes), which recurs on every reset. Lines that are mostly such bytes are now dropped before reaching the parser, so they neither pollute the carried color/prefix state nor flood the view. Clean ESPHome logs are printable ASCII plus ANSI and pass through untouched.

Note: the filter removes the bulk of the mojibake, but an occasional clean-looking fragment (a bare `m` from a split reset, a short tail) can still slip through; the cap and batching make whatever slips through harmless. Fully eliminating ESP8266 boot-baud garbage would mean reading the boot at 74880 then switching to the app baud, which most tools don't do; out of scope here.

Found while testing serial logs on an ESP8266 with a non-default logger baud.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/dashboard/actions.ts
+++ b/src/components/dashboard/actions.ts
@@ -6,7 +6,7 @@ import type { ArchivedDevice, BulkActionResult } from "../../api/types/system.js
 import type { LocalizeFunc } from "../../common/localize.js";
 import { withBase } from "../../util/base-path.js";
 import { getErrorMessage } from "../../util/error-message.js";
-import { ESPHomeLogParser } from "../../util/esphome-log-parser.js";
+import { ESPHomeLogParser, isLikelyGarbageLine } from "../../util/esphome-log-parser.js";
 import {
   connectToPort,
   detectChip,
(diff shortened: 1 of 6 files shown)


esphbot (changes requested):

esphbot (approved):
No blocking issues found.

PR #880, opened 2026-06-17
Reset checkbox display state when reusing a dialog

### LINKED ISSUE #1535: [Bug] Edit Friendly name "Install Immediately" option persistence issue
### What happened?

1. Open the ESPHome dashboard
2. Open the 3-dot/expansion menu for any device
3. Select "Edit friendly name"
4. Edit the name
5. Uncheck the "Install immediately..." option.
6. Click "Save"
7. Repeat the above through Step 4 for any device. Option already appears unchecked.

<br>

Image

<br>

8. Click "Save". Even though the option is unchecked, the user is presented with the "How do you want to install the firmware?" popup. 

Image

<br>

The only way to avoid this is to check then re-uncheck the option.

### Dashboard version

0.1.0b113

### How did you install it?

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

Fixes a checkbox state desync in two reused singleton dialogs. The "Install immediately" box in Edit friendly name and the encryption box in the adopt dialog were bound with an attribute binding (`?checked`), which only feeds the checkbox's `defaultChecked`; once the user clicks the box it latches its own dirty state and ignores the attribute. So reopening the dialog after an uncheck left the box visually unchecked while the backing state had already reset to checked, and Edit friendly name then fired the install flow even though the box looked unchecked. Switching both to a property binding (`.checked`) makes the box track the reset state on every reopen.

Added a regression test for the adopt dialog; the friendly-name dialog mocks out the WebAwesome checkbox in tests so the sticky state can't be reproduced there, verified that path in the browser instead.

- fixes https://github.com/esphome/device-builder/issues/1535

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/adopt-dialog.ts
+++ b/src/components/adopt-dialog.ts
@@ -328,7 +328,7 @@ export class ESPHomeAdoptDialog extends LitElement {
               <label class="checkbox-row">
                 <input
                   type="checkbox"
-                  ?checked=${this._encryption}
+                  .checked=${this._encryption}
                   ?disabled=${this._busy}
                   @change=${(e: Event) => {
                     this._encryption = (e.target as HTMLInputElement).checked;
--- a/src/components/friendly-name-dialog.ts
+++ b/src/components/friendly-name-dialog.ts
@@ -222,7 +222,7 @@ export class ESPHomeFriendlyNameDialog extends LitElement {
         </div>
         <div class="install-row">
           <wa-checkbox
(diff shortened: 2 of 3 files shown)


esphbot (approved):
No blocking issues found.

PR #881, opened 2026-06-18
Show ESPHome version on paired build server rows

bdraco:

The paired build server rows in Settings only showed the host and port, so there was no way to tell which ESPHome version a connected build server runs; the Known Dashboards list right below it already shows that. This adds the version to each paired row, reusing the same line. It only renders once the receiver has handshaked, and it stays out of the way when the versions differ since the existing mismatch note already states both versions.

It also keeps the shown version fresh. The backend already sends `esphome_version` on `offloader_peer_link_opened`, but the frontend dropped it, so a row kept its snapshot value until a full reload. The event payload now carries it and the handler merges it, so an upgraded build server shows up on the next reconnect without a reload.

- Related to https://github.com/esphome/device-builder/issues/1545

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/api/types/event-subscription.ts
+++ b/src/api/types/event-subscription.ts
@@ -68,8 +68,8 @@ export enum DeviceEventType {
   // to the receiver enters / leaves the post-handshake parked
   // state. Drives the PairingSummary.connected indicator on
   // the offloader-side Paired-build-servers list. Both events
-  // share the same OffloaderPeerLinkSessionEventData shape;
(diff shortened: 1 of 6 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #903, opened 2026-06-19
Remove redundant height:100% from device-card host

### LINKED ISSUE #1581: [Bug] extremely large height cards Safari 17.6
## Bug report

In Safari 17.6 on macOS, the device cards in ESPHome Device Builder are rendered with an extremely large height.

The cards stretch almost to the full height of the browser window, leaving a lot of empty space below each device.

### Environment

- Browser: Safari 17.6
- OS: macOS
- Page: ESPHome Device Builder device list
- Layout: grid/card view

### Expected behavior

Device cards should have a compact height based on their content.

### Actual behavior

Each device is displayed inside a very tall card/container.

### Screenshot

Attached screenshot shows the issue.

### Dashboard version

2026.6.1

### How did you install it?

### Where is it running?

Local — Linux

### Relevant logs / errors

```shell

```

### Anything else?

Image

bdraco:

Remove the redundant `height: 100%` from the `<esphome-device-card>` `:host`.

The device grid (`.devices-grid`) already equalizes card heights via
`align-items: stretch` (the grid default), and the inner `.device-card` fills
the host, so a height on the host adds nothing. It is also fragile: a percentage
height on a grid item resolves against the row track, and WebKit can compute that
as the viewport-tall grid box rather than the auto content track, stretching each
card to nearly the full window height with a large empty gap below the content.

Reported on Safari 17.6, which is below the supported baseline (`browserslist`
"last 2 Safari versions" currently resolves to 26.3/26.4), but the rule was
redundant and fragile, so removing it is correct regardless. Layout is identical
in Chromium with and without the rule (measured 169px both ways at the reporter's
viewport), so there is no behaviour change in supported browsers.

- fixes https://github.com/esphome/device-builder/issues/1581

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device-card/styles.ts
+++ b/src/components/device-card/styles.ts
@@ -17,7 +17,12 @@ export const deviceCardStyles = [
     :host {
       display: block;
       outline: none;
-      height: 100%;
+      /* Intentionally no height: 100%. The .devices-grid already equalizes
+         card heights via align-items: stretch (its default) and the inner
+         .device-card fills this host, so a height here is redundant. It is
+         also fragile: a percentage height on a grid item resolves against the
+         track, which WebKit can compute as the viewport-tall grid box instead
+         of the auto content track, stretching the card to full-window height. */
     }
 
     .device-card {


esphbot (approved):
No blocking issues found.

PR #910, opened 2026-06-19
Try the cachedPort fallback on a non-NetworkError open failure

### LINKED ISSUE #909: openLiveSerialPort skips the cachedPort fallback on a non-NetworkError open failure
## Summary

`openLiveSerialPort` in `src/util/post-install-logs.ts` abandons its remaining port candidates (including the `cachedPort` fallback) when a freshly-granted handle fails to open with anything other than `NetworkError`. The fail-fast is applied at the wrong granularity, so a still-usable UART-bridge port is never tried.

## Location

`src/util/post-install-logs.ts`, in `openLiveSerialPort` (around lines 207-233):

```ts
const candidates = [
  ...granted.filter((p) => p !== cachedPort && matchesDevice(p.getInfo(), want)),
  cachedPort,
];
for (const p of candidates) {
  if (p.readable) return p;
  try {
    await p.open({ baudRate });
    return p;
  } catch (err) {
    lastErr = err;
    const name = err instanceof DOMException ? err.name : "";
    const message = err instanceof Error ? err.message : "";
    if (name === "InvalidStateError" && /already open/i.test(message)) {
      return p;
    }
    if (name !== "NetworkError") {
      console.error("[Web Serial] Failed to reopen port for logs:", err);
      return null;   // <-- abandons the rest of `candidates`, incl. cachedPort
    }
  }
}
```

## Impact

The candidate order is `[fresh re-enumerated handles..., cachedPort]`. The whole point of the loop is the `cachedPort` fallback (Firefox / UART bridges that reopen the original handle in place). If a freshly-granted handle fails to open with a non-`NetworkError` (e.g. a `SecurityError`, or a phantom handle on a board that exposes both a native USB-CDC and a separate UART bridge), this returns `null` before `cachedPort` is ever attempted, and post-install logs silently go unavailable even though the cached UART port would have opened.

Niche (needs a fresh handle that errors non-`NetworkError` while the cached port is good), but a real correctness gap: fail-fast should happen only after exhausting this round's candidates, not skip the fallback the function exists to provide.

## How it surfaced

The dev USB flasher (esphome/device-builder#1601) copied this function; automated review (Copilot + Kōan) flagged the early-return there, which is fixed in that PR. The frontend original predates it and was not touched by the flash-via-usb work.

## Origin

Introduced in #870 ("Reconnect Web Serial logs with a fresh port after a flash reset"). Live on `main`.

## Proposed fix

Mirror the flasher fix: try every candidate each round (fall through on any open failure), capture the last error, and only give up at the deadline. Surface the captured error in the failure path instead of the generic timeout message. Add a unit test over `openLiveSerialPort` covering: fresh-handle non-`NetworkError` then a good `cachedPort`, all-`NetworkError` until deadline, and immediate success.

Optional follow-up: prefer the genuinely new (re-enumerated) handle over VID/PID-only `matchesDevice` by snapshotting authorized ports before the reset, so two identical boards don't cross logs.

bdraco:

`openLiveSerialPort` (`src/util/post-install-logs.ts`) reopens the serial port after a post-install hard reset by walking a candidate list `[fresh re-enumerated handles…, cachedPort]`. The whole point of the loop is the `cachedPort` fallback that Firefox and UART bridges reopen in place. But the open loop early-returned `null` the moment a candidate failed to open with anything other than `NetworkError`, abandoning the rest of the list, including `cachedPort`.

So a freshly-granted handle that fails non-`NetworkError` (a `SecurityError`, or a phantom handle on a board that exposes both a native USB-CDC and a separate UART bridge) made the function give up before the cached UART port was ever tried, and post-install logs went silently unavailable even though the cached port would have opened.

The fail-fast was at the wrong granularity. This falls through to the next candidate on any open failure and gives up only at the deadline, surfacing the captured error instead of a generic timeout. It mirrors the fix already made in the dev USB flasher copy of this function (esphome/device-builder#1601); the frontend original predates that work and was not touched by it.

Trade-off, intended: a non-`NetworkError` that never recovers now waits out the re-enumeration window before failing rather than bailing immediately, which is the cost of guaranteeing the `cachedPort` fallback is always attempted.

A unit test pins the gap (fresh non-`NetworkError` handle then a good `cachedPort` streams the cached port); the existing all-`NetworkError`-to-deadline and immediate-success tests stay green.

- Fixes esphome/device-builder-frontend#909

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/util/post-install-logs.ts
+++ b/src/util/post-install-logs.ts
@@ -221,15 +221,10 @@ async function openLiveSerialPort(
         if (name === "InvalidStateError" && /already open/i.test(message)) {
           return p;
         }
-        // NetworkError means the device is still gone mid-re-enumeration: keep
-        // retrying the next candidate / round. Anything else (claimed by
-        // another app, driver / security error) won't fix itself by waiting —
-        // fail fast rather than stall the whole window behind a misleading
-        // "still restarting" message.
-        if (name !== "NetworkError") {
-          console.error("[Web Serial] Failed to reopen port for logs:", err);
-          return null;
-        }
+        // Unusable this round — still re-enumerating (NetworkError), claimed by
+        // another app, or a transient driver / security error. Fall through to
+        // the next candidate (including the cached handle this function exists
+        // to fall back to) and only give up at the deadline below.
       }
     }
     if (Date.now() >= deadline) {
--- a/test/util/post-install-logs.test.ts
+++ b/test/util/post-install-logs.test.ts
@@ -38,8 +38,9 @@ function openPort(
   } as unknown as SerialPort;
 }
 
-// A closed port whose open() always rejects. Defaults to a non-NetworkError
-// (the reopen bails fast); pass a NetworkError to exercise the retry window.
+// A closed port whose open() always rejects. Defaults to a non-NetworkError;
+// the reopen falls through to the next candidate either way, so pass a
+// NetworkError only when a test cares about the error kind.
 function deadPort(
   error: unknown = new DOMException("blocked", "SecurityError")
 ): SerialPort {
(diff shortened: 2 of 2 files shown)


esphbot (approved):
No blocking issues found.

PR #913, opened 2026-06-20
Add the missing ESP32 variant chips to the board picker

bdraco:
# What does this implement/fix?

ESP32-P4 was missing from the board picker's platform filter chips, so the ESP32-P4 boards already in the catalog (esp32-p4, esp32-p4-evboard, generic-esp32p4, m5stack-tab5-p4, and the rev.300 variants) couldn't be filtered to or surfaced under a P4 chip. This adds the `esp32p4` entry to `WIZARD_BOARD_PLATFORMS`, alongside the other ESP32 variants. The variant was already known to `chip-variant.ts` and the translations, so this one line is all that was missing.

**Related issue or feature (if applicable):**

- 

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [ ] New feature (non-breaking change which adds functionality) — `new-feature`
- [ ] Enhancement to an existing feature — `enhancement`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [ ] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `docs`
- [ ] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Checklist

- [x] The code change is tested and works locally.
- [x] `npm run lint` passes.
- [x] `npm run test` passes.
- [ ] Tests have been added to verify that the new code works (where applicable).

PR #915, opened 2026-06-20
Add the missing ESP32 variant chips (C2/C5/C61) and tighten the filter row

bdraco:

Follow-up to #913. The board-picker filter row was missing chips for the ESP32-C2, ESP32-C5, and ESP32-C61 variants even though the catalog already has boards for each (C2, C5, C61 devkits), so those boards were unfilterable. This adds the three chips; P4 already landed in #913. The variants are already wired through `en.json` and `chip-variant.ts`, so the chips filter and the serial-detect mapping work with no other change.

Ten ESP32 chips on one row gets long, so the row is tightened a little to keep the wrap reasonable: the gap drops from 6px to 4px and each chip's horizontal padding from 12px to 8px. The full `ESP32-` labels are kept. This is the minimal change; a proper redesign of the row is a separate follow-up since variants keep getting added.

- 

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


bdraco:
This whole thing needs a redesign as we are going to keep getting new ones. Trying to make the most minimal change possible here

esphbot (changes requested):

PR #919, opened 2026-06-20
Prune the dead web-download install path

bdraco:
# What does this implement/fix?

#904 consolidated the install picker's USB option onto `web-flash` (the external secure-context flasher), so nothing dispatches the old `web-download` installer anymore. The whole web.esphome.io "download + manual flash" path was unreachable and tracked for removal in #907; this is that focused, behaviour-free removal.

Removed:

- the `"web-download"` `Installer` union member and `installWebDownload()`
- the now-always-false `isWebFlasher` branch in `startDownload` (`installBinaryDownload` is its only remaining caller; `failNoBinaries` keeps the `isWebFlasher` param for the live `web-flash` path)
- the web-download `download-ready` rendering (title / detail / status-extra / footer) and its orphaned `web_download_*` + `install_method_web_download_desc` translation keys
- the web-download test cases; the `no_flashable_binary` assertion is preserved on the live `web-flash` path (usb-flash test)

`binary-download` (manual download via Advanced options) and `web-flash` are untouched. `npm run lint` and `npm run test` pass.

**Related issue or feature (if applicable):**

- closes #907

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [ ] New feature (non-breaking change which adds functionality) — `new-feature`
- [ ] Enhancement to an existing feature — `enhancement`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [x] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `docs`
- [ ] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Checklist

- [x] The code change is tested and works locally.
- [x] `npm run lint` passes.
- [x] `npm run test` passes.
- [x] Tests have been added to verify that the new code works (where applicable).

Issue #907, opened 2026-06-19
Prune the dead web-download install path


bdraco:
## Prune the dead `web-download` install path

PR #904 consolidated the install picker's USB option and switched the
`web-download` method to `web-flash` (the external secure-context flasher).
Nothing dispatches `web-download` anymore, so the whole web.esphome.io
"download + manual flash" installer is now unreachable. It was left in place
deliberately (the prune cascades through the install-dialog state machine and
its test), tracked here so it doesn't quietly rot, since a future reader sees
live-looking references and won't realize the entry point is gone.

### Now-dead code to remove

- `src/components/apply-install-method.ts` — no `web-download` case remains (already gone); confirm nothing re-adds it.
- `src/components/firmware-install-dialog.ts`
  - `Installer` union member `"web-download"` (L67)
  - `installWebDownload()` (L148-150)
- `src/components/firmware-install-dialog/install-flow.ts`
  - the `isWebFlasher = host._installer === "web-download"` branch in `startDownload` (L307) and the `isWebFlasher` plumbing through `failNoBinaries` / binary selection (L286-323)
- `src/components/firmware-install-dialog/renderers.ts`
  - the web-download rendering (~L193, L269)
- `src/components/firmware-install-dialog-download-binary.test.ts` (and any other test) — drop the cases that exercise the `"web-download"` installer
- `src/translations/en.json` — `install_method_web_download_desc` becomes orphaned once `renderers.ts:193` is removed; delete it then (its siblings `install_method_web_download` / `_link` were already removed in #904)

### Notes

- Keep `binary-download` (the manual "download the binary, flash with your own tool" path) — it's still live via the Advanced options.
- Do this as its own focused PR after #904 lands, so it's reviewable as one behavior-free removal.
- Context: part of esphome/backlog#151.

PR #920, opened 2026-06-20
Fix USB hand-off error banner and pop-up-blocked recovery

bdraco:

Addresses the two Copilot review comments that landed on #904 right after it merged. Both are on the external USB flash hand-off (`install-flow.ts` / `renderers.ts`):

1. **Stale failure banner during an in-tab retry.** After a non-terminal flasher `error`, an in-tab retry resumes through `progress` / `status` frames, but `_statusMessage` stayed "USB install failed" (set by `_fail`) until a `status` frame arrived. A `progress` frame usually lands first, so the UI showed a failure headline over a flash that was actually running. Resuming from error now also clears `_errorMessage` and restores the flashing status.

2. **Pop-up blocked forced a full recompile.** When the browser blocked the flasher pop-up, the dialog went to a terminal error, so retrying meant recompiling/redownloading. It now stays on download-ready with the firmware still in hand and surfaces the pop-up-blocked message in the ready-screen detail, so the user can allow pop-ups and click Open again.

The third post-merge comment (a `#904` / `#907` PR-number deprecation marker on `installWebDownload`) is already resolved: #919 deleted that method entirely.

Added `firmware-install-dialog-usb-handoff.test.ts` covering both: pop-up-blocked parks on download-ready and keeps the firmware; an in-tab progress frame after an error clears the banner.

- Follow-up to #904

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/firmware-install-dialog/install-flow.ts
+++ b/src/components/firmware-install-dialog/install-flow.ts
@@ -459,16 +459,25 @@ export function handOffToFlasher(host: ESPHomeFirmwareInstallDialog): void {
   host._step = "flashing";
   host._flashPercent = 0;
   host._statusMessage = host._localize("firmware.usb_flashing");
+  host._errorMessage = "";
   const deviceName = host._device ? host._device.friendly_name || host._device.name : "";
+  // An in-tab retry after a failure resumes via progress/status frames; leave
+  // the error view and clear the failure banner so it doesn't headline a flash
+  // that's already running (a progress frame often lands before a status one).
+  const resumeFromError = () => {
+    if (host._step !== "error") return;
+    host._step = "flashing";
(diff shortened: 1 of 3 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
Thanks Kōan — suggestion #1 covered in `11a3ccb7`: added a renderer test asserting `cardStatusDetail` surfaces `_errorMessage` on the web-flash download-ready screen (the user-visible half of the pop-up-blocked fix) and falls back to `usb_built_body` when there is no error. Full suite green (3563).

esphbot (approved):
No blocking issues found.

PR #936, opened 2026-06-21
Drop header and footer for more breathing room

KTibow:

The header and footer take up a lot of space without adding anything right now. When you already know what page you're on based on your tab title, there's no point in telling users that ESPHome is where you "Create and make your smart devices and automations". This moves the back button and overflow menu (which now includes version info) into a more natural position on each page.
image
image
image

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [ ] `npm run test` passes. can't run for reasons explained in first pr


esphbot (changes requested):
Blocking issues found.

- Back-to-dashboard drops the dashboard's search/filter URL state
- Secrets back button also drops dashboard query state
- Version badge + release-link logic moved here but left untested

esphbot (changes requested):
Blocking issues found.

- Back button drops dashboard search/filter URL state
- Device board-info back drops dashboard URL state
- Relocated version badge/link logic left untested
- Absolute-positioned menu can be clipped by overflow:hidden ancestors

esphbot (changes requested):
Blocking issues found.

- Menu max-height uses vh only and can still be clipped by overflow:hidden ancestors

bdraco:
For the Home Assistant addon, I think this makes sense, but for the desktop app, I think we should keep the header

bdraco:
Before

Screenshot 2026-06-21 at 4 56 20 PM

Also need to think about mobile

esphbot (changes requested):
Blocking issues found.

- Dashboard loses its only top-level heading

bdraco:
This PR

Screenshot 2026-06-21 at 4 57 12 PM

carlhye:
Hi @KTibow 

Thank you for your contribution.
A change like this would need extensive UX work and testing to be implented, so we're not going to merge it for now.

Erioldoesdesign:
Just to add a UX expert comment here. There is a scenario where we reduce complexity of headings, headers and sections and also relocate essential buttons/functions to spaces that don't require an 'over-arching' header but as Carl said, this needs some usability testing for people that are gaining familiarity in ESPHome and aren't already building on existing knowledge from the previous GUI. 

he ideas are interesting and good though so we'll be sure to consider them in our work going forward 👍

PR #939, opened 2026-06-22
Filter already-set keys and complete nested mappings in the YAML editor

bdraco:

The YAML editor's autocomplete had regressed against the legacy dashboard in two ways:

1. It re-suggested keys that are already set in the surrounding mapping (fields already written under a block, or top-level blocks like `wifi:`/`api:` that already exist). The legacy editor filtered these out.
2. It offered nothing inside nested mappings; typing under `esp32:` then `framework:` returned no suggestions, where the legacy editor offered `advanced` and the rest of the framework fields.

Both behaviours are client-side; completion is computed in the frontend from the component catalog, so this is a frontend-only fix.

Changes:

- Add `collectSiblingKeys` (AST) and drop completions whose key is already present in the cursor's mapping, at plain key positions only. List-item positions (automation actions, filters) stay unfiltered since they are legitimately repeatable, and YAML map-key uniqueness is what makes the plain-key filter safe.
- Descend the catalog's nested `config_entries` along the cursor's AST key path (`getKeyPath`), so nested groups like `framework` and `framework: advanced:` complete. The catalog already models the nesting the visual form uses, so no schema or backend change is needed.

- Reported with editor autocomplete screenshots; no issue filed.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/util/yaml-ast.ts
+++ b/src/util/yaml-ast.ts
@@ -244,6 +244,28 @@ export function resolveBundleContext(
   return { topLevelKey, platformValue: getPlatformValue(state, pos) };
 }
 
+/** Iterate the direct ``Pair`` children of a ``BlockMapping`` node — the
+ *  shared primitive behind the mapping-key collectors below. */
+function* mappingPairs(map: SyntaxNode): Generator<SyntaxNode> {
+  for (let pair = map.firstChild; pair; pair = pair.nextSibling) {
+    if (pair.name === "Pair") yield pair;
+  }
+}
+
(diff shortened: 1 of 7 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #943, opened 2026-06-22
Show value completion at an empty trailing-space value

bdraco:

In the YAML editor, value completion appeared at `key:` but vanished as soon as you typed the space (`key: `). It was most visible on nested fields; e.g. under `esp32: framework:` typing `type: ` offered nothing, and `framework: advanced: minimum_chip_revision: ` showed no value options.

Root cause: a cursor in an empty trailing-space value resolves to the document root, so `getKeyPath` returned `[]` and the nested value descent could not find the entry. The fix re-anchors `getKeyPath` on the line's last non-space character (the `:` or key) when the cursor sits in empty trailing whitespace, so the key path (and the value options) still resolve.

This is a follow-up to #939.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/util/yaml-ast.ts
+++ b/src/util/yaml-ast.ts
@@ -182,7 +182,16 @@ export function getTopLevelKey(state: EditorState, pos: number): string | null {
  */
 export function getKeyPath(state: EditorState, pos: number): string[] {
   const keys: string[] = [];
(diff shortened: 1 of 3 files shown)


esphbot (approved):
No blocking issues found.

PR #950, opened 2026-06-22
Prefer the dotted platform impl over a same-named component

bdraco:

In the YAML editor, completing under `ota:` with `platform: esphome` offered the core `esphome:` component's fields (`name`, `area`, `areas`, `friendly_name`, `build_flags`, ...) instead of OTA's (`password`, `port`, `id`, ...).

Root cause: the platform merge in `resolveAvailableEntries` checked the bare platform value (`esphome`) before the dotted per-domain impl (`ota.esphome`). Both are catalog ids, so the bare core component won and its fields were merged in. The fix checks the dotted `<domain>.<platform>` id first; the bare lookup stays as the fallback. Every other platform (e.g. `sensor: - platform: dht` -> `sensor.dht`) is unaffected since the dotted id was already the match.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/util/yaml-completion-catalog.ts
+++ b/src/util/yaml-completion-catalog.ts
@@ -220,10 +220,13 @@ export async function resolveAvailableEntries(
   const componentEntries = async (componentId: string): Promise<ConfigEntry[]> => {
     let mergeId: string | null = null;
     if (platformValue) {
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

PR #951, opened 2026-06-22
Auto-offer completion at a fresh list-item dash

bdraco:

Typing a fresh list item (`- `) under a platform domain produced no suggestions: e.g. under `ota:` adding a second `  - ` offered nothing, so there was no `platform:` prompt until you typed a character. The key-position completion bailed on an empty partial for implicit (typing) completion.

Now a list-item dash auto-opens the popup, the same way `key:` already auto-fires value completion, so `platform:` (and the action registry inside automation bodies) appears as soon as you type `- `. A plain blank key line still waits for a partial or an explicit trigger, so this doesn't make ordinary key editing noisier.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/util/yaml-completion.ts
+++ b/src/util/yaml-completion.ts
@@ -277,7 +277,11 @@ export function createYamlCompletionSource(api: ESPHomeAPI) {
     const indent = leading.length;
     const keyFrom = pos - partial.length;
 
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

PR #962, opened 2026-06-23
Satisfy a catalog dependency a present component provides

### LINKED ISSUE #1664: [Bug] LibreTiny PWM dependency does conflict on bk72xx
### What happened?

On a generic bk72xx with the bk7231t board variant I tried to configure three PWM ports. TinyLibre PWM is supposed to be the correct one one, according to its description I guess:

Image

When I select it, it complains that a dependency is missing:

Image

So I added the dependency:

Image

Then I've added my first PWM port.

But when I want to save after this, it complains that TinyLibre is not supported under bk72xx, despite it's own description mentioning that it's the right one.

Image

### Dashboard version

1.0.12

### How did you install it?

Other / not sure

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

Adding the LibreTiny PWM output on a bk72xx board falsely reported a missing `libretiny` dependency. The dependency check matched dependency names against literal YAML keys only, so it never saw that the board's `bk72xx:` block already provides `libretiny`; following the banner's prompt then inserted an invalid `libretiny:` block that ESPHome rejected on save.

The check now treats a dependency as satisfied when a present top-level component provides it, resolved through the backend `provides` index, so on a bk72xx board the banner clears and Submit enables with nothing extra to add.

The catalog list filter in `component-catalog/filters.ts` has the same literal match blindness; it did not block this flow (the component stayed selectable), so it is tracked separately in #963.

- fixes https://github.com/esphome/device-builder/issues/1664

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/add-component-deps.ts
+++ b/src/components/device/add-component-deps.ts
@@ -1,4 +1,6 @@
+import type { ESPHomeAPI } from "../../api/index.js";
 import { ComponentCategory } from "../../api/types/components.js";
+import { providerIds } from "../../util/provides-cache.js";
 import {
   parseConfiguredPlatforms,
   parseTopLevelComponents,
@@ -43,4 +45,47 @@ export function findMissingDependencies(
     if (!PLATFORM_DOMAINS.has(dep) && platformStems.has(dep)) return false;
     return true;
   });
+}
+
+/**
+ * Of the deps `findMissingDependencies` still flags, the subset a present
+ * top-level component already *provides* under a different name.
+ *
+ * A `bk72xx:` block provides `libretiny` (the `output.libretiny_pwm` dep);
+ * a `tca9548a:` / `usb_uart:` block provides `i2c` / `uart`. The
+ * literal-name scan can't tie provider to dep, so each still-missing dep is
+ * matched against its providers from the `provides` index. Lookups are
+ * cached for the process lifetime (`providerIds`), so re-resolving on every
+ * YAML change costs one query per interface total; empty `missing`
+ * short-circuits with no round trip.
+ */
+export async function depsSatisfiedByProvides(
+  api: ESPHomeAPI,
+  missing: readonly string[],
+  present: ReadonlySet<string>,
+  ctx: { platform?: string | null; boardId?: string | null }
+): Promise<ReadonlySet<string>> {
+  const satisfied = new Set<string>();
+  // Dotted `<domain>.<platform>` deps are resolved by `findMissingDependencies`
+  // and never key the bare-id `provides` index, so a query for them always
+  // comes back empty — skip them rather than pay the round trip.
+  const resolvable = missing.filter((dep) => !dep.includes("."));
+  if (resolvable.length === 0) return satisfied;
+  await Promise.all(
+    resolvable.map(async (dep) => {
+      const providers = await providerIds(
+        api,
+        dep,
+        ctx.platform ?? undefined,
+        ctx.boardId ?? undefined
+      );
+      for (const id of providers) {
+        if (present.has(id)) {
+          satisfied.add(dep);
+          break;
+        }
+      }
+    })
+  );
+  return satisfied;
 }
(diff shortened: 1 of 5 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
Thanks. Addressed both in `5cb6906`:

1. **Stale `_providedDeps` at submit** — went with the fail-closed option: `_resolveProvidedDeps` now drops the prior set up front, before the async lookup, so during the in-flight window the gate reflects literal-missing (banner shown, Submit disabled) until the fresh result lands. The `_providesSeq` guard still prevents an older resolution from clobbering a newer one. Empty stays empty, so no extra re-render for the common no-provides case.

2. **`limit: 200`** — noted inline; one page holds every provider since an interface has at most a handful (same cap config-entry-form's provider fetch uses).

Also folded in a process-lifetime cache for the provider-id lookups (the catalog is immutable), so re-resolving on YAML changes costs one query per interface total instead of refiring `getComponents` each time.

bdraco:
Both Copilot points addressed in `065cc2d`:

1. **Dotted deps** — `depsSatisfiedByProvides` now filters out `<domain>.<platform>` deps before querying. Every `provides` token in the catalog is a bare interface id (`libretiny`, `uart`, `http_request`), so a dotted dep like `ota.http_request` would always come back empty; `findMissingDependencies` already handles those. Skipping them drops a guaranteed-empty round trip. Added a test that a dotted dep issues no `getComponents` call.

2. **Board in re-resolve triggers** — `willUpdate` now re-resolves on `board` change too, so a late or changed board platform/id can't leave `_providedDeps` scoped to the wrong board. The provider-id cache keeps the extra trigger free on repeat.

Full suite green (3833); re-verified live on a bk72xx device (libretiny banner stays gone) and the esp32 negative control (banner still shown).

esphbot (approved):
No blocking issues found.

bdraco:
Done in `a51117d`. Guarded the post-resolution assignment so the empty-over-empty case is skipped:

```ts
if (seq === this._providesSeq && (satisfied.size || this._providedDeps.size)) {
  this._providedDeps = satisfied;
}
```

On the not-provided path `_providedDeps` is already empty (the eager-clear is a no-op there), so this keeps the steady state render-free while preserving the fail-closed clear at the top.

Added a regression test that mounts the form on the not-provided path, wraps `render` to count cycles, and asserts the settled provides lookup produces no extra render and leaves the same Set reference; confirmed it fails when the guard is reverted.

bdraco:
Done in `275d6f3`. The provides-lookup `catch` now warns instead of swallowing, while still failing closed:

```ts
} catch (err) {
  console.warn("[add-component-form] provides lookup failed", err);
}
```

The up-front clear already leaves the deps flagged, so a failed lookup keeps the banner up and the warning makes it observable in the field; mirrors config-entry-form's provider fetch. Added a test that mounts with a rejecting `getComponents`, asserts the banner stays and Submit stays disabled, and that `console.warn` fired.

esphbot (changes requested):
Blocking issues found.

- Bare catch swallows all errors, diverging from the sibling provider-fetch convention

bdraco:
Pushed `1c315a4` (kept on top of the main merge rather than force-pushing over it).

**Copilot — `_resolveProvidedDeps`:**
- Stale in-flight write: `_providesSeq` is now bumped at the very top, so a re-entry that early-returns (dep-free component, or no api) still invalidates an older in-flight lookup; it can no longer pass the seq guard and write a stale `_providedDeps`. Added a regression test that retargets the form to a dep-free component while a lookup is in flight and asserts the late result is discarded (verified it fails with the old ordering).
- Dep-free parse: the resolver now short-circuits on `deps.length === 0` before `parseTopLevelComponents`, so the common dep-free case does no YAML parse.

**Copilot — test mock leakage:** added `vi.restoreAllMocks()` to the file's `afterEach` so the `console.warn` spy is restored even on a mid-test failure; dropped the now-redundant manual restore.

**Kōan — bare catch:** already addressed in `275d6f3` (now in this branch); the catch binds the error and `console.warn`s while still failing closed, matching config-entry-form's provider fetch.

Full suite green (3847).

esphbot (approved):
No blocking issues found.

PR #967, opened 2026-06-23
Submit non-destructive dialogs on Enter

### LINKED ISSUE #1670: [Bug] Some dialog boxes cannot be submitted by pressing ENTER
### What happened?

There are still some dialog boxes that cannot be submitted by pressing ENTER. I don't think there's much chance of finding them all by testing them myself - is there a way to identify them in the code?

Examples:

Image

Image

### Dashboard version

1.0.14

### How did you install it?

Desktop app

### Where is it running?

Local — Windows

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

Some dialogs could not be submitted with ENTER. This wires the existing `EnterController` (via `esphome-base-dialog`'s `confirmOnEnter`, or directly on a wizard step) to the dialogs that had a clear single primary action but no ENTER handling:

- Bulk-labels apply (gated on having pending changes)
- Pair build server, both the hostname input step and the confirm step
- The import-partial wizard step ("Open device")

Destructive or fleet-wide confirms intentionally stay click-only: overwrite-device, resolve-conflicts, and update-all (bulk firmware flash) are deliberately excluded so ENTER cannot trigger them by accident.

To answer the issue's "is there a way to identify them in the code?", this adds an enumerating audit test (`test/components/dialog-enter-submit.test.ts`); a new submit dialog that forgets its ENTER wiring fails that suite.

- fixes https://github.com/esphome/device-builder/issues/1670

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/labels/bulk-labels-dialog.ts
+++ b/src/components/labels/bulk-labels-dialog.ts
@@ -272,6 +272,12 @@ export class ESPHomeBulkLabelsDialog extends LitElement {
     return this._pendingChanges.size > 0;
   }
 
+  /** Apply is reachable (button enabled, Enter armed) only with staged
+   *  changes and no in-flight save. */
+  get _canApply(): boolean {
+    return this._hasPendingChanges && !this._saving;
+  }
+
   static styles = [
     espHomeStyles,
     labelChipStyles,
@@ -343,6 +349,7 @@ export class ESPHomeBulkLabelsDialog extends LitElement {
         .label=${this._localize("dashboard.labels_bulk_dialog_title", {
           count: targetCount,
         })}
+        .confirmOnEnter=${this._canApply ? this._apply : undefined}
         @after-hide=${this._onAfterHide}
       >
         ${this._catalog.length === 0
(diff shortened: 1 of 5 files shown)


esphbot (changes requested):
Blocking issues found.

- Held Enter can stage-skip input → confirm submit, bypassing pin verification

bdraco:
Addressed the held-Enter stage-skip (esphbot, blocking): pair-build-server now drives its own `EnterController` with an `e.repeat` guard instead of base-dialog's `confirmOnEnter`, because this dialog stays open across input→confirm. A held key no longer carries from the input step into the confirm submit; confirming requires a deliberate second press after the pin fingerprint is shown. Added a regression test, `pair-build-server ignores a held (repeat) Enter on the input step`.

On the Copilot note about the `@vitest-environment happy-dom` pragma: it sits in the top-of-file docblock, which vitest does honor. The repo's existing happy-dom tests use the same block-comment form (for example test/util/web-serial-availability.test.ts), and this suite imports Lit custom elements and passes, so it runs under happy-dom as intended. Left as is.

esphbot (approved):
No blocking issues found.

PR #980, opened 2026-06-25
Fix empty toolbar button after validate completes on mobile

bdraco:

On mobile, the Validate dialog footer showed an empty button frame once validation finished; the running Stop button gets recycled by Lit into the icon-less Close button, and Safari does not re-invalidate `.term-btn:has(wa-icon)` when the icon is removed, so the label stayed clipped on a now icon-less button. The collapse is now driven by a `term-btn--with-icon` class that updates reliably across the recycle, so the Close label reappears.

- fixes <link to issue>

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/logs-dialog.ts
+++ b/src/components/logs-dialog.ts
@@ -323,7 +323,7 @@ export class ESPHomeLogsDialog extends LitElement {
           ${this._backToInstall
             ? html`<button
                 slot="toolbar-left"
-                class="term-btn term-btn--ghost"
+                class="term-btn term-btn--ghost term-btn--with-icon"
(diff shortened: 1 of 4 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
Addressed suggestion #1 in 98b7b024; the back-to-install button now routes through `renderTermButton`, so it derives `term-btn--with-icon` from its icon and the collapse class is no longer hand-maintained. The expand-btn stays inline since it is icon-only with no label span.

esphbot (approved):
No blocking issues found.

PR #982, opened 2026-06-25
Disambiguate same-domain catalog cards by platform

bdraco:

The Stepper category in "Add component" shows two cards both titled "Stepper Component" with the same description, so it looks like a duplicate. It is actually the two real platforms, stepper.a4988 and stepper.uln2003, which share the domain's docs page and therefore the same name; the category chip can't tell them apart since the category is identical and it is hidden under a single category filter.

When a card shares a name with another visible card in the same category, this surfaces the platform stem from the id as a chip, so the two cards now read "A4988" and "ULN2003". Cross category same name pairs (sensor.debug, text_sensor.debug) are left to the existing category chip.

- reported in-app (screenshot); no tracking issue

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/component-card-category-label.ts
+++ b/src/components/device/component-card-category-label.ts
@@ -37,13 +37,38 @@ export function categoryChipLabel(category: string): string {
 }
 
 /**
+ * Upper-case the platform stem of a catalog id (the part after the
+ * domain in ``<domain>.<stem>``) so two entries that share a name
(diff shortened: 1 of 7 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #984, opened 2026-06-25
Skip the empty Add form for advanced-only components

bdraco:

Adding the "Socket" component opened an Add-component dialog with an empty body (just the YAML preview link, Back, Add). Socket's only config entry is `implementation`, marked `advanced: true`, and the add-component form has no "show advanced" toggle, so it painted nothing; the user was stranded on a blank form.

A prior fix (#604) already fast-paths a configless component (zero entries, e.g. Async TCP) by adding it directly with an empty payload and toasting. Socket fell in the gap: it has one entry, but it's advanced-only, so the count gate missed it.

The gate now skips the form whenever it would paint nothing, not just when the component has zero entries. It reads the form's own render decision via `addFormPaintsAnything`, built on the same `buildFormRenderPlan` the form's `render()` consumes (fields, exclusive-group dropdowns, constraint-cluster boxes, unsatisfied-constraint banners) under the add-form's fixed filter, seeded with the same `buildInitialValues` the form uses, so the emptiness check is in lockstep with what the user actually sees; an advanced-only component like `socket` and a component whose only fields are optional non-`name` (deferred to the section editor) both fast-path. The auto-add submits `coerceFields(entries, seeded)`, the same payload the form's Add would, so a seeded `id`/pin isn't dropped. Featured entries fast-path like any other component; because `buildInitialValues` sets `seedAll` for featured ids, their locked presets ride in that coerced payload rather than being dropped. The missing-dependency carve-out is preserved (`captive_portal` still shows its deps banner), and an active prefill/detour keeps the form too.

- Reported against the dashboard (adding `socket` shows an empty dialog); no tracking issue.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/add-component-dialog.ts
+++ b/src/components/device/add-component-dialog.ts
@@ -356,11 +356,15 @@ export class ESPHomeAddComponentDialog extends LitElement {
     //
     // The probe and the payload both use the values the form would seed
     // (`buildInitialValues`) coerced exactly as the form's Add does
-    // (`coerceFields`), so the fast-path is provably equivalent to opening
-    // the form and clicking Add: a required+advanced or board-pinned entry
-    // only renders once seeding makes it material (so `{}` would wrongly
-    // fast-path and the form would render it), and a seeded `id`/pin the
-    // form would submit unrendered isn't silently dropped.
+    // (`coerceFields`), so the fast-path matches opening the form and
+    // clicking Add: a required+advanced or board-pinned entry only renders
+    // once seeding makes it material (so `{}` would wrongly fast-path and
+    // the form would render it), and a seeded `id`/pin the form would submit
+    // unrendered isn't silently dropped. The one thing it skips is the
+    // form's `validateEntries` bail, so a contradictory schema (a required
+    // field that is also `advanced` with no default — unfillable in the
+    // form anyway) surfaces a backend-error toast here instead of the
+    // form's client-side block.
     const present = parseTopLevelComponents(this.yaml);
     const seeded = buildInitialValues({
       entries: result.entry.config_entries,
--- a/src/components/device/add-component-dialog.ts
+++ b/src/components/device/add-component-dialog.ts
@@ -336,32 +336,38 @@ export class ESPHomeAddComponentDialog extends LitElement {
     }
     this._selected = result.entry;
     this._submitError = "";
-    // Skip the empty form and add directly only when there is no
-    // non-advanced field to show. The add-form has no "show advanced"
-    // toggle, so an advanced-only component like `socket` (one
-    // `advanced: true` entry) would otherwise strand the user behind a
-    // blank dialog, same dead-end as a configless component. We probe
-    // with `requiredOnly: false` so a component with an *optional*
-    // non-advanced field still opens the form (only advanced/hidden/
-    // platform-gated entries are dropped). Carve-outs that keep the form:
-    // a missing top-level dependency (e.g. `captive_portal` needs `wifi`,
-    // surfaced by the deps banner) and featured entries (their locked
-    // presets must be submitted, so `{}` would fail backend validation).
-    // The empty schema makes the fields payload provably `{}`; on the rare
-    // API failure `_submitComponent` toasts the error (no form surface).
+    // Skip the empty form and add directly only when the form would
+    // paint nothing. We probe with the add-form's exact render filter
+    // (`requiredOnly: true`, no advanced toggle — see
+    // `add-component-form.ts:415-417`, and the lockstep probe in
+    // `_anyErrorIsVisible`), so the gate matches what the user would see:
+    // an advanced-only component like `socket` (one `advanced: true`
+    // entry) and a component whose only fields are optional non-`name`
+    // (deferred to the section editor) both render blank, so both
+    // fast-path instead of stranding the user. Carve-outs that keep the
+    // form: a missing top-level dependency (e.g. `captive_portal` needs
+    // `wifi`, surfaced by the deps banner) and featured entries (their
+    // locked presets must be submitted, so `{}` would fail backend
+    // validation). The empty payload is provably `{}`; on the rare API
+    // failure `_submitComponent` toasts the error (no form surface).
     const present = parseTopLevelComponents(this.yaml);
     const renderable = collectRenderablePaths(
       result.entry.config_entries,
       {},
       {
-        requiredOnly: false,
+        requiredOnly: true,
         showAdvanced: false,
         presentComponents: present,
         targetPlatform: this.board?.esphome.platform ?? null,
       }
     );
-    const isFeatured = result.entry.id.startsWith("featured.");
-    if (renderable.size === 0 && !isFeatured) {
+    // Featured entries carry locked presets in their `config_entries`;
+    // keep the form so those submit. A degenerate featured entry with no
+    // entries has no presets to lose, so let it fast-path like any other
+    // configless component.
+    const hasFeaturedPresets =
+      result.entry.id.startsWith("featured.") && result.entry.config_entries.length > 0;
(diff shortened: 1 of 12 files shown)


esphbot (changes requested):
Blocking issues found.

- Probe uses requiredOnly: false but the form renders required-only — optional-only components still hit a blank dialog

bdraco:
Addressed in 6d901b67.

**Blocking (requiredOnly mismatch):** switched the probe to \`requiredOnly: true\` so it matches what the add form actually paints (lockstep with \`_anyErrorIsVisible\`). An optional-only, non-\`name\` component now fast-paths too, since the required-only form would render it blank either way; the misleading comment is gone. Updated the existing "has options" test to use a required field (the old optional \`ssid\` fixture was encoding the blank-form behavior), and added two pins: an optional-only component fast-paths, and an always-shown \`name\` field keeps the form.

**Suggestion (featured + zero entries):** confirmed featured *components* (not just bundles) reach \`_onComponentSelected\` via \`_onAdd\` (renderers.ts:98,148), so the edge case is real. Tightened the carve-out to \`featured.* && config_entries.length > 0\` so a degenerate featured entry with no presets fast-paths like any configless component; added a test for it. Featured entries with presets still always open the form.

esphbot (changes requested):
Blocking issues found.

- Gate probes with `{}` but the form seeds `_values`, so a seeded advanced field can diverge

bdraco:
Closed the values-axis gap in 69f73857. The gate now probes \`addFormRenderablePaths(config_entries, buildInitialValues(...), board, present)\` instead of \`{}\`, so it seeds exactly what the form mounts. A required+advanced entry with a \`default_value\` (or a board-pinned advanced pin) becomes material after seeding and keeps the form open, instead of fast-pathing and dropping the value. Socket and the other fast-path cases are unaffected (their lone fields seed to nothing). Added a test: a required+advanced field with \`default_value: "AUTO"\` now opens the form rather than auto-adding.

bdraco:
Thanks, addressed the new Copilot comments in 01c9bca6.

**Payload divergence (the substantive one):** confirmed the form submits \`coerceFields(_entries, _values)\` with \`_values = buildInitialValues(...)\`, and \`coerceFields\` keeps any entry with a defined value (it does not filter to rendered fields), so a seeded \`id\`/pin/default is emitted even in required-only mode. The fast-path now submits \`coerceFields(config_entries, seeded)\` instead of \`{}\`, so it is provably equivalent to opening the form and clicking Add. Truly configless and advanced-only cases still resolve to \`{}\`; added a test for a multi_conf component with an id + advanced field that fast-paths and submits the seeded \`id\`.

(The earlier requiredOnly/probe-seeding points were already handled in 6d901b67 and 69f73857 — this commit closes the remaining submit-side gap.)

**featured-id docstring:** reworded so it matches the prefix-only check rather than implying full-shape validation.

esphbot (approved):
No blocking issues found.

bdraco:
Addressed the three Copilot comments:

1. **Stale PR description (requiredOnly: false):** updated the description; the implementation settled on `requiredOnly: true` via `addFormRenderablePaths` (matching the form's `?required-only` mount), so the body now describes that, the seeded probe, and the coerced-fields submit.

2. **Fast-path submitted `{}` vs the form's seeded payload:** fixed in 01c9bca6 — the fast-path now submits `coerceFields(config_entries, buildInitialValues(...))`, identical to the form's Add (a seeded `id`/pin is no longer dropped).

3. **Prefill divergence (prefillReference/depPrefill):** traced the flow — `_depPrefill` is only set on the direct-dep path that assigns `_selected` directly, and `_prefillReference` only in `_submitComponent`'s restore/bundle-advance which also assign `_selected` directly; both bypass `_onComponentSelected`, so the gate never runs with prefill set today. Added a guard anyway (d7842255) so a prefilled selection can never fast-path with stale `{}`-seeded inputs, plus a test that forces that branch.

Full suite green (3963).

esphbot (approved):
No blocking issues found.

bdraco:
Addressed the suggestion in 8dc8bcb2: added a note in the gate comment that the fast-path skips the form's `validateEntries` bail, so a contradictory required+advanced+no-default schema (unfillable in the form anyway) surfaces a backend-error toast here instead of the form's client-side block. Softened the "provably equivalent" wording to "matches" so it isn't read as covering the validation path. Comment-only; suite still green.

esphbot (approved):
No blocking issues found.

bdraco:
Closed the deps axis in 4307bb9d. The gate now computes missing deps with the form's own `findMissingDependencies(deps, yaml, present)` instead of a plain `!present.has(d)` check, so a dep satisfied by a dotted match or a platform stem (e.g. `sensor: platform: atm90e32` satisfying an `atm90e32` dep) no longer keeps a blank form. The async `provides` subtraction is intentionally left out (your call), so the gate stays strictly safe — it can only keep the form a touch more often, never wrongly auto-add. Added a test for the platform-stem case. Suite green (3964).

bdraco:
Addressed both Copilot comments in 7dced4db:

1. **ReadonlySet:** widened `presentComponents` to `ReadonlySet<string>` in `addFormRenderablePaths` and propagated it down the chain it reads through (`RenderFilterOptions`/`RenderFilterSource`, `isEntryVisible`, `validateEntries`). None of them mutate it (only `.has()`), and a `Set` still satisfies a `ReadonlySet` param, so no caller changes; now it matches `findMissingDependencies`.

2. **isFeaturedId doc:** reworded so it states only the `featured.` prefix is checked, not the full `featured.<board>.<local>` shape.

Suite green (3964).

esphbot (changes requested):
Blocking issues found.

bdraco:
Reworked the gate per the design discussion (d8f8dd5a) — this closes the render-axis gap properly instead of patching it.

Extracted the form's render decision into a shared pure helper `buildFormRenderPlan` (new `config-entry-form-plan.ts`): it computes exactly what `render()` paints — exclusive-group dropdowns, constraint-cluster boxes, and the filtered field set. `ESPHomeConfigEntryForm.render()` now consumes it instead of inlining that split, so there's one source of truth. The dialog gate calls `addFormPaintsAnything` (built on the same plan + the same `collectUnsatisfiedConstraints` the banners use), so "would the add-form be blank" can no longer drift from the actual paint — exclusive groups / clusters / required-group banners are all covered now, not just renderable leaves.

Added tests for an advanced-only exclusive group and an advanced-only constraint cluster (both keep the form; both would have wrongly fast-pathed before). Full suite green (3966), including the form's own render tests over the refactor.

esphbot (approved):
No blocking issues found.

bdraco:
Good catch — and it turned out the carve-out was obsolete, not just imprecise, so I removed it entirely (253ab86b). Once the fast-path payload became `coerceFields(buildInitialValues(...))`, and `buildInitialValues` already sets `seedAll` for featured ids, a featured entry's locked presets ride in the payload whether it fast-paths or opens the form. So the `isFeaturedId && length > 0` proxy protected nothing — it only kept a blank form open for a presetless featured entry (the dead-end you described). Featured entries now go through the same general path as everything else; replaced the "keeps the form" test with one asserting the preset is submitted on the fast-path (`fields: { implementation: "lwip" }`).

bdraco:
Updated the description to match: featured entries fast-path like any other component, and their locked presets ride in the coerced payload because `buildInitialValues` sets `seedAll` for featured ids (no longer "excluded"). Also fixed a stale `addFormRenderablePaths` reference in the same paragraph to `addFormPaintsAnything` / `buildFormRenderPlan`, which is what the gate actually reads now.

esphbot (approved):
No blocking issues found.

PR #986, opened 2026-06-25
Align a templatable value inside a map row

bdraco:

In a map value editor (for example an mDNS service `txt` record), a value that uses the Value / λ Lambda toggle rendered misaligned; the value input dropped below the toggle and offset to the right, and its redundant "Value" label re-appeared.

The map-row styles only compacted a plain `.field` value (suppress its label and description, collapse its gap). A templatable value nests its field one level deeper under `.templatable-field` (the toggle plus value column), so those rules never reached it. This extends the existing selectors to the templatable case so the value lines up like every other map row.

- n/a (reported from the mDNS service editor)

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/config-entry-form.styles.ts
+++ b/src/components/device/config-entry-form.styles.ts
@@ -391,14 +391,35 @@ export const configEntryFormStyles = css`
 
   /* Inside a map row the value's label and description are
      redundant (the map itself has those at the top) — suppress
-     them so each row stays compact. */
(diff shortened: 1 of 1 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #987, opened 2026-06-25
Skip the C++ overlay for an empty lambda value

bdraco:

The YAML editor crashed (`Invalid inner parse ranges given: [{from:N,to:N}]`) the moment a value was switched to a lambda via the Value / λ Lambda toggle. Flipping the toggle writes an empty `!lambda` (no code yet); the C++ overlay then tried to highlight a zero-length range, and parseMixed's checkRanges rejects an empty overlay range and throws, taking down the editor.

The mixed parser now computes the lambda's value span once and skips the C++ overlay when that span is empty (a just-toggled `!lambda`, an empty quoted `!lambda ''`, or an empty block `!lambda |-`); once real code is typed the overlay applies as before. Added tests that forcing a full parse over each empty-lambda shape no longer throws.

- n/a (reported from the mDNS service editor while toggling a txt value to lambda)

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/util/esphome-yaml-lang.ts
+++ b/src/util/esphome-yaml-lang.ts
@@ -16,7 +16,7 @@ import {
   foldNodeProp,
   indentService,
 } from "@codemirror/language";
-import type { Input, SyntaxNodeRef } from "@lezer/common";
+import type { Input, SyntaxNode, SyntaxNodeRef } from "@lezer/common";
 import { parseMixed } from "@lezer/common";
 import { parser as yamlParser } from "@lezer/yaml";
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
Addressed Suggestion 1: added a bare `!lambda` (no value node) test — `parse("x: !lambda\\n")` asserts no throw, pinning the lambdaSpan-returns-null path. 2c2767d

esphbot (approved):
No blocking issues found.

PR #988, opened 2026-06-25
Render a lambda map value inline instead of the edit-in-YAML placeholder

bdraco:

Switching a map value (for example an mDNS service txt record) to a lambda via the Value / λ Lambda toggle made the row fall back to "Complex value — edit in YAML", so the value could no longer be edited inline and the toggle disappeared (no way back to a literal from the visual form).

The map renderer flagged any non-primitive value as complex, which swept up lambda values; a lambda is not a structural complex value (list / dict), it is a templatable scalar the value template already edits inline in lambda mode. Treat a lambda value as not-complex so the row renders the value template (which shows the inline lambda editor and keeps the toggle) instead of the placeholder; genuine lists / dicts still get the edit-in-YAML placeholder.

- n/a (reported from the mDNS service editor while toggling a txt value to lambda)

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/config-entry-renderers/lists.ts
+++ b/src/components/device/config-entry-renderers/lists.ts
@@ -1,4 +1,5 @@
 import { html, nothing } from "lit";
+import { isLambdaValue } from "../../../api/types/automations.js";
 import type { ConfigEntry } from "../../../api/types/config-entries.js";
 import { ConfigEntryType } from "../../../api/types/config-entries.js";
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #989, opened 2026-06-26
Preserve in-progress form values across the dep detour

bdraco:

When you fill some fields in the Add-component form (an SPI device's `cs_pin`, line frequency, etc.) and then click "+ Add <bus>" to add a missing dependency, returning to the form wiped everything you had typed; the form re-mounts and re-seeds from scratch, carrying back only the new dependency's id reference.

This snapshots the form's in-progress values when a "+ Add <dep>" detour starts (in the dialog, before the form unmounts) and overlays them on return via a new `restoredValues` step in `buildInitialValues`. The overlay lands after the seeded defaults and before `prefillReference`, so what you typed survives while the just-added dependency's id still wins for the reference field. The snapshot is only applied to the original form on return, not to the dependency's own form during the detour, so the dependency keeps its own auto-generated id (no id collision).

This is the input-preservation half; the separate stale-pin-display issue (a pin select showing a value it never held after a detour) is fixed in esphome/device-builder-frontend#990.

- n/a (found in testing the bus-dependency add flow)

## Types of changes

## Frontend coordination

- [x] No frontend change needed

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/add-component-dialog.ts
+++ b/src/components/device/add-component-dialog.ts
@@ -452,6 +452,7 @@ export class ESPHomeAddComponentDialog extends LitElement {
     this._depDomain = null;
     this._prefillReference = null;
     this._depPrefill = null;
+    this._returnValues = null;
     this._bundleQueue = rest;
     this._bundleProgress = {
       current: 1,
@@ -470,7 +471,12 @@ export class ESPHomeAddComponentDialog extends LitElement {
     // sending them to the catalog and losing context.
     if (this._returnTo) {
       const restore = this._returnTo;
+      // Back out of the detour like a submit-return: keep the snapshot across
+      // the reset so the user's in-progress values survive on the original
+      // form (the binding reads `_returnValues` with `_returnTo` now null).
+      const snapshot = this._returnValues;
       this._resetDetourState();
+      this._returnValues = snapshot;
       this._selected = restore;
       this._submitError = "";
       return;
--- a/src/components/device/add-component-dialog.ts
+++ b/src/components/device/add-component-dialog.ts
@@ -216,12 +216,19 @@ export class ESPHomeAddComponentDialog extends LitElement {
   /** See ``navigateToDep`` for the seq-counter contract. */
   private _depNavSeq = 0;
 
-  private _resetDetourState() {
+  /** Null every in-flight dep-detour field. Shared with `_onBundleSelected`,
+   *  which abandons the detour but must NOT bump the selection seqs (its
+   *  hydrate already validated against the current token). */
+  private _clearDetourFields() {
     this._returnTo = null;
     this._depDomain = null;
     this._prefillReference = null;
     this._depPrefill = null;
     this._returnValues = null;
+  }
+
+  private _resetDetourState() {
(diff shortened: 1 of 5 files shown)


esphbot (changes requested):
Blocking issues found.

- `_onBundleSelected` abandons the detour but leaves `_returnValues` set, leaking stale values onto the bundle's first form
- Cancelling the detour with Back discards the in-progress values the PR set out to preserve

bdraco:
Both lifecycle gaps fixed in 49b80544.

1. **Bundle path:** added `this._returnValues = null;` to `_onBundleSelected`'s clear block, alongside the other detour-state resets, so picking a bundle mid-detour no longer overlays the prior component's values onto the bundle's first form.

2. **Back path:** made `_onBack`'s `_returnTo` branch preserve the snapshot across `_resetDetourState()` (capture, reset, restore), so backing out of a dep returns to the original form with your input intact, consistent with the submit-return path. I went with preserving rather than discarding since, as you noted, both Back and completing-the-dep "return to the form."

Added two tests pinning each: bundle-mid-detour clears the snapshot, and Back preserves it (restores `_selected` to the original with `_returnValues` intact). Suite green (3986).

esphbot (changes requested):
Blocking issues found.

- `_returnValues` leaks into the next bundle step after a mid-bundle dep detour

bdraco:
Ran /simplify and tightened the `_returnValues` lifecycle (ecda9e37):

- **DRY:** extracted `_clearDetourFields()` (the 5 detour fields, no seq bumps) shared by `_resetDetourState` and `_onBundleSelected`, removing the duplicated field list. The seq bumps stay in `_resetDetourState` only, since bumping them in the bundle path would invalidate the token its own hydrate already validated against.
- **Found + fixed a staleness gap:** the bundle-advance branch in `_submitComponent` left `_returnValues` set, so a bundle step that detoured would bleed its snapshot onto the next step. Now cleared there (the normal-add branch already resets; submit-return intentionally keeps it for the restore).
- **Coverage:** rewrote the test into a full lifecycle suite — capture (with/without a mounted form), the restore gate (mid-detour vs return vs fresh), and every exit (submit-return keeps, back keeps, bundle-pick clears, bundle-advance clears, reset clears).

The `_restoredValuesForMount` getter, the `_onBack` snapshot dance, `currentValues`, and the overlay ordering were all judged correctly minimal and kept. Suite green (3991).

esphbot (approved):
No blocking issues found.

bdraco:
Addressed both remaining nits in 16009a18:

1. **buildInitialValues doc:** added the restore step to the numbered list (step 4, between board-pin seeding and the prefill steps) and renumbered, so the comment matches the actual overlay order.
2. **currentValues:** now returns a shallow copy `{ ...this._values }` so the snapshot is stable even if the form keeps editing its values after the dialog captures it.

The earlier Copilot "_returnValues leaks into bundle-advance" note predates ecda9e37 — that path now explicitly clears the snapshot (with a test). Suite green (3991).

Ari4ka (approved):

PR #990, opened 2026-06-26
Clear a stale pin select when its value is unset

bdraco:

Adding a bus device (e.g. an `atm90e32` sensor), clicking "+ Add spi" to add the bus, then returning left the required CS Pin select showing a GPIO it never actually held; the Add button stayed greyed out because the value was never in form state. The same pin select DOM node is reused across the detour (sensor to spi to sensor), so the CS Pin kept displaying the SPI Clk Pin's GPIO that bled in from the detour.

The pin select opts out of the generic value sync (`data-no-value-sync`) and runs `_syncSelectedAttr`, which bailed when no option was `selected`; that left the stale `.value` in place. It now syncs to the selected option including the empty case, so a pin whose value is unset clears to empty. The unit picker (the other `data-no-value-sync` user) always has a default unit selected, so it never hits the empty branch.

- n/a (found in testing the bus-dependency add flow)

## Types of changes

## Frontend coordination

- [x] No frontend change needed

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/config-entry-form.ts
+++ b/src/components/device/config-entry-form.ts
@@ -516,10 +516,15 @@ export class ESPHomeConfigEntryForm extends LitElement {
       "wa-option[selected]"
     );
     const desired = selectedOption?.value ?? "";
-    if (!desired) return;
     const current = Array.isArray(select.value)
       ? (select.value[0] ?? "")
       : (select.value ?? "");
+    // Sync to the `?selected` option, INCLUDING the empty case: a pin select
+    // whose value is unset must clear, not keep a stale `.value`. The same
+    // pin-select DOM node is reused across a dep detour (atm90e32 -> spi ->
+    // atm90e32), so without this the CS Pin would keep showing the SPI Clk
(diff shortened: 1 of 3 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
Addressed both doc nits in 93ac4c8e:

1. Reworded the `config-entry-form.ts` comment generically: "the other `data-no-value-sync` selects always render a `?selected` option (the unit pickers keep the in-use unit; exclusive group selects a non-empty placeholder sentinel), so only the pin select reaches the empty branch" — no longer singles out the unit picker.
2. Updated the now-stale `exclusive-group.ts` `NO_SELECTION` comment: dropped the "(which no-ops on "")" clause this PR invalidated, kept the round-trip rationale.

Copilot's regression-test suggestion is already covered by `config-entry-form-pin-sync.test.ts` (clear-on-unset / sync-to-selected / no-op-when-correct).

esphbot (approved):
No blocking issues found.

PR #991, opened 2026-06-26
Recognize the renamed rp2 platform key in raw-YAML surfaces

### LINKED ISSUE #1695: rp2040 is being renamed rp2
https://github.com/esphome/esphome/pull/17145

Will need to handle both for back compat

bdraco:

ESPHome is renaming the `rp2040` target platform to `rp2` (esphome/esphome#17145), and the legacy `rp2040:` key stays as a deprecated alias, so a user's YAML may now use either `rp2:` or `rp2040:`. The backend folds `rp2` onto `rp2040` everywhere it ingests a platform string (esphome/device-builder#1704), so device-list driven UI keeps showing `rp2040`. The remaining gaps are the few surfaces that parse the user's raw `rp2:` YAML block client-side:

- `CORE_KEYS` (yaml-sections): a `rp2:` block now lands under Core in the editor navigator instead of Components.
- The navigator row icon map keeps the 32-bit MCU chip glyph for a `rp2:` section.

The catalog-driven wizard and component-validation paths need no change here: the board and component catalogs stay keyed on `rp2040` until the coordinated esphome bump, so both sides stay consistent.

- fixes esphome/device-builder#1695

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/navigator-row-icons.ts
+++ b/src/components/device/navigator-row-icons.ts
@@ -131,6 +131,8 @@ const DOMAIN_ICON: Record<string, readonly [string, string]> = {
   esp32: ["cpu-32-bit", mdiCpu32Bit],
   esp8266: ["cpu-32-bit", mdiCpu32Bit],
   rp2040: ["cpu-32-bit", mdiCpu32Bit],
+  // esphome#17145 renames the rp2040 platform key to rp2
+  rp2: ["cpu-32-bit", mdiCpu32Bit],
   bk72xx: ["cpu-32-bit", mdiCpu32Bit],
   rtl87xx: ["cpu-32-bit", mdiCpu32Bit],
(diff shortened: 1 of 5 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
Good catch; you are right that `wizard.tag.rp2` is unreachable today. `prettyDomain` just capitalizes the raw key for the navigator header, and the `wizard.tag.${tag}` lookups only run over board-catalog tags, which stay `rp2040` until the coordinated esphome bump. The label served nothing here, so I dropped it (fd0fa3a0) and will land it with the catalog flip. The two real surfaces, `CORE_KEYS` and `DOMAIN_ICON`, are unchanged.

PR #993, opened 2026-06-26
Collect multi_value nested lists so usb_uart channels submit

bdraco:

Adding the USB Host UART Interface (`usb_uart`) component through the Add component dialog failed with `invalid_args: Missing required field: Channels` even after filling in the Type and a channel row (ID, Baud Rate).

Channels is a repeatable list of mappings, modeled like `esphome.devices` / `esphome.areas` as a `NESTED` entry with `multi_value: true`. The renderer drew the list and validation passed it, but `coerceFields` only coerced the single map NESTED case; an array value fell back to `{}`, produced an empty result, and the field was never added to the WS payload, so the backend rejected the missing required field. This coerces each item of a `multi_value` NESTED entry and keeps the array, reusing the existing `asMappingList` helper so render, validate, and collect all share one list coercion. Empty lists and empty items still prune out, and inner scalars (baud_rate) coerce to numbers.

- fixes https://community.home-assistant.io/t/setting-up-component-usb-host-uart-interface-for-device-esp32-s2-devkitc-1-does-not-work/1015007

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/add-component-form-coerce.ts
+++ b/src/components/device/add-component-form-coerce.ts
@@ -1,7 +1,7 @@
 import type { ConfigEntry } from "../../api/types/config-entries.js";
 import { ConfigEntryType } from "../../api/types/config-entries.js";
 import { coerceIntFieldValue } from "../../util/int-input.js";
-import { asMappingList } from "../../util/nested-values.js";
+import { asMappingList, asRecord } from "../../util/nested-values.js";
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

PR #999, opened 2026-06-26
Detect bare-integer pin values in the used-pin scan

bdraco:

In the GPIO pin picker, a pin used elsewhere in the config as a bare integer (for example `uart: tx_pin: 1`) was not flagged as "Already used by ...", and no cross-section conflict fired; a config with `tx_pin: 1` and `stepper: dir_pin: GPIO1` showed GPIO1 as free in the stepper dropdown.

The used-pin map is built by a line scan that only matched the prefixed pin token forms (`GPIOn`, and the LibreTiny/nRF52 `P...` forms), since those prefixes are distinctive enough to scan unbounded. A bare integer carries no prefix, so it never entered the map. Pins written as `GPIOn` strings (what the builder emits) were already caught; bare ints come from hand-edited YAML. This parses the value of a pin-field key through the existing `parsePinGpio` helper so bare ints register alongside the prefixed forms. Pin-field keys are detected by the `_pin`/`_gpio` suffix convention (the same one in `board-pin-defaults.ts`, mirroring the backend's `_pin_feature_from_name`) plus the long-form `pin:` / `number:` sub-keys, so numeric non-pin values like `baud_rate: 9600`, `phy_addr: 0`, or `data_bits: 8` are not mistaken for pins.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/util/config-entry-yaml-scan.ts
+++ b/src/util/config-entry-yaml-scan.ts
@@ -17,7 +17,7 @@
  * type into a field) from O(N) per keystroke to O(1).
  */
 import type { ComponentCatalogEntry } from "../api/types/components.js";
-import { scanPinGpios } from "./pin-gpio.js";
+import { parsePinGpio, scanPinGpios } from "./pin-gpio.js";
 import { collectIdsAtPath, parseYamlTopLevelSections } from "./yaml-sections-core.js";
 
 /**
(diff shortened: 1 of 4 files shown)


esphbot (approved):
No blocking issues found.

PR #1002, opened 2026-06-26
Resolve substitutions in the id-reference picker labels

### LINKED ISSUE #1709: Substitutions not resolved in the id-reference / target picker
### The problem

In the Visual Editor, `${...}` substitutions are resolved and previewed on text fields but not on the id-reference / target picker.

A component `name:` text field shows its raw value `${device_friendly_name} Switch` plus a `{}` chip whose tooltip previews the resolved string ("WIFI Switch Switch"). But the id-reference / target picker (for example the **Target** of a `switch.turn_off:` action, or a `switch.output` reference) renders its option labels straight from the raw YAML `name:` scalar, so it shows the literal `${device_friendly_name} Relay` and never resolves the substitution. Both the open dropdown options and the closed select's selected value are affected.

### Expected

The picker should display the resolved name (e.g. "WIFI Switch Relay"), matching how text fields resolve substitutions.

### Example config

```yaml
substitutions:
  device_friendly_name: WIFI Switch

binary_sensor:
  - platform: gpio
    name: "${device_friendly_name} Switch"
    pin: P6
    on_press:
      then:
        - switch.turn_on: relay
    on_release:
      - switch.turn_off: relay
    filters:
      - delayed_on_off: 50ms

switch:
  - platform: output
    name: "${device_friendly_name} Relay"
    output: relayoutput
    id: relay
```

The `switch.turn_off: relay` action's Target picker shows `${device_friendly_name} Relay` instead of `WIFI Switch Relay`.

### Notes

This is a frontend bug; the id-reference picker renderer is a separate code path from the text-field renderer and does not pass the option label through the substitution resolver, even though the parsed substitution map is already on the render context. Fix will land in `esphome/device-builder-frontend`.

bdraco:

The id-reference / target picker (for example the Target of a `switch.turn_off:` action, or a `switch.output` reference) rendered its option labels straight from the raw YAML `name:` scalar, so a name like `${device_friendly_name} Relay` showed unresolved, even though text fields already resolve and preview substitutions. This resolves `${...}` against the file's `substitutions:` block for the option labels and the sole-candidate placeholder, reusing the existing `resolveSubstitutions` helper and the substitution map already on the render context.

The label is display-only; the stored value stays the component id, so selection and round-trip are unchanged. An unknown substitution (e.g. one defined in a `packages:` include the scan can't see) degrades to the literal text, matching the previous behaviour. Because the option `.label` drives the closed select's display too, the selected value resolves as well, not just the open dropdown.

- fixes esphome/device-builder#1709

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/config-entry-id-reference-renderer.ts
+++ b/src/components/device/config-entry-id-reference-renderer.ts
@@ -11,6 +11,7 @@ import {
   findReferenceCandidates,
   resolveSoleCandidate,
 } from "../../util/config-entry-yaml-scan.js";
+import { resolveSubstitutions } from "../../util/substitutions.js";
 import {
   effectiveDisabled,
   fieldKeyAttr,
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

PR #1004, opened 2026-06-27
Show the resolved substitution hint on the automation Target field

### LINKED ISSUE #1711: Automation editor Target field does not show the resolved substitution hint
### The problem

Follow-up to #1709 / device-builder-frontend#1002. The id-reference / target *picker* now resolves `${...}` substitutions, but the automation editor's read-only **Target** identity field at the top of the editor still shows the raw value.

For a `binary_sensor` named `${device_friendly_name} Switch`, opening one of its automations (e.g. On Press) shows the Target field as `${device_friendly_name} Switch (binary_sensor.gpio)` with no resolved preview, while the component's editable `name:` field shows the raw value plus the `{}` resolved-hint chip ("WIFI Switch Switch").

### Expected

The Target field should show the same `{}` resolved-hint chip underneath it (e.g. "WIFI Switch Switch (binary_sensor.gpio)"), matching the `name:` field.

### Notes

The read-only Target value is built in the automation editor (`automation-editor.ts` `_targetMetadataValue` / `_renderIdentityFields`) from the instance's raw `name:`. The substitution hint chip (`renderSubstitutionHint`) and resolver already exist and are used by the text-field renderer; the automation editor already has the device YAML, so it can parse and render the same hint. Frontend fix.

bdraco:

Follow-up to #1002. The id-reference picker now resolves `${...}`, but the automation editor's read-only **Target** identity field still showed the raw value (e.g. `${device_friendly_name} Switch (binary_sensor.gpio)`) with no preview. This renders the same `{}` resolved-hint chip under the Target field that the text fields already show, so it reads "WIFI Switch Switch (binary_sensor.gpio)".

`renderSubstitutionHint` is generalized to take `(value, substitutions, localize)` instead of the whole `RenderCtx` (it only used those two), so the automation editor (which has the device YAML and a localize, but no render context) can reuse it. The substitution-note chip styles are extracted into a small shared module included in the automation editor's shadow root; the config-entry form keeps its own copy (its selectors are shared with `.secret-note`).

- fixes esphome/device-builder#1711

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/automation-editor/automation-editor.styles.ts
+++ b/src/components/device/automation-editor/automation-editor.styles.ts
@@ -1,5 +1,6 @@
 import { css } from "lit";
 
+import { substitutionNoteStyles } from "../substitution-note.styles.js";
 import { automationEditorActionStyles } from "./automation-editor-actions.styles.js";
 import { automationEditorRowStyles } from "./automation-editor-rows.styles.js";
 import { automationEditorScriptParamStyles } from "./automation-editor-script-params.styles.js";
@@ -229,4 +230,5 @@ export const automationEditorStyles = [
   automationEditorScriptParamStyles,
   automationEditorRowStyles,
   automationEditorActionStyles,
+  substitutionNoteStyles,
 ];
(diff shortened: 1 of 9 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #1014, opened 2026-06-27
Show installed to target version on the Update button hover

bdraco:

Restores a feature from the legacy dashboard: hovering the Update button shows which ESPHome version a device is currently running and what an update would install, so you can decide whether to update without opening the drawer.

A shared `updateButtonTitle` helper builds the localized hover text "Update available: {installed} → {target}" (installed comes from the device's `deployed_version`, target from `current_version`); it falls back to the button's own label when either version is unknown. It is wired into all three Update buttons: the dashboard grid card, the dashboard table row, and the device editor footer. Both versions are already on the wire, so there is no backend change.

- Requested in https://github.com/orgs/esphome/discussions/3730

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


esphbot (approved):
No blocking issues found.

PR #1019, opened 2026-06-28
Tighten and balance the device-editor code-pane padding

### LINKED ISSUE #1018: Device editor: tighten and balance the YAML code-pane padding
### The problem

The device editor's YAML code pane drifted to too-heavy, unbalanced padding: the left inset (before the line numbers) is larger than the right, and the overall inset is heavier than intended, shrinking the usable text width. This was unintended drift, not a deliberate design — users notice the cramped text area.

### Cause

The pane was left at the 16px (`--wa-space-m`) inset the config form needs, but the code editor brings its own line-number gutter that sits inside the left padding and adds a 5px inner inset, so the left reads heavier than the bare-16px right. The top inset is also 16px while the bottom reserves a strip for the floating Install/Validate/Save row, so the vertical insets feel over-padded once the sides are trimmed. None of this was an intentional inset choice for the editor.

### Fix

Trim the code pane to an even, tighter inset using spacing tokens:
- Pane top + left + right → `--wa-space-xs` (8px), measured equal on both sides.
- Line-number gutter inner padding → 0, so the column sits flush with the 8px pane padding (numbers no longer read as extra left inset).
- Floating action row dropped to an `--wa-space-xs` bottom/right inset; the bottom reserve is recomputed so the buttons still sit *below* the content (never overlapping it).
- The gutter inset is a single shared token reused by the editor and the sticky-scroll overlay so they can't drift again.

Measured after: left 8px ≈ right 8px, line-number glyph ~9px from the edge, wider text area.

bdraco:

Fixes unintended padding drift in the device editor's YAML code-pane. The pane was left at a 16px (`--wa-space-m`) inset (the value the config form needs), but the code editor brings its own line-number gutter that sits inside the left padding and adds a 5px inner inset — so the left read heavier than the bare-16px right and the text area was cramped. This asymmetry/heaviness was incidental drift, not a deliberate inset choice.

Changes (all spacing tokens, DRY):
- Editor pane top + left + right → `--wa-space-xs` (8px); measured equal on both sides.
- Line-number gutter inner padding → 0, so the column sits flush with the 8px pane padding (numbers no longer read as extra left inset). The gutter inset is a single shared token (`EDITOR_GUTTER_PADDING_LEFT`) reused by the editor theme and the sticky-scroll overlay's first-paint fallback so they can't drift.
- Floating Install/Validate/Save row dropped to an 8px bottom/right inset; the bottom reserve is recomputed (`calc(2.25rem + var(--wa-space-xs) * 2)`) so the buttons still sit **below** the content and never overlap it (a UX users complained about previously).

Measured empirically after the change: left 8px ≈ right 8px, line-number glyph ~9px from the edge, wider text area; the sticky-scroll overlay stays aligned (it measures the gutter inset at runtime).

- fixes esphome/device-builder-frontend#1018

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/device-editor.styles.ts
+++ b/src/components/device/device-editor.styles.ts
@@ -93,8 +93,8 @@ export const deviceEditorStyles = css`
 
   .editor-floating-actions {
     position: absolute;
-    bottom: var(--wa-space-m);
-    right: var(--wa-space-m);
+    bottom: var(--wa-space-xs);
+    right: var(--wa-space-xs);
     z-index: 10;
     display: inline-flex;
     align-items: center;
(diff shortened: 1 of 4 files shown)


esphbot (approved):
No blocking issues found.

PR #1022, opened 2026-06-28
Offer a pre-checked full setup in the create wizard

bdraco:

Adds a pre-checked "Set up with all recommended components" option to the create wizard's "Name your device" step, shown only for a complete onboard board (`full_config`) the backend synthesized an `all_recommended` bundle for. On finish, the new device is created and then the board's recommended components are applied to it, so a pre-configured device lands ready to use instead of empty.

It uses the backend's synthesized `all_recommended` bundle order (from esphome/device-builder#1726). The option is only offered when that bundle exists, so a board whose recommended set is only reachable through a chained curated bundle isn't offered one-click setup (a blind apply would add components out of order). Each member is added via the existing `devices/add_component` path (best-effort, so a member the presets can't fill on their own is skipped and the user finishes it in the editor). The checkbox defaults on because a complete onboard device is almost always wanted whole.

Chained on #1017 (the one-shot bundle add); depends on the backend `full_config` attribute from esphome/device-builder#1726.

- fixes <link to issue>

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

bdraco:
Good catch. Reconciled the description, kept the code: dropped the "falls back to every featured component otherwise" claim (that fallback existed in an earlier revision but was removed in the simplify pass to avoid the out-of-order-add hazard). The body now states the option is only offered when the synthesized `all_recommended` bundle exists, matching `boardOffersFullSetup` / `fullSetupComponentIds` and the `boardOffersFullSetup` docstring. No code change.

bdraco:
All three addressed:

1. Description: dropped the unreachable 'falls back to every featured component' claim; the body now states the option is only offered when the synthesized all_recommended bundle exists (matches the gate + docstring). No code change.
2. Partial full setup is no longer console-only: ec6b845d toasts a warning naming how many recommended components couldn't be added automatically, so a partially-applied device isn't presented as complete. Added a test.
3. The wa-checkbox is now read via e.currentTarget with a structural `{ checked: boolean }` type instead of mis-casting e.target to HTMLInputElement.

esphbot (approved):
No blocking issues found.

PR #1026, opened 2026-06-29
Pre-fill a featured hub's pins when reached via the dependency detour

### LINKED ISSUE #1727: Imported LED-driver-hub boards don't pre-fill clock/data pins (BP5758D, SM2135, ...)
### The problem

On an imported board that drives an LED-driver hub (BP5758D, SM2135, MY9231, BP1658CJ), adding the driver from the component catalog leaves the hub's **Clock Pin** and **Data Pin** empty. The board knows the real GPIOs (they are in the upstream devices.esphome.io config), but they never reach the form, so the user has to guess the pins.

### Reproduce

1. Add the board **Arlec 9W 92mm Grid Connect Smart RGB+CCT LED Downlight (ALD295HA)** (bk72xx / CB2S).
2. Open the device, Add Component, pick the featured **BP5758D LED driver**.
3. The output step pre-fills (ID `output_red`, Channel `3`), then the wizard descends into the BP5758D hub dependency.
4. **Clock Pin** and **Data Pin** are empty dropdowns.

Expected: Clock Pin `GPIO26` and Data Pin `GPIO24` pre-filled and locked (the device's real wiring), exactly like the I/O-expander hubs (e.g. the `es32a08` sn74hc595 board) already pre-fill via the `requires` chain.

### Root cause

The devices.esphome.io importer (`script/sync_esphome_devices.py`) only lifts top-level hub blocks that are referenced by a pin preset (the I/O-expander path). An LED-driver hub is bound by the `output.<driver>` platform's catalog `dependencies` and owns its own `clock_pin` / `data_pin`, so its top-level block was dropped; no hub featured component, no `requires`, no pins. Separately, `_gpio_number` could not parse libretiny `P<n>` pin names (bk72xx `P26`), so even the pins it saw were skipped.

### Affected imported boards

`arlec_grid_connect_smart_led_9w_92mm_rgbcct_ald295ha`, `havells_glamax_9w_b22_wifi_led_rgb_cct_bulb` (BP5758D); `anwio_a19_e26_rgbww_8_5w_light_bulb` (SM2135); `ledvance_smart_wifi_b40_5w_rgbw_e14` (BP1658CJ); `zemismart_bulb_br30` (MY9231).

bdraco:

When a board has a featured hub component (e.g. a bp5758d / sm2135 LED driver with locked `clock_pin` / `data_pin`), the `requires` prerequisite flow pre-fills it only when the user picks the featured consumer directly. Reaching the same hub through the missing-dependency detour (the orange "Add …" banner) or a full-setup bundle (which doesn't re-resolve a queued member's `requires`) fetched the bare component and left the Clock/Data pins empty.

`navigateToDep` now looks up the board's featured component that materializes the dependency (matched by `component_id`) and seeds the dep form with its preset values, merged with any bus-constraint prefill. So the hub's pins arrive pre-filled however the user reaches it.

Pairs with the backend hub capture in esphome/device-builder#1728.

- fixes https://github.com/esphome/device-builder/issues/1727

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


esphbot (approved):
No blocking issues found.

bdraco:
Good call — fixed in 9d85c4fc (rebased onto the latest main).

`navigateToDep` now clones the fetched component with `locked: true` on the entries the featured entry locks (`applyFeaturedLocks`); `effectiveDisabled` reads `entry.locked`, so the hub's pins render **disabled** through the detour/bundle path, matching the direct featured pick — not just pre-filled-but-editable. The cached component isn't mutated (immutable clone). Added a test asserting `config_entries[clock_pin/data_pin].locked === true` after the detour.

esphbot (approved):
No blocking issues found.

bdraco:
Added the comment (commit on the branch). Worth noting it isn't reachable from imported boards: the importer's `_sole_hub_block` skips multi-hub configs (`test_ambiguous_multi_hub_is_skipped`), so it never emits two featured entries for the same hub `component_id` with diverging pins — only a hand-curated board could. Documented as an intentional limitation.

PR #1027, opened 2026-06-29
Keep the board grid columns equal width

bdraco:

In the "Select your board" wizard step, the 2-column board grid could render the first column wider than the second once boards with wide content scrolled in. The grid used `grid-template-columns: repeat(2, 1fr)`, and `1fr` is `minmax(auto, 1fr)`, so a card whose content has a large min-content width (a long unbreakable token in a name or an imported description) expands its whole column and steals width from the other; grid columns are shared across rows, so the whole column looked wider.

Apply the same defenses the component catalog grid already uses: `repeat(2, minmax(0, 1fr))` so columns ignore content min-width and stay equal, `min-width: 0` plus `overflow: hidden` on the card so it shrinks to its track, and `overflow-wrap: anywhere` on the title and description so long tokens wrap instead of forcing width.

- Reported via dashboard testing (ESP32-S3 board picker, first column wider than the second); no separate tracking issue.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/wizard/wizard-step-board.styles.ts
+++ b/src/components/wizard/wizard-step-board.styles.ts
@@ -141,7 +141,10 @@ export const wizardStepBoardStyles = css`
 
   .boards-grid {
     display: grid;
-    grid-template-columns: repeat(2, 1fr);
+    /* minmax(0, 1fr), not 1fr (= minmax(auto, 1fr)): a card with wide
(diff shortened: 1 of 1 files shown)


esphbot (approved):
No blocking issues found.

PR #1029, opened 2026-06-29
Fill featured reference fields from their preset id among many candidates

bdraco:

Adding a featured bundle through the editor's Add Component dialog could leave the last step's id-reference dropdowns empty. Concretely, the Arlec ALD295HA bundle adds 5 `bp5758d` outputs and then an `rgbww` light whose Blue/Green/Red/Cold White/Warm White fields each reference one of those outputs; the form opened with all 5 dropdowns blank, even though the outputs were already in the draft. At device-create time the same bundle fills correctly, so this was specific to the interactive form path.

Two causes, both in `add-component-form-seed.ts`:

1. `seedDefaults` resolved every reference field by sole-candidate, which returns nothing when more than one same-domain component exists (5 outputs share the `output` domain), so it discarded each field's authored preset id (`blue: output_blue`, etc.).
2. `buildInitialValues` then applied the chained bundle `prefillReference`, which carries only the last-added sibling (`output_warm`) and `findReferencePath` writes it into the first matching reference field, clobbering it.

The fix makes a featured preset authoritative for a reference field when the preset id names a component actually present in the live config, keeping sole-candidate as the fallback for a renamed single bus. It also skips a preset-pinned reference in `findReferencePath`, so the chained prefill fills only an unpinned reference (the dep-detour case) instead of overwriting a preset one. The form still opens for the member (the user reviews it) but now every dropdown is pre-filled.

- fixes <link to issue>

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/add-component-form-seed.ts
+++ b/src/components/device/add-component-form-seed.ts
@@ -54,6 +54,12 @@ export function findReferencePath(
       continue;
     }
     if (entry.references_component === domain) {
+      // Skip a field the featured preset already pinned (`from_preset` + a literal
+      // `default_value`): preset seeding fills it from the live candidates, so a
+      // chained bundle `_prefillReference` (the last sibling added) must not
+      // overwrite an earlier preset-pinned field. A non-preset ref (the dep
+      // detour) still matches, so its just-added dep id lands as before.
+      if (entry.from_preset && entry.default_value != null) continue;
       return [...prefix, entry.key];
     }
   }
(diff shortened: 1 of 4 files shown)


esphbot (changes requested):
Blocking issues found.

- findReferencePath skip is broader than seedDefaults' actual fill — a dep-detour ref can be stranded empty

bdraco:
Both addressed.

🟡 Important #1 (findReferencePath skip broader than the fill): fixed in a34b09a8. The skip now keys off what `seedDefaults` actually filled, not the `from_preset` flag. `buildInitialValues` snapshots the seeding result and threads it into `findReferencePath`, which skips a reference only when `getIn(seeded, path) !== undefined`. So a stale or ambiguous `from_preset` reference that seeding left empty is no longer skipped; the chained `prefillReference` still lands on it, closing the dep-detour stranding case you described. The snapshot is taken before the `restoredValues` overlay, so a restored value is still overridden by a legit prefill (the existing restore-vs-prefill test stays green). Added a `findReferencePath` case for the empty-seed path and an end-to-end `buildInitialValues` test (stale preset id + two candidates, prefill still fills the field).

Suggestion (seedReference now unused): already removed in 7cdc8032; the `seedDefaults` reference tests cover sole-candidate, ambiguity, no-match, and the packages-merge defer.

(Your review HEAD was 83855ad, before those two commits.)

esphbot (approved):
No blocking issues found.

PR #1030, opened 2026-06-29
Seed a multi_value list default element-wise

### LINKED ISSUE #1736: devices/add_component internal_error on a list-valued locked pin (octal SPI data_pins)
### Symptom

Running the Lanbon L9 "full setup", step 3 adds the **SPI Bus**. Clk Pin (GPIO21) and Data Pins (6,7,15,16,10,9,46,3) are correctly pre-filled and locked, but clicking **Add** fails with:

```
internal_error: Command failed: devices/add_component
```

`internal_error` (not a typed `CommandError`) means an unhandled exception escaped the handler.

### Likely cause

This is the first runtime path to receive a **list-valued locked pin** (octal SPI `data_pins`), emitted by the catalog as `{"value": [6,7,15,16,10,9,46,3], "locked": true}` (added in #1734).

`devices/add_component` (`controllers/devices/add_component.py`) calls `_apply_featured_presets` (`controllers/devices/helpers.py`) then `merge_component_yaml` (`helpers/yaml/component.py`). For a PIN-typed field `_apply_featured_presets` runs the value through `_normalize_pin_value`, and `merge_component_yaml` renders the pin; both were written for a scalar pin (bare GPIO / `{number: ...}` mapping), not a list. A list value most likely throws there.

### Repro

1. Open the Lanbon L9 device, run the full-setup wizard (or add the SPI Bus featured component directly).
2. Leave the locked Clk Pin / Data Pins as-is.
3. Click Add -> `internal_error: Command failed: devices/add_component`.

### Fix direction

Handle a list-valued PIN preset end to end in `_apply_featured_presets` (normalize/compare per element) and `merge_component_yaml` (emit the list as `data_pins:`), with a unit test for an octal SPI bus add. Related to the conflict-detector gap #1735 (same list-valued-pin root, different code path).

bdraco:

A featured component with a locked `multi_value` PIN field whose preset value is already a list (octal SPI `data_pins`, e.g. `[6, 7, 15, 16, 10, 9, 46, 3]`) rendered the whole list as a single comma-joined row ("6,7,15,16,10,9,46,3") instead of eight locked rows, and clicking Add submitted that lone string, which the backend rejected with `internal_error: Command failed: devices/add_component`.

`seedDefaults` (`add-component-form-seed.ts`) seeded a multi_value default as `[String(entry.default_value)]`, which for a list default collapses it to a one-element array of the joined string. This spreads an already-list default element-wise (preserving each value's type, so pin ints stay ints) and only wraps a scalar default into a one-element array. The multi-value field then renders one locked row per pin and Add submits the real 8-int list.

Surfaced by the new backend bus-lift (esphome/device-builder#1734), which is the first to emit list-valued locked pins.

- fixes https://github.com/esphome/device-builder/issues/1736

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/add-component-form-seed.ts
+++ b/src/components/device/add-component-form-seed.ts
@@ -136,8 +136,13 @@ export function seedDefaults(
       continue;
     }
     if (entry.default_value != null) {
+      // A multi_value default can already be a list (octal SPI `data_pins`,
+      // `[6, 7, 15, ...]`); spread it element-wise so each value seeds its own
+      // row with its original type. Only wrap a scalar default into a list.
       out[entry.key] = entry.multi_value
-        ? [String(entry.default_value)]
+        ? Array.isArray(entry.default_value)
+          ? [...entry.default_value]
+          : [String(entry.default_value)]
         : entry.default_value;
     } else if (entry.multi_value && entry.required) {
       out[entry.key] = [];
--- a/test/components/device/add-component-form-seed-fns.test.ts
+++ b/test/components/device/add-component-form-seed-fns.test.ts
@@ -230,6 +230,25 @@ describe("seedDefaults", () => {
     expect(seedDefaults(entries, "", localize)).toEqual({ tags: ["a"] });
   });
 
+  it("spreads a multi_value list default element-wise (octal SPI data_pins)", () => {
+    const entries: ConfigEntry[] = [
+      makeConfigEntry({
+        key: "data_pins",
+        type: ConfigEntryType.PIN,
+        required: true,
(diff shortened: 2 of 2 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
Thanks for the reviews. esphbot approved and Copilot's overview raised no actionable inline comments; no changes needed. The fix preserves list element types so octal `data_pins` seed as individual rows and submit as the int list.

PR #1034, opened 2026-06-29
Collapse advanced settings into a section below the form

bdraco:
> **Three alternative approaches to the same problem; we'll merge one and close the other two. My recommendation is #1035 (the switch): advanced fields are sometimes inline within an otherwise non-advanced section, so a "Show advanced settings" switch frames it better than a collapsible "Advanced settings" section, which implies grouping.**
>
> - #1020 — move the control into the section/editor header, above the form
> - #1034 — collapsible "Advanced settings (N)" disclosure, advanced fields below it
> - #1035 — keep the "Show advanced settings" switch at the boundary, advanced fields below it (recommended)

One of two alternate approaches to the same problem (the other is the sibling PR with a plain switch); we'll keep whichever the team prefers.

The "Show advanced settings" control rendered below the form, and advanced fields were filtered inline. That created a few problems:

- Turning advanced on grew the form and pushed the control (and the Delete button under it) down the page, often below the fold; a user who flipped it on couldn't easily find the control to turn it back off.
- The control jumped position every time it toggled.
- Some structured configs (e.g. `esp32:`) already have their own advanced section, so a control sitting inline in the config area looked like one of the structured fields; it wasn't obvious it was a UI control rather than part of the config.

This approach renders basic fields, then a collapsible "Advanced settings (N)" disclosure, then the advanced fields below it. Revealing advanced settings grows the form downward; the control stays put and reads clearly as a UI control, not a config field. The count tells you how much is behind it before expanding.

It lives inside `config-entry-form` (opt-in via an `advanced-section` attribute, so the add-component dialog, trigger picker, and condition tree are untouched). The form emits `advanced-toggle` and the hosts keep owning `showAdvanced`, so caret-follow auto-reveal, the script editor's Parameters block, and the all-advanced force-open (e.g. the `delay` action) all still work. Groups are classified advanced only when every member is; an all-advanced form renders with no control; a pre-filled advanced field opens the section so it isn't hidden.

## Screenshots

Screenshot 2026-06-29 at 7 59 18 AM

Screenshot 2026-06-29 at 7 59 09 AM

The downside to this option is sometimes all the advanced options are inside a non-advanced section form

Screenshot 2026-06-29 at 9 23 18 AM
Screenshot 2026-06-29 at 9 23 10 AM

- N/A; reported via dashboard screenshots

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


bdraco:
Alternate approach to the same problem; sibling PR #1035 keeps the switch and renders advanced fields below it. We'll merge whichever the team prefers.

bdraco:
Three alternative approaches to the same problem (the advanced-settings control got pushed below the fold when advanced fields were revealed, and read like a config field rather than a UI control). We'll merge whichever the team prefers and close the other two:

- #1020 — move the control up into the section/editor header, above the form.
- #1034 — a collapsible "Advanced settings (N)" disclosure at the basic/advanced boundary; advanced fields render below it.
- #1035 — keep the switch at the boundary; advanced fields render directly below it.

esphbot (changes requested):
Blocking issues found.

- Disclosure control is inert (dead affordance) when a pre-filled advanced field forces the section open

bdraco:
Addressed in 71f7e8e. When a pre-filled advanced field forces the section open, the disclosure now renders as a static "Advanced settings (N)" label with no chevron (no dead collapse affordance), while still showing the fields so it stays clear they are advanced. Also applied the shared fixes from the #1035 review: control visibility is now recursive (a nested advanced field in a mixed group keeps its reveal control), force-advanced-control survives the all-advanced case (script Parameters stays reachable), and the form syncs host showAdvanced on force-open.

esphbot (approved):
No blocking issues found.

bdraco:
Thanks. Applied suggestion #1 (aria-controls) in 66eb3a6: the disclosure button now carries `aria-controls="advanced-fields"` pointing at the revealed region, which is a `display: contents` wrapper so the advanced fields stay direct flex children (no layout/gap change). Skipped `role="status"` on the static locked label deliberately: it is static text read in document order (not skipped), and `role="status"` would make it an aria-live region that announces on change, which is the wrong semantic for a fixed label. The latent all-advanced chevron divergence is not reachable (a script keeps a non-advanced required `id`, so `basic` is never empty), so no change there.

bdraco:
Both a11y polish items are now in (53b3b22): the interactive disclosure has `aria-controls="advanced-fields"` pointing at a `display: contents` wrapper around the revealed region (no layout change), and the static locked label now carries `role="status"` so it announces as informational, matching the repo convention for non-control rows. Tests cover both.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

bdraco:
Updated recommendation: I now favor #1035 (the switch) over #1034 (the disclosure). Advanced fields are sometimes inline within an otherwise non-advanced section, so a "Show advanced settings" switch frames it better than a collapsible "Advanced settings" section, which implies grouping. Summaries updated.

PR #1036, opened 2026-06-30
Consume the auto-closed brace when accepting a substitution completion

### LINKED ISSUE #1761: Dashboard editor inserts extra } when accepting substitution autocomplete
### What happened?

1. Open the ESPHome Builder dashboard
2. Edit a device's configuration
3. Define a substitution at the top of the config, e.g.:
```
substitutions:
  name: "living-room-thermostat"
``` 
4. Elsewhere in the file, type `${`
5. The autocomplete dropdown appears showing the available substitutions variables
6. Select your substitution name and press enter or tab to accept
7. The variable appears like, `${name}}` instead of `${name}`

### Dashboard version

2026.6.3

### How did you install it?

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell

```

### Anything else?

Screenshot of what is seen.
Image

bdraco:

In the YAML config editor, typing `${` after declaring `substitutions:` opens a completion list of substitution names. Accepting one produced `${name}}` — a doubled closing brace — instead of `${name}`.

The completion itself was correct (its `apply` is `${name}`, one brace). The doubling came from CodeMirror's `closeBrackets` extension (pulled in via `basicSetup`): the instant the user types `{`, it auto-inserts a matching `}`, so the buffer already reads `${name}` with the cursor before the brace. The substitution `CompletionResult` was returned with `from` but no `to`, so CodeMirror defaulted `to` to the cursor and replaced only `${name`, leaving the auto-closed `}` behind.

Fix: extend the result's replaced range over a `}` sitting at the cursor (`to: pos + (next char is "}" ? 1 : 0)`), so accepting an option replaces `${name}` rather than inserting alongside the stray brace. When no brace follows (paste, or bracket-matching off) the completion still supplies its own single `}`. Scoped to the substitution completion — `closeBrackets` is unchanged for `[]`/`{}`/quotes elsewhere.

Added `test/util/yaml-completion-substitution.test.ts` pinning both paths (auto-closed brace consumed; lone completion still emits one brace).

- fixes https://github.com/esphome/device-builder/issues/1761

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/util/yaml-completion.ts
+++ b/src/util/yaml-completion.ts
@@ -147,8 +147,14 @@ export function createYamlCompletionSource(
       if (subRefMatch) {
         const subs = collectSubstitutionKeys(state);
         if (subs.length > 0) {
+          // closeBrackets (basicSetup) auto-inserts a matching ``}`` when
+          // the user types ``{``, so the buffer is already ``${<cursor>}``.
+          // Extend the replaced range over that brace so accepting an option
+          // doesn't leave a doubled ``}``.
+          const closeBrace = state.sliceDoc(pos, pos + 1) === "}" ? 1 : 0;
           return {
             from: valueFrom,
+            to: pos + closeBrace,
             options: subs.map((name) => ({
               label: `\${${name}}`,
               apply: `\${${name}}`,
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

PR #1037, opened 2026-06-30
Hide out-of-sync indicators until mDNS is the live source

### LINKED ISSUE #1758: Device Builder shows "Out of sync" after successful CLI OTA update
### What happened?

# Device Builder shows "Out of sync" after successful CLI OTA update

## Which version of ESPHome has the issue?

2026.6.3

## What type of installation are you using?

Docker container

## Which version of Home Assistant has the issue?

Not applicable (standalone ESPHome)

## What platform are you using?

ESP32

## Board

ESP32 (generic)

## Component causing the issue

Device Builder / Deployment Status

## Example YAML snippet

Not configuration-specific.

## Anything in the logs that might be useful?

The device reports the correct firmware version after the OTA update:

```text
INFO ESPHome 2026.6.3
INFO Starting log output from solarboard.local using esphome API

[08:35:26.534][I][app:151]: ESPHome version 2026.6.3 compiled on 2026-06-30 08:24:46 +0200
```

## What happened?

I updated an ESP32 via the ESPHome CLI:

```bash
esphome run solarboard.yaml
```

(or `esphome upload`).

The OTA update completes successfully, and the device reports the new firmware version (2026.6.3) over the native API.

However, the Device Builder UI still shows:

- **Current Version:** 2026.6.3
- **Deployed Version:** 2026.6.1
- **Status:** `Out of sync — flash to update`

So the firmware running on the device is correct, but the deployment status is not updated.

## What did you expect to happen?

After a successful CLI OTA upload, the Device Builder should either:

1. Update the stored deployment information so that **Deployed Version** matches the running firmware, or
2. Clearly indicate that **Deployed Version** only tracks installations initiated from the Device Builder UI.

Otherwise, the UI incorrectly suggests that the device still needs to be flashed.

## Additional information

This appears to happen when using the standalone ESPHome Docker container and performing OTA updates via the CLI instead of the Device Builder UI.

The device itself is running the expected firmware, so this seems to be a synchronization issue between CLI deployments and the Device Builder's deployment metadata rather than an OTA failure.

### Steps to reproduce

1. Run the standalone ESPHome Docker container.
2. Build and upload firmware via the CLI:

   ```bash
   esphome run <device>.yaml
   ```

3. Wait for the OTA update to complete successfully.
4. Open the Device Builder UI.
5. Observe that:
   - **Current Version** reflects the running firmware.
   - **Deployed Version** still shows the previously deployed version.
   - The device is marked as **Out of sync**.

### Notes

The issue appears to be limited to deployments initiated from the CLI. Flashing the same firmware through the Device Builder UI appears to synchronize the deployment status.

### Dashboard version

1.0.21

### How did you install it?

Other / not sure

### Where is it running?

Docker container

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

Hides the out-of-sync / update indicators until mDNS is the device's live source, so an mDNS-dark setup stops showing a false "Out of sync".

The deployed version and config hash come only from the device's mDNS broadcast. When mDNS is dark (a ping or MQTT only setup, e.g. a standalone Docker-bridge dashboard), those values go stale, so the dashboard flags a device as out of sync even though it is running the right firmware; an out-of-band CLI OTA is the reported trigger. We gracefully degrade instead of probing live devices.

Adds a single `mdnsOnline` helper (`src/util/device-sync.ts`) keyed on the new backend `active_source` field, and routes every modified / update indicator through it: the table dots and buttons, the card dots and buttons, the drawer badges, the drawer version and config-hash sections (which fall back to the existing "waiting for mDNS" empty state), the update-status facets and the update-all dialog, and the device editor install button.

The gate is precise, not blanket. `update_available` (device version vs server) is purely mDNS-sourced, so it always needs a live mDNS. `has_pending_changes` is gated only when it came from the config-hash compare (the backend's `pending_changes_via_hash`); a local mtime-driven YAML edit is trustworthy without mDNS, so it still cues install when mDNS is dark. A normal mDNS deployment is unchanged; an mDNS-dark or offline device hides only the stale mDNS-sourced indicators, and they reappear once mDNS hears from it again.

Needs the companion backend PR that adds `active_source` and `pending_changes_via_hash` to the device snapshot.

- fixes https://github.com/esphome/device-builder/issues/1758
- Companion backend PR: esphome/device-builder#1764 (adds `active_source`)

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/api/types/devices.ts
+++ b/src/api/types/devices.ts
@@ -98,6 +98,12 @@ export interface ConfiguredDevice {
   active_source?: ReachabilitySource;
   /** True until successfully compiled + deployed */
   has_pending_changes: boolean;
+  /** True when ``has_pending_changes`` came from the mDNS-sourced
+   *  config-hash compare (vs the local mtime fallback). The UI gates
+   *  only this case on a live mDNS, so a local YAML edit still cues
+   *  "install" when mDNS is dark. Optional / absent reads as a local
+   *  (non-mDNS-dependent) pending. */
+  pending_changes_via_hash?: boolean;
   /** True if compiled with older ESPHome version */
   update_available: boolean;
   /**
--- a/src/api/types/devices.ts
+++ b/src/api/types/devices.ts
@@ -3,6 +3,10 @@
  *
  * Part of the src/api/types.ts barrel split.
  */
+// Type-only: ``reachability.ts`` imports ``DeviceState`` from here, so a
+// value import would cycle. ``ReachabilitySource`` is a string-literal union,
+// erased at runtime, so the type-only edge is free.
+import type { ReachabilitySource } from "./reachability.js";
 
 // ─── Devices ─────────────────────────────────────────────────
 
@@ -84,6 +88,14 @@ export interface ConfiguredDevice {
    * mtime-based change detection.
    */
   deployed_config_hash: string;
+  /** Reachability channel currently driving the device's online state
+   *  (``mdns`` > ``mqtt`` > ``ping``). The deployed version / config
+   *  hash come only from the mDNS broadcast, so the out-of-sync /
+   *  update indicators are gated on ``active_source === "mdns"`` (see
+   *  ``src/util/device-sync.ts``): when mDNS is dark those values are
+   *  stale. Optional on the wire — absent / ``"unknown"`` means no
+   *  source has claimed the device yet, which reads as "not mDNS". */
+  active_source?: ReachabilitySource;
   /** True until successfully compiled + deployed */
   has_pending_changes: boolean;
   /** True if compiled with older ESPHome version */
--- a/src/components/dashboard/device-drawer-content.ts
+++ b/src/components/dashboard/device-drawer-content.ts
@@ -44,6 +44,7 @@ import {
   localizeContext,
 } from "../../context/index.js";
 import { espHomeStyles } from "../../styles/shared.js";
+import { devicePendingChanges, deviceUpdateAvailable } from "../../util/device-sync.js";
 import { getEncryptionState } from "../../util/encryption-state.js";
 import { registerMdiIcons } from "../../util/register-icons.js";
 import {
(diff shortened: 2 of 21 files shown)


esphbot (changes requested):
Blocking issues found.

- Gating `has_pending_changes` on mDNS may hide a legitimate local-edit "needs install" cue

bdraco:
Addressed the blocking concern via the backend split. The frontend now consumes `pending_changes_via_hash`: a hash-driven (mDNS-sourced) modified signal hides when mDNS is dark, but a local mtime-driven YAML edit still cues install. update_available stays gated since it is purely mDNS-sourced. Also fixed the stale facets test comment. Needs the companion backend PR esphome/device-builder#1764.

esphbot (changes requested):
Blocking issues found.

- Gated has_pending_changes leaks into the encryption lock indicator (card vs drawer divergence)

bdraco:
Addressed the encryption-lock divergence (Kōan #1). The card and table fed the mDNS-gated pending flag into the 4-state encryption lock; it now reads the raw has_pending_changes everywhere, so an mDNS-dark hash-pending encrypted device shows the same lock on the card/table as the drawer. Added a table name-cell regression test. While here, clarified the naming so has_* / hasPendingChanges always mean the raw device truth and the gated display decision is show_* (helpers showPendingChanges / showUpdateAvailable; card/row/editor props showModified / showUpdate). The stale facets test comment is fixed too.

esphbot (approved):
No blocking issues found.

bdraco:
Addressed the suggestion: added a card-level regression test (test/components/device-card.test.ts) mirroring the table one, asserting the encryption lock-clock shows (raw has_pending_changes) while the modified dot stays hidden (gated showModified) for an mDNS-dark hash-pending encrypted device.

esphbot (approved):
No blocking issues found.

Issue #1040, opened 2026-06-30
Components dialog should switch to all when there are no recommended


jesserockz:
When the user searches for a component with a device like starter kit, the dialog defaults to the recommended filter only which may have 0 results.

If a user searches, we should remove the category filtering altogether, or switch to all.

<img width="1814" height="1170" alt="Image" src="https://github.com/user-attachments/assets/78dbc5a9-14fe-46ca-9c63-2937e488dfc0" />

PR #1053, opened 2026-07-02
Keep featured matches when searching the Recommended category

bdraco:

Searching the "Recommended" (Featured) category in the Add component dialog returned "No components found" for terms that plainly match a featured card; for example, typing "onboard" on a starter kit board hid the "Onboard RGB LED" recommendation instead of showing it. The search input handler switched the category from Featured to All on every keystroke, so the query ran against the general catalog rather than the board's recommendations, which do not carry those names.

This keeps the category on search, so the fetch stays scoped to the board's recommendations (getComponents category=featured, query); the existing empty grid fallback in updated() drops to All only when no recommendation matches. That preserves the #1040 behavior, a term matching nothing recommended still falls through to the full catalog, without discarding recommendations that do match.

- Follow-up to #1040

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


esphbot (approved):
No blocking issues found.

PR #1056, opened 2026-07-02
Keep a pin field's current value selectable when it sits on a reserved pin

bdraco:

A reserved pin option (available: false) is disabled unless the board's locked_pins rescues it for the section being edited. When a field's current value lands on a reserved pin outside that set, from a hand edited YAML or a board manifest preset missing its lock, the selected option was disabled and the wa-select head rendered blank, hiding the real config value; this is how the Apollo ESK-1 accessory power switch showed an empty Pin field. Add the current value's GPIO to the rescue set so its option stays pickable and the value always renders, while every other reserved pin stays disabled. The manifest side of the ESK-1 case is fixed in esphome/device-builder#1800, which also adds a validator guard; this change covers the hand edited YAML case that no manifest fix can. Verified in the live dashboard with a config pointing at a reserved pin the board locks to a different section.

- companion backend PR: esphome/device-builder#1800

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/config-entry-pin-renderer.ts
+++ b/src/components/device/config-entry-pin-renderer.ts
@@ -42,9 +42,10 @@ interface PinOptionView {
   warn: boolean;
   /** Board-unavailable (occupied / tied to flash) — grouped under "Reserved". */
   reserved: boolean;
-  /** Non-selectable. A reserved pin is disabled *except* when it's locked to the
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

PR #1057, opened 2026-07-02
Reuse the full board body when re-entering the wizard setup step

### LINKED ISSUE #1798: [Bug] First-device wizard: going back from the Wi-Fi credentials step then forward again gets stuck, next step never shows
### What happened?

During first device creation, using the back button on the Wi-Fi credentials step breaks the wizard. Going forward again never loads the next step.

1. Fresh setup, create your first device
2. Click **Finish setup**, it advances to the Wi-Fi credentials section
3. Click back
4. Try to go forward again
5. The install is stuck, the next config flow never shows

Screenshot below.

Image

### Dashboard version

1.0.25 (ESPHome 2026.6.4)

### How did you install it?

Desktop app

### Where is it running?

Local — Windows

### Relevant logs / errors

None captured.

bdraco:

During first-device creation the setup wizard advances to the Wi-Fi credentials step, but pressing the header back arrow and re-picking the same board dropped that step and dead-ended the flow.

The Wi-Fi step only shows when the setup step reads `board.requires_wifi === true`, which is a full-body field derived only by `boards/get_board`; the picker emits slim `getBoards` entries whose `requires_wifi` hydrates to `false`. On first entry the dialog upgrades the slim entry to the full body, so the step shows. But it tracked that upgrade with a single-id flag, so after backing out to the picker and re-selecting the same board the upgrade was skipped and `_selectedBoard` was left as the slim entry; `requires_wifi` went `false` and the Wi-Fi step silently disappeared.

Cache fetched full bodies by id and reuse them on re-entry, so setup is never entered on a slim entry. A regression test drives back-then-re-pick and asserts the full body (with `requires_wifi`) is used, and that the cached body avoids a redundant refetch.

- fixes esphome/device-builder#1798

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/wizard/create-config-dialog.ts
+++ b/src/components/wizard/create-config-dialog.ts
@@ -80,9 +80,11 @@ export class ESPHomeCreateConfigDialog extends LitElement implements ImportFlowH
   @state()
   private _selectedBoard: BoardCatalogEntry | null = null;
 
-  /** Board id whose full body has already been fetched into `_selectedBoard`,
-   *  so re-entering the setup step doesn't refetch (getBoard is uncached). */
-  private _upgradedBoardId: string | null = null;
+  /** Full board bodies already fetched, keyed by id, so re-entering the setup
+   *  step reuses the full body (getBoard is uncached) rather than the slim
+   *  picker entry — whose `requires_wifi` hydrates to false and would drop the
+   *  Wi-Fi step on a back-then-re-pick of the same board. */
+  private _fullBoardById = new Map<string, BoardCatalogEntry>();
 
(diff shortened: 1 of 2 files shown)


esphbot (approved):
No blocking issues found.

bdraco:
Good catch on the cache-key asymmetry. Fixed in b76fb62a: the store now keys on the slim entry's `board.id` (matching the lookup and `openWithBoard`) instead of `full.id`, so an id the backend canonicalizes still hits the cache on re-pick.

esphbot (approved):
No blocking issues found.

PR #1060, opened 2026-07-03
Split BoardCatalogEntry into SlimBoard and the full body type

### LINKED ISSUE #1058: Split BoardCatalogEntry into distinct SlimBoard / FullBoard types
### Background

A `BoardCatalogEntry` carries both **slim** picker entries (from `boards/get_boards`) and **full** bodies (from `boards/get_board`). `requires_wifi` is derived only on the full body (backend `controllers/boards.py`); on a slim entry it is absent and `board-hydrate.ts` defaults it to `false`. So a slim `requires_wifi === false` does not mean "Wi-Fi not required", it means "unknown".

This ambiguity is a latent trap: it caused the create-wizard Wi-Fi step to silently disappear when setup was re-entered on a slim entry (esphome/device-builder#1798, fixed in #1057 by never entering setup on a slim body). The type system did nothing to prevent reading `requires_wifi` off a slim entry.

### Proposal

Split the single type into a discriminated `SlimBoard` / `FullBoard` union (a discriminator field like `full_config` already exists but is unused), so full-body-only fields such as `requires_wifi` exist only on `FullBoard`. Callers must then upgrade to a full body before reading them; misuse becomes a type error instead of a silent `false`.

### Notes

- Touches `src/api/types/boards.ts`, `src/util/board-hydrate.ts`, and every board consumer, so it is a standalone cleanup rather than part of a bugfix.
- Follow-up from #1057 review.

bdraco:

`boards/get_boards` and `boards/get_compatible_boards` return slim index entries that omit the body-only fields (`featured_components`, `featured_bundles`, `full_config`, `requires_wifi`); only `boards/get_board` derives them. The frontend used one `BoardCatalogEntry` type for both shapes, so reading `requires_wifi` off a slim entry compiled fine and silently hydrated to `false` — the latent trap behind the wizard Wi-Fi-step drop (device-builder#1798 / #1057).

This splits the type: `SlimBoard` is the base (the slim index fields), and `BoardCatalogEntry` extends it with the body-only fields. The slim producers and consumers (the board picker, the compatible-board lists, `device-board-info`, the change-board dialog) are typed on `SlimBoard`, so reading a body-only field off a slim entry is now a compile error rather than a silent `false`. `hydratePagedBoardsResponse` uses a new `hydrateSlimBoard` so slim entries no longer get body-field defaults attached; `hydrateBoard` reuses it for the full body. `boardImageUrl` is widened to `SlimBoard` (it only reads `images`).

The create wizard keeps a full-body `_selectedBoard` (only ever assigned in `_enterSetupStep` / `openWithBoard`) and tracks the in-flight pick with `_pickedBoardId`, instead of parking a slim entry in the full-typed slot to drive the stale-selection guard.

No user-visible behaviour change.

- fixes #1058

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/api/esphome-api.ts
+++ b/src/api/esphome-api.ts
@@ -24,7 +24,11 @@ import type {
   ParsedAutomation,
   YamlDiff,
 } from "./types/automations.js";
-import type { BoardCatalogEntry, PagedBoardsResponse } from "./types/boards.js";
+import type {
+  BoardCatalogEntry,
+  PagedBoardsResponse,
+  SlimBoard,
+} from "./types/boards.js";
 import type {
   ComponentCatalogEntry,
   PagedComponentsResponse,
@@ -1315,7 +1319,7 @@ export class ESPHomeAPI {
    * Boards interchangeable with this one (same PlatformIO target);
    * includes `boardId` itself.
    */
-  async getCompatibleBoards(boardId: string): Promise<BoardCatalogEntry[]> {
+  async getCompatibleBoards(boardId: string): Promise<SlimBoard[]> {
     const response = await this.sendCommand<PagedBoardsResponse>(
       "boards/get_compatible_boards",
       { board_id: boardId }
--- a/src/api/types/boards.ts
+++ b/src/api/types/boards.ts
@@ -107,7 +107,15 @@ export interface FeaturedBundle {
   image_url?: string;
 }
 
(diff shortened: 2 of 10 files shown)


esphbot (approved):
No blocking issues found.

PR #1062, opened 2026-07-03
Surface the real add-component error instead of a generic message

bdraco:

Adding a component — especially a board's recommended bundle — that failed showed a
generic **"Failed to add component"** banner that hid the actual cause, so a user
(and triage) had no idea what went wrong.

Two changes:

- The add / bundle failure catches now reuse the existing `formatApiError` helper, so
  an `APIError` surfaces its structured `details` (e.g. "GPIO 14 is already in use")
  instead of the wire-form `"<code>: <details>"` or the generic fallback.
- The body-load path had a second, sneakier source of the generic text:
  `fetchComponent` resolves but the backend omits the id (a stale or wrong backend) and
  returns `null` — not a thrown error — so there's no detail to surface, and the miss is
  cached for the session (a same-session retry won't recover). That case now gets its own
  honest message ("Couldn't load this component. Try reloading the page.") rather than
  the misleading "failed to add".

Surfaced while testing a new board's recommended bundles against a backend that didn't
have that board's components — exactly the backend-omitted-id (cached-null) case.

- n/a

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/add-component-dialog-selection.ts
+++ b/src/components/device/add-component-dialog-selection.ts
@@ -69,10 +69,11 @@ export async function hydrateForSelection(
     );
     if (seq !== host._selectionSeq) return { kind: "stale" };
     if (!entry) {
-      // A null body isn't a thrown error — ``fetchComponent`` swallows a
-      // transient miss (dropped WS, backend restart) and returns null, so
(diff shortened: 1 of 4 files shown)


esphbot (approved):
No blocking issues found.

PR #1069, opened 2026-07-05
Add bootloader update option to the install method dialog

bdraco:

Adds an "Update bootloader" option under Advanced options in the install method dialog, for ESP32 devices whose old bootloader logs "Bootloader too old for OTA rollback... Flash via USB once to update the bootloader"; with esphome 2026.6+ the bootloader can be updated over the network instead.

The option shows only when the backend reports `ota_partition_access` (the YAML sets `allow_partition_access: true` on the esphome OTA platform), the deployed config hash matches the expected hash (the running firmware actually has the OTA partition handler compiled in, per the docs prerequisite), and the device is online. Selecting it runs the normal install command flow with `bootloader: true`, which the backend turns into a compile plus `esphome upload --bootloader`. The option description carries the docs warning; a power loss or reset mid update can require a USB flash to recover.

Backend PR: esphome/device-builder#1837

- n/a (companion to esphome/device-builder#1837)

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #1071, opened 2026-07-05
Restore flash target from the chain's upload on followJob reattach

bdraco:

Selecting a bootloader install and then clicking "Build locally instead" resubmitted a plain firmware install. The public followJob restores port and bootloader from the job it attaches to, but an install chain is followed via its COMPILE head; the flash target lives on the dependent UPLOAD, so any reattach (device card busy badge, firmware tasks row) wiped `_bootloader` to false and the force local resubmit flashed the app.

followJob now resolves the compile head's dependent UPLOAD (same lookup shape as deriveFollowCommandType) and restores `_port` and `_bootloader` from it. This also fixes the same-shape address bug; reattaching to an install targeting an explicit IP no longer resets the resubmit target to "OTA".

- follow-up fix to #1069

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/command-dialog.ts
+++ b/src/components/command-dialog.ts
@@ -38,6 +38,7 @@ import { registerMdiIcons } from "../util/register-icons.js";
 import {
   deriveFollowCommandType,
   detachStream,
+  findDependentUpload,
   followJob,
   onForceLocalClick,
   resetRunState,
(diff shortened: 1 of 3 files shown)


esphbot (approved):
No blocking issues found.

PR #1073, opened 2026-07-05
Propagate ID renames to their references

bdraco:

Renaming a component's id through the visual editor now updates every reference to it in the same edit: value-position references, automation targets, substitution values holding the id, and id() calls in lambdas. Previously the rename silently broke the references and the user had to hunt them down from the resulting validation errors.

The declaring ID field also shows a hint with how many places reference the current id, so the rename doesn't feel risky. Guardrails keep half-renames impossible: no propagation when the old id survives as another section's declaration, when either side is a substitution, or across mid-edit invalid values; the pending rename waits for a valid one and chained renames stay consistent across flushes. Everything lands in one yaml-draft, so the YAML pane undoes the declaration and its references atomically.

Chained on #1072 (shares the id-reference renderer and the identifier helpers).

Verified live: renaming an ledc output id referenced by rtttl rewrites both in one draft with no validation errors appearing; the hint reads "Referenced in 1 place".

- None

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/config-entry-form.styles.ts
+++ b/src/components/device/config-entry-form.styles.ts
@@ -46,6 +46,13 @@ export const configEntryFormStyles = css`
     margin-top: var(--wa-space-2xs);
   }
 
+  /* Informational line under a field (the declaring-id reference count). */
+  .field-note {
+    color: var(--wa-color-text-quiet);
+    font-size: var(--wa-font-size-2xs);
+    margin-top: var(--wa-space-2xs);
+  }
+
   .field-description {
     font-size: var(--wa-font-size-2xs);
     color: var(--wa-color-text-quiet);
--- a/src/components/device/config-entry-form.ts
+++ b/src/components/device/config-entry-form.ts
@@ -86,6 +86,7 @@ import {
   renderExclusiveGroupField,
   renderFloatWithUnitField,
   renderIconField,
+  renderIdDeclarationField,
   renderIdReferenceField,
   renderMapField,
   renderMultiValueField,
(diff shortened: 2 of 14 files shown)


esphbot (changes requested):
Blocking issues found.

- `declarationLines` misses `globals` id declarations → reference-count hint over-counts by 1

bdraco:
The review ran against 636c299; commit 7d7d4dcc already covered all three scan findings: `declarationLines` collects globals `- id:` lines by direct scan (LIST_SECTIONS blocks are not expanded per item), inline flow sequences (`interlock: [relay1, relay2]`) are found and rewritten, and `scanSites`/`renameIdInValues` gate on `isValidEspHomeId` before any RegExp is built so a mid-typed value cannot throw inside the flush. Each has a test.

49cbd97c goes further on the false-positive tradeoff the summary called out: the free-text key denylist is gone. Whether a key's value can hold an id now comes from the containing section's cached catalog entry (`references_component`, the same schema the reference dropdowns resolve against). Non-declaring `id:` values, dotted action shorthands, substitutions values, and `id()` lambda calls stay universal; a section without a cached schema contributes no schema-keyed sites, which errs toward not touching lines it cannot classify.

On the silent-failure notes: the `entryAtPath` null case is unreachable in practice since `recordPendingIdRename` resolves the same entries the form just rendered the field from, and a failed resolution correctly declines to track a rename it cannot classify. The duplicate-survivor skip is intentional behavior rather than a broken promise: when another section still declares the old id, the references legitimately point at that survivor, so rewriting them would be the bug.

esphbot (approved):
No blocking issues found.

bdraco:
Closing; the inline unknown-id error in #1072 covers the practical need and the automatic rewrite adds more machinery than the workflow warrants.

PR #1081, opened 2026-07-05
Consolidate the dialog .actions footer style into a shared sibling

### LINKED ISSUE #1078: Consolidate the dialog .actions footer style into dialogActionButtonStyles
Surfaced by a /simplify (DRY) pass on #1076; out of scope for that PR.

The dialog footer `.actions` block is duplicated verbatim:

```css
.actions {
  display: flex;
  justify-content: flex-end;
  gap: var(--wa-space-s);
  padding: var(--wa-space-m) 0 var(--wa-space-l);
}
```

It appears identically in `update-all-dialog.ts` and now `desktop-update-dialog.ts` (#1076). `dialogActionButtonStyles` (`src/styles/dialog-action-buttons.ts`) deliberately covers only the `.btn*` classes, and its own doc note says to add a shared `.actions` variant "only when it ships with at least two consumers using identical chrome" — which is now the case.

Proposal: add the `.actions` wrapper to `dialogActionButtonStyles` (or a sibling shared style) and drop the local copy from both dialogs. Small; left out of #1076 to avoid touching the shared style file and update-all-dialog. Note other dialogs use a loosely similar flex-end footer with slightly varying padding, so scope carefully.

bdraco:

Seven dialogs carried a byte-identical `.actions` footer block (right-aligned button row with the standard dialog-body padding). This adds a `dialogActionsRowStyles` sibling export next to `dialogActionButtonStyles` and drops the local copy from all seven; `update-all-dialog` and `desktop-update-dialog` (which already import the button styles) plus `clone-device`, `adopt`, `friendly-name`, `rename-device`, and `change-board`.

It is a deliberate sibling rather than folding `.actions` into `dialogActionButtonStyles`, because the five self-styling dialogs define their own `.btn` chrome; importing the full button styles would drag in a second, slightly divergent `.btn--cancel:hover` rule. The sibling lets them share the `.actions` row while keeping their button styles local. Those `.btn` copies are still duplicated among themselves; folding that in would change the hover rule, so it is left for a separate follow-up. No behaviour change.

- fixes https://github.com/esphome/device-builder-frontend/issues/1078

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/adopt-dialog.ts
+++ b/src/components/adopt-dialog.ts
@@ -5,6 +5,7 @@ import type { ESPHomeAPI } from "../api/esphome-api.js";
 import type { AdoptableDevice } from "../api/types/devices.js";
 import type { LocalizeFunc } from "../common/localize.js";
 import { apiContext, localizeContext } from "../context/index.js";
+import { dialogActionsRowStyles } from "../styles/dialog-action-buttons.js";
 import { dialogChromeStyles } from "../styles/dialog-chrome.js";
 import { inputStyles } from "../styles/inputs.js";
 import { espHomeStyles } from "../styles/shared.js";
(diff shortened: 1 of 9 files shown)


esphbot (approved):
No blocking issues found.

PR #1117, opened 2026-07-06
Add ethernet clk_mode migration banner to the component editor

bdraco:

Adds a one-click migration banner to the ethernet component panel, matching the shape of the "Enable encryption" security nudge. When the config carries the deprecated flat `clk_mode: GPIO<n>_(IN|OUT)` (removed upstream in ESPHome 2026.9.0), a notice appears above the form with a Migrate button that rewrites the draft to the nested `clk: {pin, mode}` form, dropping `clk_mode`. The rewrite is derived from the `GPIO<n>_(IN|OUT)` naming rule so any pin folds correctly; unknown or non-string values hide the banner rather than guessing.

The new `<esphome-deprecation-notice>` is registry driven like `SECURITY_SETTINGS`, so the next deprecated option is one registry entry plus copy. Shared plumbing moved out of the security notice into neutral modules: `notice-banner.ts` (the `apply-section-values` event both banners emit, with `undefined` meaning remove the key) and `notice-banner.styles.ts` (the banner CSS). The host handler was renamed `applySectionValues` accordingly; no dialog and no backend call since the rewrite is lossless and draft-only until save.

Verified in the live editor against a `clk_mode: GPIO17_OUT` config: banner shows, Migrate rewrites the buffer (siblings untouched), the clk form fills with pin GPIO17 and mode CLK_OUT, the banner self-hides, and the saved YAML passes `esphome config` with no deprecation warning.

- part of esphome/backlog#152 (backend ingest normalization is esphome/device-builder#1868)

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


esphbot (approved):
No blocking issues found.

PR #1130, opened 2026-07-07
Extract shared field styles for the text-input dialogs

### LINKED ISSUE #1125: Extract shared field styles for the text-input dialogs
The single text input dialogs each re-declare the same field stack CSS inline: `.field`, the field `label`, `.field-error`, `.field-warning`, and the helper text rule. The blocks in `rename-device-dialog.ts`, `clone-device-dialog.ts`, and `friendly-name-dialog.ts` are near verbatim copies (rename and clone differ only in the dialog width), with partial copies in `edit-pairing-endpoint-dialog.ts` and `adopt-dialog.ts`.

Proposal: add a shared `dialogFieldStyles` export under `src/styles/` and adopt it in the three dialogs, keeping per dialog width and body padding local. `adopt-dialog.ts` has a documented cascade ordering constraint, so it only joins if that comment survives cleanly.

Note there is already a second, parallel field styling convention in `src/styles/form-fields.ts` (`.field-label` / `.error`) used by the add-automation family; unifying the two conventions is worthwhile but out of scope here.

Found during a duplication audit of the codebase.

bdraco:

The rename-device, clone-device, and friendly-name dialogs each re-declared the same field stack CSS inline (`.field`, the field `label`, `.helper`, `.field-error`, `.field-warning`). This moves those rules into a shared `dialogFieldStyles` export (`src/styles/dialog-fields.ts`) and adopts it in the three dialogs; per dialog width and body padding stay local.

Verified no visual change by comparing computed styles in the running dashboard before and after, including the error and underscore warning states. The `edit-pairing-endpoint` and `adopt` dialogs keep their own field CSS on purpose, their spacing values differ so adopting the shared block would be a visual change. The parallel `formFieldStyles` convention used by the add-automation family is out of scope, see the issue for context.

- fixes https://github.com/esphome/device-builder-frontend/issues/1125

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/clone-device-dialog.ts
+++ b/src/components/clone-device-dialog.ts
@@ -8,6 +8,7 @@ import {
   dialogActionsRowStyles,
 } from "../styles/dialog-action-buttons.js";
 import { dialogChromeStyles, quietCloseButtonStyles } from "../styles/dialog-chrome.js";
+import { dialogFieldStyles } from "../styles/dialog-fields.js";
 import { inputStyles } from "../styles/inputs.js";
 import { espHomeStyles } from "../styles/shared.js";
 import { getDeviceNameWarning, validateDeviceName } from "../util/config-validation.js";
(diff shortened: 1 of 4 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1131, opened 2026-07-07
Route imperative-open dialogs through base-dialog's confirmOnEnter

### LINKED ISSUE #1126: Route imperative-open dialogs through base-dialog's confirmOnEnter
`esphome-base-dialog` grew a `confirmOnEnter` property precisely so form dialogs stop re-wiring their own Enter handling, and the add-api-action, add-automation, add-script, bulk-labels, pair-build-server, and secrets-structured-editor dialogs already use it. Six older dialogs never migrated and still hand-roll the full scaffold (a private `EnterController`, mostly a `_resolved` one shot latch, and an `after-hide` teardown):

- `rename-device-dialog.ts`
- `clone-device-dialog.ts`
- `friendly-name-dialog.ts`
- `confirm-dialog.ts`
- `unsaved-changes-dialog.ts`
- `yaml-validation-dialog.ts`

That is roughly 15 to 25 lines of behavioral boilerplate per dialog, and the comments in friendly-name and rename already call out the inconsistency between the siblings.

Proposal: bind `.confirmOnEnter` on the base dialog in these six and delete the local EnterController wiring. The base detaches Enter the instant `open` flips false rather than at `wa-after-hide`, so each dialog needs a small audit of whether its `_resolved` latch is still required (the base's contract says the callback must self-guard regardless); the stale comments should be updated as part of the change. None of the six should open a nested dialog while staying open, which is the documented `confirmOnEnter` caveat; that needs a quick check per dialog.

Wizard steps and onboarding also use `EnterController` directly but are not base-dialog consumers, so they stay as they are.

Found during a duplication audit of the codebase.

bdraco:

`esphome-base-dialog` grew `confirmOnEnter` so form dialogs stop re-wiring their own Enter handling, and the add-automation family already uses it. Six older dialogs never migrated and still hand-rolled the full scaffold (a private `EnterController`, a one shot latch, and an `after-hide` teardown): rename-device, clone-device, friendly-name, confirm, unsaved-changes, and yaml-validation. This routes all six through `.confirmOnEnter` and deletes the local wiring.

Behavior notes: the base detaches Enter the instant `open` flips false rather than at `wa-after-hide`, so the stale comments describing the old teardown are updated; the `_resolved` style latches stay because the detach is still asynchronous and buttons remain clickable through the hide animation. The confirm and yaml-validation callbacks stay bound even when they no-op (destructive prompt, unusable go-to-error) so the dialog keeps claiming the Enter keydown. Tests that pinned the old detach-at-after-hide contract are updated, and the friendly-name suite now mocks `wa-dialog` instead of `base-dialog` so the real Enter wiring is exercised.

Verified in the running dashboard: Enter confirms the duplicate dialog exactly once on a double press, an unchanged rename is a no-op, and Escape still dismisses.

- fixes https://github.com/esphome/device-builder-frontend/issues/1126

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1141, opened 2026-07-08
Use official ESPHome monochrome logomarks in header and nav

elcaptain:
Swaps the dashboard logos for the official ESPHome brand marks. The header now uses the monochrome white logomark and the device navigator rows use the monochrome black logomark. Adds the canonical color logomark as a reference asset and drops the old off brand esphome.svg.

The header logo was a non brand color variant that no other OHF project ships. This replaces it with the approved monochrome marks: white on the primary header, black in the nav rows. It renames esphome-mono.svg to esphome-mono-black.svg, adds esphome-mono-white.svg for the header, and keeps esphome-color.svg as the canonical color reference. Call sites and tests are updated to the new filenames.

N/A

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


bdraco:
Without the blue color we might end up with the wrong color in the icon if the header color changes

esphbot (changes requested):
Blocking issues found.

- Header logo swapped from full-color mark to monochrome — duplicates existing esphome-mono.svg and needs a design decision

elcaptain:
@bdraco the thing is the previous colored version was wrong. there is only [one colored logomark,](https://github.com/OpenHomeFoundation/brand-assets/tree/main/esphome/logo/screen/logomark) which is white E on blue house (yes, we have not published brand guidelines yet, but it's in the pipeline ;) ). If the header changes to a bright color we'd need to change to the black monochrome version.

Added a new commit with new logomark variants.

bdraco:
What I'm concerned about is the inner transparency uses whatever color is behind it for the circuit and thats not going to be an ESPHome color all the time.

elcaptain:
@bdraco but that is totally fine. the monochrome logomark version is supposed to work like that. my concern is brand consistency and that the currently used variant isn't "approved" by the ohf marketing team and no other ohf project has such a variant.

Erioldoesdesign (approved):
LGTM

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Header mono-white logo renders its "E" as a background-color cutout — unresolved maintainer concern

bdraco:
Thanks for the review. Addressed the cleanup:

**Suggestion #1 (Figma cruft in `esphome-mono-white.svg`)** — dropped the three nested full-viewBox `clip-path` `<g>` wrappers and their `<defs>` (they clip nothing). I deliberately **kept** the path's `fill="#F2F4F9"` plus the `color(display-p3 …)` style rather than reducing to a bare sRGB fill: removing the P3 override measurably changes the rendered white. I verified this empirically in headless Chrome (draw both SVGs to canvas, compare `getImageData`):

- full strip (drop display-p3): up to **3/255** delta at 40px and **7/255** at 800px, across essentially the whole house fill;
- clips-only removal (what shipped): **0** differing pixels at both 40px and 800px — pixel-identical to the approved original.

So the mark renders exactly as before; only the inert wrappers are gone.

**Suggestion #2 (`esphome-color.svg` unused)** — intentional. It's kept as the canonical color logomark reference asset for the repo; leaving it in on purpose rather than wiring it into a call site.

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco (approved):

bdraco:
Thanks @elcaptain @Erioldoesdesign

PR #1142, opened 2026-07-08
Persist table columns and rows per page across card/list view toggle

### LINKED ISSUE #1899: [Bug] Rows per page/Column settings not stored
### What happened?

Change to List View.
Change visible Columns and/or Rows per page number.
Change to Card View.
Change to List View. 
:( Columns and Rows per page again set to defaults.

### Dashboard version

ESPHome Device Builder v1.0.23 ESPHome 2026.6.4

### How did you install it?

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

Switching from List view to Card view destroys the table element, and switching back re-seeded it from host state that was only ever written at page load; any in-session change to visible columns, rows per page, or sorting was lost. `saveTablePreference` now mirrors each change onto the host fields that seed a remounted table, so the toggle keeps the user's settings; the table also ignores the echo of its own page-size dispatch so it cannot reset the current page.

- fixes https://github.com/esphome/device-builder/issues/1899

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/dashboard/device-table.ts
+++ b/src/components/dashboard/device-table.ts
@@ -128,15 +128,15 @@ export class ESPHomeDeviceTable extends LitElement {
   @property({ attribute: false })
   recentJobs = new Map<string, FirmwareJob>();
 
-  /** Initial sorting from preferences — applied once when first set. */
+  /** Sorting from preferences — the host mirrors saves back, so a remount reseeds it. */
   @property({ attribute: false })
   initialSorting: SortingState | null = null;
 
-  /** Initial column visibility from preferences — applied once when first set. */
+  /** Column visibility from preferences — the host mirrors saves back, so a remount reseeds it. */
   @property({ attribute: false })
   initialColumnVisibility: VisibilityState | null = null;
 
-  /** Initial page size from preferences — applied once when first set. */
+  /** Page size from preferences — the host mirrors saves back, so a remount reseeds it. */
   @property({ type: Number, attribute: "initial-page-size" })
   initialPageSize = 25;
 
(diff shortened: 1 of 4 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
Applied a cleanup pass: trimmed the mirroring comments, updated the stale applied-once prop docs on the table, dropped a redundant test. @esphbot review

bdraco:
Checked the reconnect race: it can't happen. _devicesLoaded is written exactly once, events.ts sets it true on the first INITIAL_STATE and nothing ever resets it to false, so a reconnect never re-fires the false to true transition that gates loadPreferences in dashboard.ts. The reconnect snapshot goes through applyPreferences on the app shell, which only touches theme and settings prefs, not the dashboard table fields. The only path that re-runs loadPreferences is a fresh dashboard mount after navigation, which fetches from the backend by design and predates this PR. Leaving as is.

esphbot (approved):
No blocking issues found.

bdraco:
Took the suggestion: the persist rejection is now logged via console.warn instead of the empty catch, the three catch copies are folded into one updatePreferences call, and a test pins that a failed write logs and keeps the mirror. The mirror is intentionally kept on failure, cosmetic prefs, self healing on next mount, per your analysis. @esphbot review

esphbot (approved):
No blocking issues found.

PR #1147, opened 2026-07-08
Pinpoint dedented and over-indented lines in more YAML indent errors

bdraco:

The indentation analysis behind the invalid banner and the hover auto fix misread several shapes and blamed a nearby marker instead of the line that moved: a key dedented out of its block, a property dedented to the list marker column, a first child indented too deep, and a property indented too deep right after its marker. Each shape now resolves to a repair on the line the user actually moved, using the sibling structure around the blamed line as evidence.

Applying a fix could also fail with a false "config changed" message; the staleness check re ran the analysis anchored at the fix line, which cannot reproduce a diagnosis anchored at the scanner's blamed line. The check is now structural, the fix records the target line's indent and only applies while that line still has the same key and indent.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/yaml-editor.ts
+++ b/src/components/yaml-editor.ts
@@ -31,11 +31,12 @@ import { ESPHOME_YAML_INDENT, esphomeYaml } from "../util/esphome-yaml-lang.js";
 import { idleCompletion } from "../util/idle-completion.js";
 import { getKeyPath, isInsideBlockScalar } from "../util/yaml-ast.js";
 import { createYamlCompletionSource } from "../util/yaml-completion.js";
-import { analyzeIndentMismatch, type YamlAutoFix } from "../util/yaml-error-analysis.js";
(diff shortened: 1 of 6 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
Extracted the two new shapes into named helpers in 71acfe08; continuationScalarFix and misplacedOpenerFix each return IndentMismatch | null and analyzeIndentMismatch is now a dispatch on the line above the blamed one, the walk keeps only the pre-existing shallowerIndent tracking. Behaviour is pinned by the existing per-shape tests, all 4690 pass unchanged.

@esphbot review

esphbot (approved):
No blocking issues found.

PR #1148, opened 2026-07-08
Dedent a section head nudged off its level instead of the generic hint

bdraco:

A top level section key nudged one space to the right, for example i2c: indented to column 1, closes the list above it at a level nothing occupies; the scanner blames the key line but the analysis had no matching shape, so the banner showed the generic indentation hint with no auto fix. A key sitting strictly between a list's parent level and its markers now dedents back to the parent level, with the usual one click repair.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1151, opened 2026-07-08
Repair top-level indent slips without marker context

bdraco:

Two indent slips at the top level had no repair because the analysis bailed at column 0. First, a top level key such as wifi: whose first child ssid: went one step too deep; the opener walk refused to treat a column 0 key as the block opener, so the over deep child reading never ran. It now does, and the banner offers the dedent on the ssid line. Second, a childless section head nudged one space off the margin, for example captive_portal: at column 1 with only plain mappings above it; the walk reached the top level and gave up. It now dedents the key back to the margin, but only when the key opens no deeper block, since a nudged key with deeper children could as well belong one level in.

The generic indentation hint also claimed every problem was about list item markers; it is now worded for plain mappings too.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1161, opened 2026-07-09
Keep the app header and dashboard toolbar pinned on page-scroll views

### LINKED ISSUE #1923: [Bug] Have header consistent over Views
### What happened?

Currently in List View screen have non-scrollable header view easy and convinient access to View switcher, Search input field and ...menu.
But in Grid and YAML all the window scrolls, leading to extendive scrolling to reach commonly used controls.

IMHO this part to be common header with instant access.
Image

### Dashboard version

Device Builder 1.1.0 ESPHome 2026.6.5

### How did you install it?

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

In List view the app header and the toolbar (view switcher, search, Filters, count row) never scroll away because the table scrolls internally. Grid and YAML views scroll the whole page, so all of those controls scroll off the top and reaching search or the view switcher on a long device list means scrolling all the way back up. The legacy dashboard kept its command header position fixed while the card grid scrolled, so the controls were always reachable there; the page-scroll views lost that.

This pins both on the page-scroll views with position sticky: the app header sticks to the top of the scrollport, and the dashboard toolbar sticks directly below it in cards and yaml modes, tracking the compact mobile header via the existing height token. The toolbar gets an opaque surface background and a small bottom padding so content clips cleanly while stuck. Table view is untouched (its host never scrolls, so sticky never engages), and the page-scroll override from #438 is unchanged; sticky relies on it. z-indexes slot under the discovered-devices pill (5) so its expanded panel still paints over the toolbar, and under every fixed overlay (select bar, dropdown menus, drawer).

Verified headlessly against a 36-device backend: cards and yaml scrolled 900px keep header + toolbar pinned with the actions menu usable, the expanded discovered panel paints above the stuck toolbar and slides under the header, the drawer covers everything, table view still does not page-scroll, and the 390px viewport stacks the 40px compact header correctly.

The sticky strip also stays on mobile, where the wrapped toolbar rows take more height; that matches List view, which already keeps the same rows permanently visible there. If the maintainers prefer, the sticky rules could be scoped to desktop widths only.

No backend change; this is CSS only.

- fixes https://github.com/esphome/device-builder/issues/1923

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/dashboard/device-grid-styles.ts
+++ b/src/components/dashboard/device-grid-styles.ts
@@ -159,13 +159,14 @@ export const deviceGridStyles = css`
     padding: var(--wa-space-l) var(--content-gutter);
   }
 
-  /* When the grid follows the toolbar's count row (whether directly
-     or with the empty-search pivot wedged between), the count row
-     already provides spacing above the first card. Tighten the
-     grid's top padding so the two rows don't double up. */
+  /* When the grid follows the toolbar (whether directly or with the
+     empty-search pivot wedged between), the toolbar's own bottom
+     padding owns the seam — it has to live on the toolbar side so it
+     survives sticking. Zero the grid's top padding so the gap isn't
+     paid twice. */
   .toolbar + .devices-grid,
   .toolbar + .empty-search + .devices-grid {
-    padding-top: var(--wa-space-xs);
+    padding-top: 0;
   }
 
   /* Only the configured-device grid needs FAB clearance: it's the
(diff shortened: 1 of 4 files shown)


bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1165, opened 2026-07-09
Keep yaml_only fields in YAML key completion

bdraco:
# What does this implement/fix?

Fixes a regression where YAML key completion stopped suggesting fields the schema marks `visibility: yaml_only`. The 2026.7.0b1 catalog sync (esphome/device-builder#1915) started mapping that upstream hint onto the catalog's `hidden` flag, and the key completion provider dropped hidden entries entirely; under `esp32: framework:` the popup lost `advanced`, `components`, `sdkconfig_options`, `platform_version`, `release` and `source`, which are exactly the fields the hint says must be edited in YAML.

The provider now keeps hidden entries in the completion list, demoted the same way advanced fields are. The visual form also treats them like advanced fields when the YAML already sets them: `isEntryVisible` shows a hidden entry whose value is present so the form reflects what the config contains, and `validateEntry` validates a set value normally while still never blocking on an unset one. New adds and unset fields stay hidden from the form, which is what the hint means.

**Related issue or feature (if applicable):**

- no linked issue, regression surfaced after the 2026.7.0b1 catalog sync

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [ ] New feature (non-breaking change which adds functionality) — `new-feature`
- [ ] Enhancement to an existing feature — `enhancement`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [ ] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `docs`
- [ ] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Checklist

- [x] The code change is tested and works locally.
- [x] `npm run lint` passes.
- [x] `npm run test` passes.
- [x] Tests have been added to verify that the new code works (where applicable).

PR #1166, opened 2026-07-09
Re-anchor trailing-space positions onto the line's own pair

bdraco:

Fixes idle value completion going silent when the caret rests after `key: ` with a trailing space and more lines follow. `resolveAnchoredNode` only re-anchored onto the line's own pair when the raw resolution found no enclosing pair at all, which covers the end of document case; mid document the trailing whitespace resolves into an enclosing container that does have an ancestor pair, so the anchor stayed on the wrong node, `getKeyPath` collapsed to the top level key and the popup opened with no options. Typing the same key without the space worked, which is how it surfaced.

The re-anchor now always fires when the caret sits in trailing whitespace after the line's content; the no-enclosing-pair fallback stays for the other empty value shapes. Pinned at both layers, `getKeyPath` mid document and the completion source end to end.

- no linked issue, found while testing #1164 and #1165 together

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/util/yaml-ast.ts
+++ b/src/util/yaml-ast.ts
@@ -227,18 +227,23 @@ export function getKeyPathWithListIndices(
 
 /**
  * The deepest node at *pos*, re-anchored for the empty-value case: a
- * cursor in an empty value / trailing whitespace (``key: ``) resolves to
- * the document root, not the pair, so re-resolve on the line's last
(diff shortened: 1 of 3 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1169, opened 2026-07-10
Accept case-only differences when validating select options

bdraco:

The ESP32 Platform editor showed a false "Not a valid option" error on the `variant` field for any generic-ESP32 device, even though the config compiles and flashes fine.

The catalog's `variant` options are uppercase (`ESP32`, `ESP32C2`, …, from esphome's `VARIANTS` constants and `cv.one_of(*VARIANTS, upper=True)`), but boards and the field's own description write lowercase (`variant: esp32`). `validateEntry` did an exact-case match against the option values, so `esp32` failed even though `upper=True` normalizes it at validate time and the form already resolves the display value case-insensitively.

The fix accepts a case-only difference: the option check now also passes when `nearCanonicalOption` finds a case-insensitive match (the existing helper that powers the "did you mean" hint), so `esp32` against an `ESP32` option no longer hard-fails. A value that matches no option even case-folded (`esp32xx`) still errors. This is general, not variant-specific: it fixes every `cv.one_of(..., upper=True / lower=True)` enum where the written value's case differs from the catalog option's.

Verified live: a generic-ESP32 device (`variant: esp32`) no longer shows "Not a valid option" on the Variant field.

Surfaced while testing esphome/device-builder#1948; no tracking issue.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/util/config-validation.ts
+++ b/src/util/config-validation.ts
@@ -288,8 +288,15 @@ export function validateEntry(entry: ConfigEntry, raw: unknown): ValidationError
   // for fields that opt into custom values (combobox-style entries treat
   // `options` as suggestions, not a fixed set).
   if (entry.options && entry.options.length > 0 && !entry.allow_custom_value) {
+    const rawStr = String(raw);
     const allowed = entry.options.map((o) => o.value);
-    if (!allowed.includes(String(raw))) {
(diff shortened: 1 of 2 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1171, opened 2026-07-11
Show the resolved substitution hint under multi-value list rows

bdraco:

Single-value string fields show a resolved-substitution chip under the input (the `{}` badge with the value the `${var}` resolves to), but rows of a multi-value list never did. An sntp `servers:` list holding `${sntp_server_1}` / `${sntp_server_2}` / `${sntp_server_3}` rendered three bare inputs with no hint of what they resolve to, while `update_interval: ${sntp_update_interval}` right above showed its `6h` chip.

`renderMultiValueField` now renders the same `renderSubstitutionHint` under each row, resolved from the stored item value (not the escaped display form). Each row's input and hint stack in a new `multi-value-cell` column so the remove button stays beside the input; rows without a substitution reference render exactly as before. The unresolved-reference marker works per row too.

A first commit formats `test/util/device-search.test.ts` (one import-order line); it landed on main unformatted and the pre-commit `format:check` blocks every commit until it's fixed.

Verified live against a config mirroring the report: all three server rows show their resolved `pool.ntp.org` chips, matching Update Interval.

- reported directly with a screenshot of the sntp time editor; no issue filed

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/config-entry-form.styles.ts
+++ b/src/components/device/config-entry-form.styles.ts
@@ -295,6 +295,15 @@ export const configEntryFormStyles = css`
     cursor: not-allowed;
   }
 
+  /* Stacks a row's input above its substitution hint while the remove
+     button stays beside the input column. */
+  .multi-row .multi-value-cell {
+    flex: 1;
+    min-width: 0;
+    display: flex;
+    flex-direction: column;
+  }
+
   .combobox-input {
     font-family: inherit;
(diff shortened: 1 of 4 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1172, opened 2026-07-11
Resolve variant depends_on against the board's implied variant

bdraco:

A config field gated on the top-level esp32 `variant` — the clearest case is `framework.advanced.sram1_as_iram` — stayed hidden in the visual editor whenever the board was selected but `variant:` wasn't written in the YAML. With

```yaml
esp32:
  board: esp32-poe-iso
  framework:
    type: esp-idf
    advanced:
      sram1_as_iram: true
```

the field only appeared after manually adding `variant: ESP32`.

The gate is `depends_on: variant` / `depends_on_value_any: ["esp32", "ESP32"]`, and `isEntryVisible` resolves `depends_on` only against explicit form values (local scope, then the component root). A board *fixes* its variant (`esp32-poe-iso` → `esp32`, exposed as `board.esphome.variant`) even when the user never writes `variant:`, but that implied value never entered the resolution, so `depends_on_value_any.includes(undefined)` was false and the field was filtered out.

`renderFilterOptions` now seeds the platform section's `rootValues` with the board's `esphome.variant` before `depends_on` resolution, layered *under* explicit values so YAML still wins. It's scoped to the platform section (`sectionKey === board.esphome.platform`), so other component forms and a boardless variant-only config are untouched. `variant` is the only board-implied `depends_on` target in the whole catalog, so nothing else changes. The seed runs after any `rootValues` override, so the nested renderer's `scopeValues([])` path inherits it too.

Completes the `sram1_as_iram` surfacing work (backend esphome/device-builder#1952 exposed the field + variant gate; this is the frontend half that makes the gate resolve from the board).

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/config-entry-render-filter.ts
+++ b/src/components/device/config-entry-render-filter.ts
@@ -105,6 +105,29 @@ export interface RenderFilterSource {
    *  covered without remembering to pass it. A nested-scope caller whose local
    *  ``values`` isn't the root passes an explicit ``rootValues`` override. */
   values?: Record<string, unknown>;
+  /** The YAML section this form renders (``esp32`` / ``wifi`` / ``sensor`` …).
+   *  Used to seed board-implied values only on the platform section. */
+  sectionKey?: string;
+}
+
+/**
+ * Layer the board's implied ``esphome`` fields under *rootValues* when the form
+ * renders its platform section (``esp32`` / ``esp8266`` / ``rp2040``).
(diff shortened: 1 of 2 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1174, opened 2026-07-11
Hydrate the component body in the editor hover's catalog fallback

bdraco:

Hovering a catalog-only key in the YAML editor showed no tooltip. Example: `sram1_as_iram` under `esp32: framework: advanced:` has no hover, while the Visual Editor panel renders its full description.

`resolveHoverTarget`'s catalog fallback read `config_entries` off the slim `getComponents` index, which never carries bodies (they hydrate lazily through `components/get_component_bodies`), so the fallback always searched an empty array. Keys whose docs come from the schema bundle (`framework:`) still worked, which made the miss look depth dependent; any key that only the catalog documents was silent.

The fallback now hydrates the component body through the shared `fetchComponent` cache, the same path the completion source uses, and resolves the entry at the exact nested path via `descendNestedEntries`, falling back to the previous leaf-key search when the YAML structure doesn't mirror the body nesting. The hover test fixture no longer puts `config_entries` on the slim index, so the suite exercises the production shape.

- reported directly with screenshots of the missing sram1_as_iram hover; no issue filed

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/yaml-editor.ts
+++ b/src/components/yaml-editor.ts
@@ -573,15 +573,16 @@ export class ESPHomeYamlEditor extends CodeMirrorEditorElement {
         }),
         lintErrorLineGutter
       );
+      // The device's platform/board, read per invocation so completion
+      // and hover share the structured editor's hydrated-body cache bucket.
+      const getDeviceTarget = () => ({
+        platform: this.board?.esphome.platform,
+        boardId: this.board?.id,
+      });
       // Schema-driven completion off the components catalog.
       extensions.push(
         autocompletion({
-          override: [
-            createYamlCompletionSource(this._api, () => ({
-              platform: this.board?.esphome.platform,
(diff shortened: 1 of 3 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
Applied in 19b978bc, the fallback now checks catalog.byId.has before fetching so substitutions and friends never reach the wire, with a test pinning that no request is made. Same commit also threads the device target into the hover so it reuses the body cache bucket completion and the structured editor already fill, and reworks the test to stub getComponentBodies instead of module-mocking fetchComponent.

@esphbot review

esphbot (approved):
No blocking issues found.

PR #1175, opened 2026-07-11
Surface only YAML-present advanced fields in the section editor

bdraco:

In the structured (visual) device editor the rule is "a field present in the YAML always shows". That works for basic fields, but for advanced fields it over-fired: as soon as one advanced field carried a value (e.g. `id:` / `internal:` on the starter-kit `switch.gpio`), the whole "Advanced settings" section flipped open and dumped every empty advanced field on screen (`command_retain`, `device_id`, `disabled_by_default`, `entity_category`, …), with the toggle stuck on and showing a large count. So one field the user set in YAML dragged a dozen fields they never touched into view.

Two coupled mechanisms in `config-entry-form.ts` drove it. `_renderWithAdvancedSection` builds its plan with `showAdvanced: true` and paints the entire advanced group whenever `forceOpen` (any material-value advanced field) is set. And the `updated` hook mirrors that by emitting `advanced-toggle(true)`, which flips the host's `showAdvanced` on, re-rendering everything expanded.

The fix broadens the section editor's per-host gate flag (`gate-all-advanced` to `gate-advanced`, set only by `device-section-config.ts`). Under it, YAML-present fields (basic or advanced) paint inline, and only the empty advanced fields stay behind the control, now a real unlocked toggle that is off by default and counts just the hidden ones. The `updated` auto-flip is skipped under the flag (the pre-filled field already paints inline, so nothing needs the section forced open). Automation action/trigger/condition nodes and the script editor leave the flag off and keep their existing inline / force-open paint; the `inlineAdvanced` / `open` / `locked` / `count` expressions all reduce to today's values when the flag is off.

Verified live on the starter-kit config: the GPIO Switch now shows `id` / `internal` / `setup_priority` (the YAML-present fields) inline with "Show advanced settings (6)" off; turning the toggle on reveals the empty fields; `captive_portal` still shows just the control (the earlier all-advanced behaviour is preserved).

- fixes <link to issue>

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/config-entry-form-plan.ts
+++ b/src/components/device/config-entry-form-plan.ts
@@ -24,6 +24,8 @@ export interface FormRenderPlan {
   clusters: ConstraintCluster[];
   /** Keys folded into a cluster, dropped from the normal flow. */
   memberKeys: Set<string>;
+  /** Each cluster keyed by its first member's key — the slot it paints at. */
+  clusterByFirstKey: Map<string, ConstraintCluster>;
   /** Plain (non-exclusive, non-cluster) entries that pass the filter. */
   visible: Set<ConfigEntry>;
 }
@@ -36,11 +38,12 @@ export function buildFormRenderPlan(
 ): FormRenderPlan {
   const ordered = orderExclusiveGroups(entries);
   const { clusters, memberKeys } = buildConstraintClusters(entries, requiredGroups);
+  const clusterByFirstKey = new Map(clusters.map((c) => [c.members[0].key, c]));
   const nonExclusive = entries.filter(
     (entry) => !entry.exclusive_group && !memberKeys.has(entry.key)
   );
   const visible = new Set(filterRenderable(nonExclusive, values, opts));
-  return { ordered, clusters, memberKeys, visible };
+  return { ordered, clusters, memberKeys, clusterByFirstKey, visible };
 }
 
 /**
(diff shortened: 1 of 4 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
On suggestion #1 (empty reveal when every advanced field is YAML-present): leaving it as-is, deliberately. The naive `&& !(this.gateAdvanced && gatedAdvanced.length === 0)` fold would regress nested-advanced reachability.

`showControl` is driven by `hasAdvanced` (recursive, `anyAdvancedEntry`) precisely so the control still appears when the only advanced field is a **child of a basic group** — the section `showAdvanced` toggle is what reveals it (the nested renderer reads `renderFilterOptions(this).showAdvanced`). That case is pinned by the existing test "shows the control when the only advanced field is nested in a mixed group". Under `gateAdvanced`, such a component has `gatedAdvanced.length === 0` (the advanced field is nested, not a top-level unit) while `hasAdvanced` is true, so the suggested suppression would hide the control and strand the nested child — a silent regression (that test runs with `gateAdvanced=false`, so it wouldn't catch it). A provably-safe version needs a recursive "any hidden nested advanced" check, which is disproportionate to a rare, harmless empty toggle. So: considered skip.

Also pushed 3b4197ba, a no-behaviour-change tidy from a cleanup pass:
- lifted `clusterByFirstKey` onto the render plan (one shared map instead of `_makeItemRenderer` and the split each rebuilding it);
- named the force-open kernel `_effectiveForceOpen()` and consumed it from `open`, `locked`, and the `updated()` guard, which spelled the same idea three ways — so the render split and the auto-toggle mirror can't drift;
- run the prefilled/empty partition in one pass under `if (gateAdvanced)`, skipping the map + closure on the non-gated hosts.

@esphbot review

esphbot (approved):
No blocking issues found.

PR #1181, opened 2026-07-11
Reopen a deferred install as an Install with the queued message

bdraco:

Reopening a running deferred install from the Firmware tasks drawer showed it as "Compile — device" and finished with "Compilation complete!" instead of the queued update message. The command type derivation only recognized an install by its dependent upload, and a deferred install is a lone COMPILE with none; the drawer row had the same blind spot. A shared `effectiveJobType` now presents a deferred install as the Install it is, in the drawer row, the reopened dialog title, and the completion banner.

Pairs with the backend change that re-checks reachability after the build: the follow result's queued flag moved to a dedicated `queued_update_armed` key with armed semantics (a failed deferred compile no longer looks queued), and the queued check runs before the dependent flash chase, so a chain the backend converted mid build (device went offline during the compile, held upload cancelled) lands on the queued banner instead of priming into the cancelled upload. The same path shows the queued banner for an OTA upload that failed against an offline device, matching what the backend armed.

Companion backend PR: esphome/device-builder#1964

- fixes the wrong completion message when reopening a deferred install from Firmware tasks

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/api/esphome-api.ts
+++ b/src/api/esphome-api.ts
@@ -54,6 +54,7 @@ import type {
 import type {
   FirmwareBinary,
   FirmwareJob,
+  FirmwareJobResult,
   RemoteBuildSubmitTarget,
 } from "./types/firmware-jobs.js";
 import type {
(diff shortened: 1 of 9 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1182, opened 2026-07-11
Mark every YAML section header with its domain icon

bdraco:

Each config section now reads as the same thing in the navigator and the YAML editor. The editor pins a quiet domain icon to the end of every top level section header (`esphome:`, `wifi:`, `switch:`, `sensor:`, and so on), the same glyph the navigator row already shows for that domain, so you can tie a block in the code back to its navigator entry at a glance.

The markers are always on; they are derived from the document rather than the current selection, so every section carries its glyph whether or not it is selected. They are monochrome and sit in the content flow sized to the line, so they never shift the line number gutter. They reuse the existing `iconForDomain` registry and the top level key lexer (`TOP_LEVEL_KEY_RE`), so there is no new YAML parsing, and the marker set tracks live edits as sections are added or removed.

This came out of editor feedback that it was hard to visually tie the navigator, the section form, and the YAML together. Earlier iterations tried per domain colour and a text label chip; both were dialled back to this calmer, icon only, no colour form.

- Editor UX enhancement from navigator / editor linking feedback; no filed issue.

## Screenshots

Quiet domain icons on each top level section header in the editor, matching the navigator glyphs.
Screenshot 2026-07-11 at 11 53 40 AM

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/yaml-editor.ts
+++ b/src/components/yaml-editor.ts
@@ -57,6 +57,7 @@ import {
   relintEffect,
   type YamlDiagnosticsDetail,
 } from "../util/yaml-lint-backend.js";
+import { withBase } from "../util/base-path.js";
 import { TOP_LEVEL_KEY_RE } from "../util/yaml-section-lexer.js";
 import type { YamlSection } from "../util/yaml-sections.js";
 import {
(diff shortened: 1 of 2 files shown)


bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

bdraco:
Closing this without merging.

The goal was to help tie the three editor panes (navigator, section form, YAML) together visually. We went through a few treatments: per domain colour on the nav icons and YAML block, then an icon plus text label chip on the selected block, then this monochrome domain icon pinned to every top level section header (inert via `pointer-events: none`, with the ESPHome logo for `esphome:`).

Each version worked, but on review the icons read as decorative rather than additive. They duplicate the section key that is already on the line (`wifi:` next to a wifi icon), and the thing that actually ties the panes together is the interaction that already exists: clicking a navigator row highlights and scrolls to the block, and moving the caret updates the navigator. A passive marker layered on top of that does not add much.

The real reason for closing is that we did not have a clear enough picture of the underlying friction to justify adding persistent UI. If the "hard to tie it together" feeling comes back with a concrete moment attached (YAML to nav, nav to YAML, form to YAML, or deep nesting), that gives us something specific to design for instead of guessing. Until then, better to not ship chrome.

The branch is preserved if any of this is worth reviving later.

PR #1197, opened 2026-07-12
Keep Edit clickable while a job runs

### LINKED ISSUE #1196: Edit is busy-disabled in the drawer and card but not the table row
Small consistency gap entrenched by #1193's tests: the Edit affordance has two different busy policies across surfaces for the same device.

- Drawer footer Edit: `?disabled=${this.busy}` (`device-drawer.ts`) — disabled while a job runs (now pinned by `device-drawer-busy-actions.test.ts`).
- Table row Edit button (`table-columns.ts`): no busy gate — always clickable.
- The editor page itself is expressly designed for mid-job use (busy-routed Update/Install, `_showActiveJobProgress`).

Edit merely navigates to the editor; there's no job-starting hazard. Either enable the drawer's Edit while busy (matching the row and the editor's capabilities — probably the right call) or gate the row's too. The card's Edit button is also busy-disabled (`device-card.ts:227`) — same decision applies there.

bdraco:

Removes the busy-while-building disable from the Edit buttons on the device card, the table row, and the drawer footer. Edit only navigates to the editor page, which is expressly designed for mid-job use (busy-routed Update/Install that re-attaches to the running job), so the gate prevented nothing real — it just blocked the most common "check the YAML while it builds" flow.

One correction to the issue text: the table row's Edit button *was* also busy-disabled (`table-columns.ts`), so all three surfaces are converted rather than two. Tests updated: the drawer suite's "Edit keeps disabling" pin is flipped, and the card/table busy suites now pin Edit as clickable.

Destructive/mutating actions (rename, duplicate, clean build files, archive, delete) keep their busy gates — those genuinely conflict with a running job.

- fixes https://github.com/esphome/device-builder-frontend/issues/1196

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/dashboard/device-drawer.ts
+++ b/src/components/dashboard/device-drawer.ts
@@ -355,7 +355,6 @@ export class ESPHomeDeviceDrawer extends LitElement {
         <div class="footer">
           <button
             class="action action--primary"
-            ?disabled=${this.busy}
             @click=${() => this._emitAction("edit-device")}
           >
             <wa-icon library="mdi" name="pencil"></wa-icon>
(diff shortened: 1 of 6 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1205, opened 2026-07-12
Batch the install dialog's log streaming like the command dialog

### LINKED ISSUE #1203: Batch install-dialog log streaming instead of per-line array copies
Perf debt noted during #1201's review, pre-existing and now present at three sites in `src/components/firmware-install-dialog/install-flow.ts`: every streamed output line does `host._logLines = [...host._logLines, line]` (web-serial flow, `compileAndWait`, `waitForRunningJob`). That's O(n²) accumulation over a compile log — thousands of lines each copying the whole array — because `_logLines` is `@state` and consumers assume reference-change semantics.

Fix as one change converting all three sites together (diverging in one of three would be worse): mutate a persistent buffer (`push`) plus a coalesced `requestUpdate()` — microtask- or rAF-batched flush — so the render sees a stable reference cadence without the per-line copy. The command dialog's `_enqueueLine`/`_flushPendingLines` buffering in `command-dialog/commands.ts` is prior art for the batching shape.

bdraco:

The firmware-install dialog appended streamed log output one line at a time via `_logLines = [...prev, line]` — a full array copy per line, O(n²) over a compile log, at three sites (the web-serial esptool stream, `compileAndWait`, and #1201's `waitForRunningJob`). The command dialog already solved the identical problem with an rAF-coalesced buffer (#348).

Per the direction on #1203, this extracts that mechanism into a shared `LineBatcher` util (`src/util/line-batcher.ts`: `enqueue` schedules one flush per animation frame, `flush` drains immediately, `reset` drops pending lines and cancels the frame) and points both dialogs at it:

- **Command dialog**: `_enqueueLine` / `_flushPendingLines` / `_resetPendingLines` keep their exact API (callers in `commands.ts` and existing tests untouched) but delegate to the shared batcher.
- **Install dialog**: the three per-line copy sites now enqueue; flushes land at the terminal follow callbacks, `_fail` (the expanded error log must show every line, not race the rAF), and `_detachStream`; `_init` resets the pending batch alongside `_logLines`.

The four install-dialog test harnesses gained synchronous stand-ins for the sink so their assertions stay deterministic, and a new unit suite pins the `LineBatcher` contract (one append per frame, immediate flush, reset cancels the frame, batching across frames).

- fixes https://github.com/esphome/device-builder-frontend/issues/1203

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/command-dialog.ts
+++ b/src/components/command-dialog.ts
@@ -36,6 +36,7 @@ import { espHomeStyles } from "../styles/shared.js";
 import { initialDarkMode } from "../util/dark-mode.js";
 import { configurationStem, downloadAnsiText } from "../util/download-text.js";
 import { LightDismissController } from "../util/light-dismiss-controller.js";
+import { LineBatcher } from "../util/line-batcher.js";
 import { dispatchShowLogsAfterInstall } from "../util/post-install-logs.js";
 import { registerMdiIcons } from "../util/register-icons.js";
 import { RunTimerController } from "../util/run-timer-controller.js";
@@ -142,8 +143,9 @@ export class ESPHomeCommandDialog extends LitElement {
 
   // rAF batch buffer for streamed output — coalesce per-line writes
   // into one render per frame instead of one per line (#348).
-  private _pendingLines: string[] = [];
-  private _flushScheduled = 0;
+  private _lineBatch = new LineBatcher((batch) => {
+    this._lines = [...this._lines, ...batch];
+  });
 
   // Distinguishes user-stopped from backend-failed. Both flip _state to "error"
   // but only real failures get the reset-build-env hint.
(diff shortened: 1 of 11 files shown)


bdraco:
@esphbot review

esphbot (changes requested):
Blocking issues found.

bdraco:
Both points addressed:

- **Warning (missed consumer):** `downloadInstallLogs` now flushes the pending batch before reading `_logLines`, mirroring command-dialog's `_downloadOutput`.
- **Suggestion #1 (dialog-level coverage):** new `firmware-install-dialog-log-batching.test.ts` drives a REAL `LineBatcher` through the dialog: a buffered line stays pending until `_fail` / `_detachStream`, which must land it synchronously — a removed flush call now fails a test.
- **Suggestion #2 (logs-dialog's capped copy):** deferring as suggested; filing a follow-up issue so it doesn't get lost.

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
Suggestion #1 taken — the batching suite now drives `renderLogs`' download button with a real batcher: one landed line plus one buffered, and `downloadAnsiText` must receive both (removing the flush now fails the test).

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1207, opened 2026-07-12
Move logs-dialog's capped log batching onto LineBatcher

### LINKED ISSUE #1206: Move logs-dialog's capped log accumulation onto LineBatcher
Noted in #1205's review: `logs-dialog.ts` still carries its own (capped) variant of the per-line log accumulation pattern that #1205 consolidated for the command and install dialogs via the shared `LineBatcher` util.

Follow-up: convert the logs dialog onto `LineBatcher` too. Its stream is capped (drops the oldest lines past a limit), so the util would need an optional cap — e.g. a `maxLines` constructor option applied at flush time — rather than a third bespoke implementation.

bdraco:

Converts the last hand-rolled copy of the rAF log-batching pattern — the logs dialog's — onto the shared `LineBatcher` util that #1205 extracted for the command and install dialogs.

The logs dialog's variant differed in one way: it's capped. That splits into two caps with different owners, and the conversion respects the split:

- **Pending-buffer bound** (rAF doesn't fire in hidden tabs, so a flood piles up unflushed): moved into `LineBatcher` as an optional `maxLines`, with the same 2× headroom trim the dialog hand-rolled (one slice per `maxLines` pushes, not per push). The other two dialogs pass no cap and are unchanged.
- **Visible-buffer cap** (newest `MAX_LOG_LINES` after merge): stays in the dialog's `_appendCapped` append callback — it owns that array, and the direct recovery-path append (`setSerialOpenFailed`) uses it too.

The dialog's `_enqueueLine` / `_flushPendingLines` / `_resetPendingLines` keep their exact API and delegate, so the serial/OTA stream callers and existing tests are untouched (one test's internal `_pendingLines` peek repointed at the batcher). New unit case pins the `maxLines` headroom-trim behavior at the util level; the existing logs-dialog batching suite (coalescing, both caps, recovery append, reset) passes unchanged.

- fixes https://github.com/esphome/device-builder-frontend/issues/1206

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/logs-dialog.ts
+++ b/src/components/logs-dialog.ts
@@ -21,6 +21,7 @@ import { fullscreenMobileDialog } from "../styles/dialog-mobile.js";
 import { espHomeStyles } from "../styles/shared.js";
 import { initialDarkMode } from "../util/dark-mode.js";
 import { configurationStem, downloadAnsiText } from "../util/download-text.js";
+import { LineBatcher } from "../util/line-batcher.js";
 import { notifyError } from "../util/notify.js";
 import { registerMdiIcons } from "../util/register-icons.js";
 import { logsDialogStyles } from "./logs-dialog.styles.js";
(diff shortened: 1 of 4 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1209, opened 2026-07-12
Suppress the slow-build offload hint on ESPHome Desktop

bdraco:

When a local compile runs past five minutes, the dialogs nudge the user toward "send builds to a faster machine" (Settings → build offload). On ESPHome Desktop the build machine already *is* the user's computer — if their desktop is slow, there is likely no faster machine to offload to, so the nudge is noise.

`shouldShowOffloadHint` now takes a `desktop` flag and stays hidden when it's set. Both dialogs feed it from the handshake's `desktop_version` via the existing `desktopVersionContext`, so all three surfaces are covered through the one gate: the command dialog's inline suggestion slot, its timer-detail popover nudge, and the install dialog's suggestion slot. Non-desktop behaviour is unchanged.

- None (requested directly by maintainer)

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/command-dialog.ts
+++ b/src/components/command-dialog.ts
@@ -129,8 +129,6 @@ export class ESPHomeCommandDialog extends LitElement {
   @state()
   _pairings: Map<string, PairingSummary> | null = null;
 
-  // Suppress the offload hint on the Desktop app — the build machine already
-  // is the user's computer.
   @consume({ context: desktopVersionContext, subscribe: true })
   @state()
   _desktopVersion = "";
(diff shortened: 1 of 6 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1216, opened 2026-07-12
Pin the block-scalar cursor-path test to the exact resolved path

bdraco:

Follow up to #1215, which merged while this was in flight. The block scalar test in `test/util/yaml-cursor-paths.test.ts` asserted only that the derived path doesn't contain a phantom `id` field, which would also hold if the test accidentally targeted the `lambda: |-` line itself. Pin the exact resolved paths (`["sensor", "lambda"]` plus the indexed sibling) so the assertion can only pass on the scalar content line with the AST fallback doing its job.

Prompted by a Copilot review round on #1215; the other three comments in that round confused the test file's line numbers with the YAML fixture's own 1-indexed lines and needed no change.

- fixes <link to issue>

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/test/util/yaml-cursor-paths.test.ts
+++ b/test/util/yaml-cursor-paths.test.ts
@@ -40,9 +40,13 @@ describe("pathsForYamlLine", () => {
   });
 
   it("does not treat block-scalar content as fields", () => {
-    // ``id: looks_like_a_key`` inside the lambda body is literal text.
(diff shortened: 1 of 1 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1217, opened 2026-07-12
Fall back to the dependency's default when resolving depends_on visibility

### LINKED ISSUE #1972: [Bug] Missing fields in SPI Bus configuration
### 2. Add SPI Bus with default settings

Image

Image

### 3. MISO and MOSI PINs got added to YAML, but there are no fields shown in the structured editor

Image

Image

### 4. Open the `Type` Dropdown and re-select `single` (which is the default setting)

Image

### 5. MISO and MOSI fields are visible now

Image

### Dashboard version

1.4.4b3

### How did you install it?

Desktop app

### Where is it running?

Local — Windows

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

Adding an SPI bus with default settings writes `miso_pin`/`mosi_pin` to YAML but the structured editor doesn't show the fields until the user re-selects `single` in the Type dropdown — the value it was already displaying.

The cause is how `depends_on` gates resolve: `miso_pin`/`mosi_pin` are gated on `type == single`, `type` defaults to `single`, and YAML omits fields sitting at their default. `isEntryVisible` compared the gate against the raw value map, so an absent `type` read as `undefined` and failed the gate, even though every renderer already displays the default for an unset field.

`isEntryVisible` now falls back to the dependency's `default_value` (looked up in the sibling schema list callers already have in hand) when the dependency is unset in both the local and root scopes. All call sites — render filter, form clusters, exclusive groups, constraint banners/clusters, deprecation notice, and `validateEntries` — pass their entry list through, so paint and validation stay in lockstep. An explicit value still beats the default, and omitting the sibling list keeps the old behaviour.

A catalog scan shows 34 gates broken the same way (`spi` miso/mosi/interface, `sn74hc595` clock/data pins, `esp32_hosted` SDIO pins, `sensor.ads1118`, `output.modbus_controller`, `climate.haier`, `packet_transport`, …) — all fixed by this one fallback. Zero `depends_on_value_not` gates in the catalog have a defaulted dependency, so the fallback can only reveal fields the default legitimately satisfies, never newly hide one.

- fixes https://github.com/esphome/device-builder/issues/1972

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/add-component-form-filter.ts
+++ b/src/components/device/add-component-form-filter.ts
@@ -75,7 +75,8 @@ export function addFormNeedsUserInput(
       values,
       opts.presentComponents,
       opts.targetPlatform ?? null,
-      opts.rootValues
+      opts.rootValues,
+      entries
     );
   if (planNeedsUserInput(plan, isVisible)) return true;
   // Pure-cardinality groups with no cluster box surface a banner only when
--- a/src/components/device/config-entry-form.ts
+++ b/src/components/device/config-entry-form.ts
@@ -404,7 +404,14 @@ export class ESPHomeConfigEntryForm extends LitElement {
       cluster.members.some(
         (m) =>
           getIn(this.values, [m.key]) !== undefined ||
-          isEntryVisible(m, this.values, this.presentComponents, targetPlatform)
+          isEntryVisible(
+            m,
+            this.values,
+            this.presentComponents,
+            targetPlatform,
+            undefined,
+            this.entries
+          )
       );
     const renderedClusterKeys = new Set(
       plan.clusters.filter(clusterRenders).map((c) => c.members[0].key)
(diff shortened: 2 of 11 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
The nested-scope siblings note is now tracked in #1218 (carry the local sibling scope on RenderCtx instead of the top-level list). Both current fragilities it covers — nested defaulted deps degrading to the old hide, and the optional `entries` silently dropping the fallback — are latent-only today (0 of 46 nested-level gates have a defaulted dependency), so it's split out as structural hardening rather than folded into this fix.

PR #1223, opened 2026-07-12
Make RenderCtx.entries required

### LINKED ISSUE #1218: Carry the local sibling scope for the depends_on default fallback in RenderCtx
#1217 added a `siblings` parameter to `isEntryVisible` so an unset `depends_on` dependency falls back to the dependency entry's `default_value`. The four renderer call sites (`constraint-cluster.ts` ×2, `exclusive-group.ts`, `config-entry-form.ts`'s cluster check) feed it `ctx.entries` / `this.entries` — the form's **top-level** entry list — rather than the sibling scope of the level actually being rendered.

That's correct for every case in today's catalog (all 34 defaulted-dependency gates are root-level siblings; 0 of the 46 nested-level `depends_on` gates have a defaulted dependency), but it leaves two latent fragilities:

1. **Nested-scope mismatch.** If a future catalog schema puts a defaulted `type`-style gate inside a nested group that participates in a cluster/exclusive group, the fallback would resolve against the top-level list — silently falling through to hidden, or worse, false-matching a same-named top-level key's default (the name-collision caveat already noted in `isEntryVisible`'s docstring).
2. **Optional-typed silent degradation.** `RenderCtx.entries` is optional ("absent in lightweight contexts"), so any future ctx built without it silently loses the fallback with no type error. Both current constructions in `config-entry-form.ts` do set it, but nothing enforces that.

The deeper fix: carry the *local sibling scope at the level the ctx is rendering* (or make the sibling scope a required part of `RenderCtx`), instead of reusing the top-level label-lookup list. The contract is currently pinned only in the `RenderCtx.entries` doc comment.

Zero user-visible impact today — this is structural hardening so the #1972 fix can't quietly stop covering a future nested gate.

bdraco:

`RenderCtx.entries` was optional, but since #1217 it doubles as the `siblings` scope for `isEntryVisible`'s depends_on default fallback — a context built without it silently loses the fallback with no type error, and the label formatter fell back to a narrower lookup. Every production construction already sets it, so the optionality only ever *hid* omissions.

`entries` is now required. The dead absence fallback (`ctx.entries ?? cluster.members` in the cluster header's label resolution) goes with it, and the field's doc now pins why top-level is the correct sibling scope for every consumer: the cluster / exclusive-group / banner renderers only ever run at the top level (their paths are root-anchored — `buildFormRenderPlan` is never invoked with a nested list), and nested leaves get per-level siblings from `filterRenderable`'s own recursion. If those renderers ever become nested-capable, `entries` must be rebased alongside the paths.

The change proved its point immediately: the shared `makeRenderCtx` fixture and the constraint-cluster test's two hand-rolled ctxs (cast through `as unknown as RenderCtx`, so invisible to tsc) all omitted `entries` — exactly the silent degradation #1218 describes. They now supply the entry list they render.

No behaviour change in production paths.

- fixes #1218

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/config-entry-renderers-types.ts
+++ b/src/components/device/config-entry-renderers-types.ts
@@ -49,12 +49,14 @@ export interface RenderCtx {
   reactiveConstraintKeys: Set<string>;
   /** The form's top-level config entries, for resolving a label of a key that
    *  isn't in a given cluster's members (a cardinality key that's also an
-   *  ``exclusive_group`` member is dropped from the cluster). Absent in
-   *  lightweight contexts; callers fall back to a narrower lookup. Also fed
-   *  to ``isEntryVisible`` as the ``siblings`` scope for the depends_on
-   *  default fallback — top-level only, matching every defaulted-dependency
-   *  gate in today's catalog; a context without it loses the fallback. */
-  entries?: ConfigEntry[];
+   *  ``exclusive_group`` member is dropped from the cluster), and fed to
+   *  ``isEntryVisible`` as the ``siblings`` scope for the depends_on default
+   *  fallback. Top-level is the *correct* sibling scope for every consumer:
+   *  the cluster / exclusive-group / banner renderers only ever run at the
+   *  top level (their paths are root-anchored), and nested leaves get their
+   *  per-level siblings from ``filterRenderable``'s own recursion. If those
+   *  renderers ever become nested-capable, rebase this alongside the paths. */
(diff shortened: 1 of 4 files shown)


bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1231, opened 2026-07-14
Render per-option catalog descriptions in select menus

bdraco:

Companion to esphome/device-builder#2055 (esphbot's blocking finding on the catalog sync: the `web_server.auth.type` dropdown rendered full sentences instead of `basic`/`digest`).

The backend now ships an optional `description` per `ConfigValueOption`: when a value's schema docs read as prose, the label stays the value and the prose moves to `description`. This PR renders it:

- `ConfigValueOption` gains `description?: string`.
- `renderSelectField`'s strict `wa-select` shows the description as a quiet second line inside the option, reusing the existing `option-default-stack` shape and stacking with the *default* tag when the option is also the default.
- `esphome-options-combobox` (the `allow_custom_value` path) renders the same second line, and the type-ahead filter now also matches description text (typing "hashes" finds `digest`).
- New `.option-description-note` style beside `option-default-note` in the shared input styles: same quiet size, non-italic so prose reads as content rather than the default-tag annotation, wrapped at `42ch` so long sentences don't stretch the menu.

Old catalogs without the field render exactly as before (the stack only appears when a description or default tag is present).

Verified live against the backend branch: `web_server.auth.type` shows `basic` / `digest` with their sentences beneath and `basic` tagged default; `switch.gpio restore_mode` shows each enum value with its prose, `ALWAYS_OFF` tagged default.

- companion to esphome/device-builder#2055 (fixes the finding in https://github.com/esphome/device-builder/pull/2051#pullrequestreview-4691189630)

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/api/types/config-entries.ts
+++ b/src/api/types/config-entries.ts
@@ -12,6 +12,10 @@ export type ConfigPrimitive = string | number | boolean;
 export interface ConfigValueOption {
   label: string;
   value: string;
+  /** Prose explaining the choice, rendered as secondary text under the label.
+   *  Set when the catalog's per-value docs read as a sentence rather than a
+   *  label. */
+  description?: string;
   /** ESP32 variants that accept this value (lowercased, e.g. `esp32s3`); absent
    *  or empty means every variant. When the device resolves to an ESP32 variant
    *  the select filters to it; otherwise (non-ESP32 or unknown) all options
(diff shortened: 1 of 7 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1237, opened 2026-07-14
Give the remote-compute switch an accessible name

bdraco:

Follow-up to #1235 (a late Copilot review comment). The remote-compute `<wa-switch>` on the onboarding existing-server step was only labelled by a wrapping `<label>`, which doesn't reliably name a custom element for screen readers. Set `aria-label` directly on the switch (its title, "Only use this for remote compute"), matching the other `wa-switch` usages in the codebase.

The other Copilot comment (checking `RemoteBuildPeer.hostname` for the add-on pattern) needs no change: the Supervisor container hostname lands in `friendly_name` — which the helper already checks — while `hostname` is the SRV target (`esphome-builder-<id>.local`), which never carries the `-esphome[-channel]` suffix.

- Follow-up to esphome/device-builder-frontend#1235 (no public issue).

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/onboarding/onboarding-wizard-dialog.ts
+++ b/src/components/onboarding/onboarding-wizard-dialog.ts
@@ -301,6 +301,7 @@ export class ESPHomeOnboardingWizardDialog extends LitElement {
           <wa-switch
             ?checked=${this._remoteCompute}
             ?disabled=${this._saving}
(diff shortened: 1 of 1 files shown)


bdraco:
@esphbot review

PR #1238, opened 2026-07-14
Keep editor overlays above the find-and-replace bar

### LINKED ISSUE #2070: [Bug] Pop up from 3 dot menu next to Install appears behind the find and replace bar when open
### What happened?

Pop up from 3 dot menu next to Install appears behind the find and replace bar when open. See screenshot:

Image

### Dashboard version

2026.7.0b3

### How did you install it?

Home Assistant App

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

The 3 dot menu next to Install opens upward and rendered behind the find and replace bar. CodeMirror's search panel ships with `z-index: 300` in the library base theme, and nothing between it and the card body creates a stacking context, so it outranked the floating action row's `z-index: 10` context and painted over the open menu.

The YAML editor host now sets `isolation: isolate`, which keeps CodeMirror's internal z-indexes contained inside the editor; the menu and the action row paint above the panel again. This also fixes the same occlusion on the Secrets page, where the floating Save button was covered by an open find bar.

Verified in the live editor with a headless hit test at the overlap point: before the change the top element there is `cm-search cm-panel` on both pages; after it is the menu item, and Save on Secrets. Bumping the action row's own z-index instead was rejected because the mobile navigator drawer overlays the editor at `z-index: 100` and would be poked through.

- fixes https://github.com/esphome/device-builder/issues/2070

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/yaml-editor.ts
+++ b/src/components/yaml-editor.ts
@@ -172,7 +172,7 @@ export class ESPHomeYamlEditor extends CodeMirrorEditorElement {
       flex: 1;
       min-height: 0;
       /* Contain CodeMirror's internal z-indexes (search panel: 300) so they
-         can't paint over the action rows hosts float above the editor. */
+         can't paint over action rows that host pages float above the editor. */
       isolation: isolate;
     }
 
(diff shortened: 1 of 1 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1239, opened 2026-07-14
Don't lose the guided tour's create-device click to a mid-dispatch refresh

### LINKED ISSUE #2072: [Bug] Tour fails to recognize that “+ Create device” had been clicked
### What happened?

During my tests, there have been several cases where the tour failed to recognize that “+ Create device” had been clicked. In those cases, it does not proceed to the second step.

So far, I haven't been able to determine exactly when the error occurs.

I haven't noticed the error in the subsequent steps so far.

Image

### Dashboard version

1.5.1b4

### How did you install it?

Desktop app

### Where is it running?

Local — Windows

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

Clicking "+ Create device" on the guided tour's first step sometimes opened the wizard without the tour advancing; it stayed on "Step 1 of 11" with the wizard open in front of it, and every later attempt behaved the same.

Two problems combined:

1. Opening the wizard resets it to the method screen. When a previous wizard session was dismissed on a later screen (the hidden dialog keeps its steps mounted), that reset re-renders tour anchors while the click is still dispatching, and the tour's refresh tore down and re-added its click listener. A listener removed during dispatch is skipped even when re-added immediately, so the tour never saw the click. The refresh now leaves listeners on unchanged targets alone.
2. The intended backstop, auto-advancing into the dialog once its open animation finishes, could never fire: it required the current step's anchors to be gone, and the create button stays registered behind the modal. It now advances once the dialog has shown and the next step's anchor is visibly sized; a hidden dialog keeps its anchors registered but sizeless, so it cannot fire early. The in-dialog steps keep the old churn rule.

This is also why the reporter only ever saw step 1 fail: inside the wizard each screen swap unmounts the previous anchor, so the churn backstop rescues a missed click there.

Reproduced with real mouse input against a live backend (wizard parked on the board step, tour started, create button clicked: wizard opened, tour stuck on step 1) and verified the same sequence now advances to step 2.

- fixes esphome/device-builder#2072

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/guided-tour/esphome-guided-tour.ts
+++ b/src/components/guided-tour/esphome-guided-tour.ts
@@ -5,19 +5,16 @@ import { customElement, query, state } from "lit/decorators.js";
 import type { LocalizeFunc } from "../../common/localize.js";
 import { localizeContext, remoteComputeOnlyContext } from "../../context/index.js";
 import { espHomeStyles } from "../../styles/shared.js";
-import { isTypingTarget } from "../../util/typing-target.js";
 import { registerMdiIcons } from "../../util/register-icons.js";
+import { isTypingTarget } from "../../util/typing-target.js";
 import { guidedTourStyles } from "./esphome-guided-tour.styles.js";
-import { renderTourBubble, renderTourRecovery } from "./tour-bubble.js";
-import { TourBubbleFit } from "./tour-bubble-fit.js";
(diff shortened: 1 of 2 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1259, opened 2026-07-16
Gate the pairing display name on an explicit receiver_label_auto flag

### LINKED ISSUE #1254: Persist an explicit label_auto flag for offloader-side pairings
`pairingDisplayName()` (frontend `src/util/pairing-display-name.ts`) decides whether the offloader-side pairing label is "auto-derived" (and may therefore be replaced by the receiver's handshake `friendly_name`) by comparing `pairing.label` against `friendlyHostname(pairing.receiver_hostname)`.

That heuristic is a proxy for an explicit "was this label left at its prefill?" flag. The receiver side already has the real thing: the offloader sends `offloader_label_auto` in msg3, persisted as `StoredPeer.label_auto` and consumed by `peerDisplayName()`. The offloader side lacks the symmetric field.

Edge case (raised by Copilot on esphome/device-builder-frontend#1252): if the user edits a pairing's endpoint (hostname/IP) via the edit-endpoint dialog without changing the label, an originally auto-derived label no longer matches `friendlyHostname(new_hostname)`, so the heuristic flips to "custom" and the receiver's friendly name stops being shown.

Symmetric completion:
- Backend: add `receiver_label_auto: bool` to `StoredPairing` (offloader side), set from a new optional arg on the `remote_build/request_pair` command (the pair dialog already tracks `_receiverLabelTouched`).
- Frontend: send `receiver_label_auto: !_receiverLabelTouched` in `requestRemoteBuildPair`, and have `pairingDisplayName()` gate on the explicit flag instead of the `friendlyHostname` comparison.

Low priority — the current heuristic is correct for the common flow (pair, never edit the endpoint), and the friendly-name display is a cosmetic improvement, not a correctness requirement. This is the documented accepted tradeoff in #1252's body; filing so it isn't lost.

bdraco:

`pairingDisplayName` decided whether a build server's label was an auto-derived prefill (and could be replaced by the receiver's handshake friendly name) by comparing the label against the hostname; editing a pairing's endpoint changed the hostname but not the label, so the guess flipped to "custom" and the friendly name silently stopped showing. The receiver side already carries an explicit `label_auto` flag for exactly this; this gates `pairingDisplayName` on the symmetric `receiver_label_auto` instead, mirroring `peerDisplayName`.

The pair dialog already tracks whether the receiver-label field was edited; it now sends `receiver_label_auto: !_receiverLabelTouched`, and the new field rides `PairingSummary` from the companion backend. A pairing persisted before the flag reads `false` so its label wins, matching how the peer side handled its own rollout. Editing an endpoint no longer touches the flag, so the friendly name keeps showing.

- fixes https://github.com/esphome/device-builder-frontend/issues/1254
- companion backend PR: esphome/device-builder#2104

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/api/esphome-api.ts
+++ b/src/api/esphome-api.ts
@@ -2079,6 +2079,7 @@ export class ESPHomeAPI {
     offloader_label: string;
     pairing_key?: string;
     offloader_label_auto?: boolean;
+    receiver_label_auto?: boolean;
   }): Promise<PairingSummary> {
     return this.sendCommand<PairingSummary>("remote_build/request_pair", args);
   }
--- a/src/api/types/remote-build.ts
+++ b/src/api/types/remote-build.ts
@@ -207,8 +207,8 @@ export interface PairingSummary {
    * from the session handshake, refreshed on every session-open
    * (non-empty values only, so an older receiver can't clobber a
    * captured name). Empty until a new-enough receiver connects.
-   * `pairingDisplayName` prefers it over an auto-derived
-   * hostname label.
+   * `pairingDisplayName` prefers it over `label` when
+   * `receiver_label_auto` is set.
    */
   friendly_name: string;
   /** Receiver's HA add-on flag from the session handshake. */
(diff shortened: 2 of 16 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

PR #1267, opened 2026-07-16
Add a write-in-English note to the crash report and feedback dialogs

### LINKED ISSUE #2108: [Bug] Note regarding the use of English in crash_report texts
### What happened?

The new crash_report translation keys look promising, but wouldn't it make sense to point out that the information must be entered in English? Since the Device Builder interface will be translated for the user, this could lead to issues in any language. And because Device Builder hasn't yet been translated into many languages, many non-English-speaking users are currently still using the English interface. That's why the note would still be useful even in that case.

Same for the "Found a bug? Have an idea?" dialog.

### Dashboard version

1.6.2b1

### How did you install it?

Desktop app

### Where is it running?

Local — Windows

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

The crash report form and the "Found a bug? Have an idea?" dialog both feed free text into a GitHub issue that the maintainers read in English; the UI is localizable, so a non-English user can easily type their own language there. This adds a short "please write in English" note under the crash report's description field and on the feedback dialog's "Report a new issue" screen. It shows unconditionally, since many non-English speakers still run the English UI while their language is untranslated.

- fixes https://github.com/esphome/device-builder/issues/2108

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/crash-report-dialog.ts
+++ b/src/components/crash-report-dialog.ts
@@ -410,11 +410,14 @@ export class ESPHomeCrashReportDialog extends LitElement {
         id="crash-description"
         class="describe-input"
         rows="3"
+        aria-describedby="crash-description-note"
         placeholder=${this._localize("crash_report.describe_placeholder")}
         .value=${this._userDescription}
         @input=${this._onDescriptionInput}
       ></textarea>
-      <p class="describe-note">${this._localize("crash_report.describe_english")}</p>
+      <p id="crash-description-note" class="describe-note">
+        ${this._localize("crash_report.describe_english")}
+      </p>
       <ul class="summary">
         ${this._renderSummaryRow(
           this._localize(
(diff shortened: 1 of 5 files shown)


bdraco:
@esphbot review

esphbot (approved):
No blocking issues found.

bdraco:
Thanks, all three nits addressed in 7027fdae:

1. The crash-report note now has `id="crash-description-note"` and the textarea points at it with `aria-describedby`, so a screen-reader user who tabs straight into the field hears the write-in-English instruction; that is exactly the user this is for.
2. The feedback-dialog test now drives navigation by clicking the drill row (and the back button) instead of assigning `_screen`, so it covers the row wiring alongside the note.
3. Updated the `feedback-dialog.test.ts` docblock to describe both the version-prefill routing and the note.

@esphbot review

esphbot (approved):
No blocking issues found.

Issue #1271, opened 2026-07-16
Give the log buffer an owner; logs-dialog.ts is 802 lines


bdraco:
`src/components/logs-dialog.ts` is **802 lines** against the README's 500-600 hard limit. The grandfather clause covers files that pre-date the rule and asks that they not be made worse; #1261 added roughly 90 lines to it, which is what prompted this.

Line count is the symptom. The cause is that **the log buffer has no owner**: `_lines`, the `MAX_LOG_LINES` cap, and the mapping from "position N in the stream" to "index in `_lines`" are ad-hoc state on a Lit element that is also doing session lifecycle, toolbar, serial reconnect, and dialog rendering.

### What is actually smeared

| state | job |
|---|---|
| `_lines`, `MAX_LOG_LINES` | the buffer and its cap |
| `_streamIndex` | absolute position of the next line to arrive |
| `_indexShift` | how far `_lines` has moved under it: the cap drops from the front, decoding splices lines in |
| `_bufferEpoch` | invalidates work in flight against a buffer that has since been reset |

spread across `_appendCapped`, `_setLinesCapped`, `_replaceRegion`, `_resetCrashDecode`, `_beginSession`, and `_clearLogs`.

### Why it matters, concretely

- **The code does not trust its own arithmetic.** `_replaceRegion` computes `startIndex + _indexShift` and then re-verifies the result by comparing every line against the buffer before touching it. That check is worth keeping as a guard against churn, but needing it at all is the tell that the mapping has no home.
- **The invariant is hand-maintained across call sites.** `_lines` is reset in two places, and each has to remember to also reset the pending batch and the decode bookkeeping. `_resetPendingLines`'s own comment says "Paired with every `_lines = []` reset", which is a comment doing a type's job.
- **Every future buffer mutation inherits the tax.** Search, filter, fold-repeats, or a second consumer of positions all have to keep two counters in sync or silently mis-place lines.
- **Decode logic is only reachable through the element.** `test/components/logs-dialog-decode.test.ts` pokes `(el as any)._appendCapped`, `._lines`, `._decodeChain` because there is no smaller unit to test.

### Shape of the fix

A `LogBuffer` (plain class, or a Lit ReactiveController if it should drive updates directly) owning:

- `append(lines)` / `clear()` / `reset()`, with the cap enforced in one place
- `indexOf(streamPosition)`, so "where did line N land?" is a method on the thing that owns the invariant
[...]

PR #1276, opened 2026-07-17
Adopt LogBuffer in the command and install dialogs

bdraco:

Second of three for #1271. Stacked on #1275, which adds `LogBuffer`; **review that one first**, and this diff is only the adoption once it lands.

`command-dialog` and `firmware-install-dialog` both hand-rolled the shape `logs-dialog` had: a lines array, a `LineBatcher` writing back into it, and a reset that had to remember to clear both. Neither caps its buffer, which is why `LogBuffer` takes `maxLines` as an option rather than baking one in; both construct it without one.

`_flushPendingLines`, `_resetPendingLines`, `_enqueueLogLine` and `_flushLogLines` were one-line delegates to the batcher, so their callers now say `_log.flush()` / `_log.reset()` / `_log.enqueue()` and the delegates are gone. Uniformly:

| Before | After |
|---|---|
| `host._lines` | `host._log.lines` |
| `host._lines = []` + `host._resetPendingLines()` | `host._log.reset()` |
| `host._flushPendingLines()` | `host._log.flush()` |
| `host._enqueueLine(line)` | `host._log.enqueue(line)` |
| `host._lines = [...host._lines, detail]` | `host._log.append([detail])` |

`command-dialog._enqueueLine` stays a method, because it also does `_timer.noteLine(line)`.

- part of https://github.com/esphome/device-builder-frontend/issues/1271

## Notes for review

- **One asymmetry worth a look.** The install dialog's `_retry` cleared `_logLines` without resetting the batcher, unlike its other reset at `_init`, so a pending line from the failed run could have flushed into the buffer the comment says it is clearing. `_log.reset()` does both. In practice nothing changes: `_retry` is only reachable from the error screen and `_fail` flushes on the way there, so pending is already empty. Flagging it because it is the one place the mapping is not purely mechanical.
- Test fakes: the flow harnesses stubbed the log sink by hand ("synchronous stand-ins for the dialog's rAF-batched log sink"). They now use `fakeLogBuffer()` in `test/_fake-host.ts`, which is a **real** `LogBuffer` with only `enqueue` short-circuited to land synchronously, so the fakes exercise real lines/append/flush/reset semantics instead of a reimplementation. The batching itself keeps its own tests.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/command-dialog.ts
+++ b/src/components/command-dialog.ts
@@ -37,7 +37,7 @@ import { espHomeStyles } from "../styles/shared.js";
 import { initialDarkMode } from "../util/dark-mode.js";
 import { configurationStem, downloadAnsiText } from "../util/download-text.js";
 import { LightDismissController } from "../util/light-dismiss-controller.js";
-import { LineBatcher } from "../util/line-batcher.js";
+import { LogBuffer } from "../util/log-buffer.js";
 import { dispatchShowLogsAfterInstall } from "../util/post-install-logs.js";
 import { registerMdiIcons } from "../util/register-icons.js";
 import { RunTimerController } from "../util/run-timer-controller.js";
@@ -143,14 +143,12 @@ export class ESPHomeCommandDialog extends LitElement {
   @state() _open = false;
   @state() _commandType: CommandType = "validate";
   @state() _state: CommandState | null = null;
-  @state() _lines: string[] = [];
   @state() _statusMessage = "";
 
-  // rAF batch buffer for streamed output — coalesce per-line writes
-  // into one render per frame instead of one per line (#348).
-  private _lineBatch = new LineBatcher((batch) => {
-    this._lines = [...this._lines, ...batch];
-  });
+  // The streamed output, batched into one render per frame instead of one
+  // per line (#348). Uncapped: a compile log is finite and users scroll back
+  // through it.
+  _log = new LogBuffer(this);
 
   // Distinguishes user-stopped from backend-failed. Both flip _state to "error"
   // but only real failures get the reset-build-env hint.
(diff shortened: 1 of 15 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Nit taken, in `690a4048`. You're right that it was half-converted, and it was my own inconsistency: I routed the install-dialog fakes through a real buffer and left this one's sink stubbed, which defeats the reason for handing it a real buffer at all.

`_enqueueLine` now calls `this._log.enqueue(line)`. Confirmed it actually delivers rather than assuming: a scratch assertion after `host._enqueueLine("streamed line")` reads `['streamed line']`, where before it read `[]`.

No test changes, as you predicted; nothing asserts on line content in the command-dialog flow tests today. It just makes the two harnesses symmetric and leaves the door open to asserting streamed output there.

Thanks for checking the `_retry` reasoning independently rather than taking my word for the flush ordering.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Rebased onto #1275's latest (`a282650a`); no changes to this PR's own diff. Thanks for the approve.

Noting for your next pass: you flagged that you reviewed from `main`, which predates `LogBuffer`, so you read the diff rather than ran the suites. That is a real limitation of a stacked PR and worth being explicit about — this branch only typechecks on top of #1275. If it helps, `git fetch origin log-buffer-adopt-dialogs && git checkout FETCH_HEAD` gives you the full stack in one tree, since the branch already contains #1275's commits.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Rebased onto `main` now that #1275 has landed. It was squashed, so the branch was still carrying #1275's seven commits under their original SHAs; replayed only this PR's own two on top of the merged main (`974a37be`, `ccedf40f`). GitHub had already retargeted the base, and the diff was correct either way (it uses the merge base), but the branch history was showing work that is already in main.

Also picked up #1278, which landed alongside. Full suite green on the new base: 5521 tests, 542 files.

No changes to this PR's own content.

@esphbot review

esphbot (approved):
> [!TIP]

PR #1277, opened 2026-07-17
Split the logs-dialog session state machine out

### LINKED ISSUE #1271: Give the log buffer an owner; logs-dialog.ts is 802 lines

bdraco:

Last of three for #1271. Stacked on #1276 (which is stacked on #1275); **review those first**, and this diff is only the session split once they land.

#1275 gave the buffer an owner, which was the issue's stated cause but only worth ~35 lines. `logs-dialog.ts` was still 670 against the README's 500-600 hard limit, so the title's complaint stood. The largest thing left in the file that wasn't the element was the log source's lifecycle: every transition of `_session`, the OTA spawn, the Web Serial attach and reconnect, and the teardown.

That moves to `src/components/logs-dialog/session.ts` as free functions over the element, following the existing `command-dialog/commands.ts` precedent rather than inventing a new shape. The public API (`open`, `openPassive`, `setSerialStream`, `setSerialOpenFailed`, `abortSerialReconnect`, `close`) stays on the element as delegates, so callers and `streamSerialToDialog` are untouched. What's left is the element: its state, `render`, the log buffer, and the crash callout.

`logs-dialog.ts` 670 -> **501**, inside the cap. `session.ts` is 245.

- fixes https://github.com/esphome/device-builder-frontend/issues/1271

## Notes for review

- Members the session module reaches (`_session`, `_open`, `_api`, `_localize`, `_reconnect`, `_showStates`, `_expanded`, `_backToInstall`, `_backToInstallHandler`, `_clearLogs`, `_resetAnsiLogScroll`) lose their `private`. That is what `command-dialog` already does for the same reason; the alternative was a wide bespoke host interface, which buys nothing here.
- **Two things the tests caught that are worth knowing about**, since both would have been silent behaviour changes:
  - `_toggleShowStates` has to keep returning its promise. It was `async` and a test awaits the restart; making the delegate `void` the call broke the respawn assertion. It now returns, and I confirmed the test fails again if it doesn't.
  - `startOtaStream` is exported rather than module-private, because a test calls it directly to pin the guard that a serial session never spawns a backend stream. That guard is real and worth keeping covered.
- `render()` is still 134 lines and README:182 says >~100 signals a sub-component extraction. Left alone deliberately: the file is under the cap without it, and it is a separate concern from this issue.

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/logs-dialog.ts
+++ b/src/components/logs-dialog.ts
@@ -28,6 +28,17 @@ import { LogBuffer } from "../util/log-buffer.js";
 import { notifyError } from "../util/notify.js";
 import { registerMdiIcons } from "../util/register-icons.js";
 import { CrashDecodeController } from "./crash-decode-controller.js";
+import {
+  abortSerialReconnect,
+  onStart,
+  onStop,
+  openOta,
+  openPassive,
+  setSerialOpenFailed,
+  setSerialStream,
+  teardownSession,
+  toggleShowStates,
+} from "./logs-dialog/session.js";
 import type { ESPHomeCrashReportDialog } from "./crash-report-dialog.js";
 import { logsDialogStyles } from "./logs-dialog.styles.js";
 import {
@@ -74,14 +85,14 @@ const MAX_LOG_LINES = 5000;
 export class ESPHomeLogsDialog extends LitElement {
   @consume({ context: localizeContext, subscribe: true })
   @state()
-  private _localize: LocalizeFunc = (key) => key;
+  _localize: LocalizeFunc = (key) => key;
 
   @consume({ context: darkModeContext, subscribe: true })
   @state()
   private _darkMode = initialDarkMode();
 
   @consume({ context: apiContext })
-  private _api!: ESPHomeAPI;
+  _api!: ESPHomeAPI;
 
   @property()
   configuration = "";
@@ -92,13 +103,13 @@ export class ESPHomeLogsDialog extends LitElement {
   // The active log source + its lifecycle. Single source of truth; the toolbar
   // (streaming dot, Stop/Start, Reset enablement, source chip) all derive from
   // it. See logs-session.ts for the states and why they're a union.
-  @state() private _session: LogsSession = { kind: "idle" };
+  @state() _session: LogsSession = { kind: "idle" };
 
   @state()
-  private _expanded = false;
+  _expanded = false;
 
   @state()
-  private _showStates = true;
+  _showStates = true;
 
   /**
    * Set when this session was launched as the post-install logs
@@ -111,12 +122,12 @@ export class ESPHomeLogsDialog extends LitElement {
    * the run that asked for it.
    */
   @state()
-  private _backToInstall = false;
-  private _backToInstallHandler: (() => void) | null = null;
+  _backToInstall = false;
+  _backToInstallHandler: (() => void) | null = null;
 
   // Reconnect hook for a Web Serial session whose reader is gone (a reopen
   // failed -> `dead`); the "click Start to reconnect" recovery (#636).
-  private _reconnect: (() => Promise<void>) | null = null;
+  _reconnect: (() => Promise<void>) | null = null;
 
   // The visible log, its cap, and the stream-position map inline decoding
   // needs. Owns every line the dialog shows; the dialog holds no counters.
@@ -144,7 +155,7 @@ export class ESPHomeLogsDialog extends LitElement {
   private _crashReportDialog!: ESPHomeCrashReportDialog;
 
   @state()
-  private _open = false;
+  _open = false;
 
   @query("esphome-process-terminal")
   private _terminal?: ESPHomeProcessTerminal;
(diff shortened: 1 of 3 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
All three taken, code in `0553971a`.

**Module docblock** — moved above the imports. Same class of thing I'd just fixed in `crash-decode-controller.ts`, where the docblock sat above the *interface* rather than the class it described. Your point about a future edit silently absorbing it into `openOta`'s doc is the real risk.

**`port = OTA_PORT`** — dropped; `openOta` now requires `port`. This was the better catch of the three: the default was pure duplication of a policy the element's `open()` had already applied, in a second file where it could drift. `OTA_PORT`'s import is gone from `session.ts` too, so it's back to being one `logs-dialog.ts` concern. I checked for orphans by hand rather than trusting the typecheck, since this repo has `noUnusedLocals: false` and `tsc` is blind to dead imports; `notifyError`, `isPassive` and `isStreaming` all still have live uses.

**`_backToInstallHandler`** missing from the PR notes' de-privatized list — corrected in the body.

Also useful to have the "nothing became dead" and "no runtime import cycle" checks done independently, and confirmation that the `command-dialog/commands.ts` precedent claim actually holds. That was the load-bearing argument for the shape, so it being verified rather than accepted is what I wanted.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Rebased onto #1276's latest; this PR's own diff is unchanged at `4318b63b`.

On the branch not being fetchable: that is my doing, not your environment. I rebase the whole stack whenever a PR below changes, so `0553971a` was rewritten away before you looked for it, and your pass landed on `a3055b6` (the pre-rebase head). The branch is on origin now:

```
4318b63b  refs/heads/log-buffer-session-split
```

The three fixes are in it — module docblock above the imports, `openOta` requiring `port` with `OTA_PORT` gone from `session.ts` entirely, and `_backToInstallHandler` added to the PR notes. Worth a re-fetch if you want to confirm rather than take them on my description; you were right to flag that you had not.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Rebased onto #1276's new head, which is itself now rebased onto the merged `main`. This PR's own two commits are unchanged in content (`bb011fa8`, `53bd124b`); `logs-dialog.ts` is still 501 lines.

Full suite green on the new base: 5521 tests, 542 files.

Since you flagged last time that you could not fetch the branch: the head is `53bd124b` and it is on origin now. Same caveat as before, though, and it is worth stating plainly rather than you rediscovering it — every merge below this PR rewrites these SHAs, so anything I cite here is only valid until the next one.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Rebased onto `main` now that #1276 has landed. Same squash-duplication cleanup as last round: the branch was still carrying #1276's two commits under their pre-squash SHAs, so I replayed only this PR's own two on top of the merged main. GitHub had already retargeted the base to `main`.

Head is now `ebaf8fa9`. Content unchanged from the round you approved: three files, `logs-dialog.ts` 501 lines, `session.ts` 244. Full suite green on the new base (5521 tests, 542 files), typecheck clean.

This is the last of the three, so the SHAs I cite here should finally stop moving under you.

@esphbot review

esphbot (approved):
> [!TIP]

PR #911, opened 2026-06-20
Queued offline updates

rwalker777:
# What does this implement/fix?

Adds frontend support for queued offline updates. Installing to an offline device now compiles right away and reports "update queued" in the command dialog; the device keeps its normal Offline badge and shows a small clock next to the modified and update dots in both the card grid and the table, and the kebab menu gains a "Clear queued update" action that goes through the shared confirm dialog.

Companion backend PR: esphome/device-builder#1430

**Related issue or feature (if applicable):**

- fixes <link to issue>

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [x] New feature (non-breaking change which adds functionality) — `new-feature`
- [ ] Enhancement to an existing feature — `enhancement`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [ ] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `docs`
- [ ] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Checklist

- [ ] The code change is tested and works locally.
- [x] `npm run lint` passes.
- [x] `npm run test` passes.
- [x] Tests have been added to verify that the new code works (where applicable).

PR #1282, opened 2026-07-18
Guide first-time installs to USB instead of a dead-end queued OTA

bdraco:

Fixes an unintended consequence of the offline OTA defer feature (esphome/device-builder#1430, esphome/device-builder#1964, #911) that we didn't think about when accepting it: it also catches first time installs. Before the feature, an OTA install to a brand new never-flashed device failed loudly and the user went looking for the cable. Now the same install compiles, reports "Update queued. It will install when the device comes back online." and waits forever; a board with no firmware can never come online by itself. First time users are never told a USB cable is required for the first install, and have no way to find out.

The frontend now distinguishes a never-flashed device from a previously-flashed one that is temporarily offline, using deploy evidence the backend already persists (deployed_version, deployed_config_hash, mac_address, or being online right now).

For a never-flashed device the install dialog leads with the USB rows, shows a first-install callout ("The first install needs a USB cable; you only need to do this once"), and the network row says the compile can't reach the device until its first USB install. The queued result message points at the cable instead of promising an install on a wake that will never come. The just-created welcome banner also mentions the USB first install and gains an "Install now" shortcut, routed through the page's save-then-install path.

A previously-flashed offline device keeps "On the network" first and gains an explicit warning callout explaining the compile-now, install-on-wake behavior, so queued updates for sleeping devices stay one click away.

Frontend only, intended for a patch release. Notes for review: the never-flashed heuristic errs conservative (an mDNS-dark device with no API could read as never-flashed while not online; the only consequence is row ordering plus honest copy). The backend queue stays armed either way; after the first USB flash the queued OTA delivers the same firmware once, which is harmless.

- unintended consequence of the offline OTA defer feature (esphome/device-builder#1430 and esphome/device-builder#1964 with device-builder-frontend#911); no tracking issue filed

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

Issue #1298, opened 2026-07-18
Consolidate copy-pasted test scaffolding (webawesome mocks, storage stubs, MockObserver)


bdraco:
Several test scaffolding pieces are copy pasted across the suite:

1. Web Awesome component stubs; 248 identical `vi.mock("@home-assistant/webawesome/dist/components/X/X.js", () => ({}))` lines across 130 test files (icon 111, dialog 42, spinner 31, select 15, option 15, plus callout/badge/switch/tooltip). Sixteen files stub the exact dialog+icon+spinner trio.
2. A Map backed Storage stub with an identical 4 method body in 6 files: `test/util/split-ratio.test.ts`, `test/util/auth-token.test.ts`, `test/util/pending-highlight.test.ts`, `test/util/dashboard-filters-session.test.ts`, `test/util/just-created.test.ts`, `test/api/esphome-api.test.ts`; several also hand roll a matching throwing variant.
3. `MockObserver` (IntersectionObserver mock, 27 lines) byte identical in `test/util/intersection-controller.test.ts` and `test/components/wizard-step-board-list.test.ts`.
4. About 8 files re declare `identityLocalize` and 2 re declare `flush` even though `test/_dom.ts` already exports both; 3 controller tests declare local `stubHost` variants beside the existing `test/_fake-host.ts`.

## Plan

Split into small PRs, roughly one per numbered item; each is independent.

1. Web Awesome stubs: spike first, since `vi.mock` is hoisted per file with a static path. Preferred shape is a `test/_mock-webawesome.ts` whose top level calls `vi.mock` for the common components, imported for side effect (`import "../_mock-webawesome.js"`); verify vitest applies mocks declared in an imported module, and fall back to a `setupFiles` entry in `vitest.config.ts` with `vi.unmock` opt outs for the minority of tests that want a real component. Migrate the 16 dialog+icon+spinner trio files first, then sweep the rest mechanically.
2. Storage stub: add `test/_storage.ts` exporting `fakeStorage()` returning `{ store, stub }` parameterized by `"localStorage" | "sessionStorage"`, plus `throwingStorage()`; migrate the 6 files and their throwing variants.
3. MockObserver: add `test/_intersection-observer.ts` exporting the class unchanged (`.instances`, `.trigger` API); both suites already `vi.stubGlobal("IntersectionObserver", MockObserver)`, so the migration is import plus delete.
[...]

PR #1306, opened 2026-07-18
Share the MockObserver test mock and canonical test helpers

bdraco:

Adds `test/_intersection-observer.ts` with the `MockObserver` class that was byte identical in the intersection-controller and wizard-step-board-list suites; both now import it. Also points the true identity localize duplicates at the existing `test/_dom.ts` export (auto-apply-controller, add-component-form-seed-fns) and folds two trivial `stubHost` variants onto `test/_fake-host.ts` (enter-controller uses `FakeHost`, catalog-load-controller uses `fakeHost`).

Deliberately unchanged: board-change keeps its inline localize, its comment documents that the suite runs in node and avoids importing `_dom.ts`; component-name-resolver keeps its stubHost, which is a real variant driving connect/disconnect through the captured controller, not a copy; the other `const localize` declarations across the suite format values, so they are not duplicates. The audit's claimed `flush` re-declarations turned out to be `flushMicrotasks` wrappers, nothing to change.

Part of #1298 (items 3 and 4); the webawesome mock consolidation remains.

- part of https://github.com/esphome/device-builder-frontend/issues/1298

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/test/_intersection-observer.ts
+++ b/test/_intersection-observer.ts
@@ -1,0 +1,34 @@
+/**
+ * IntersectionObserver mock for suites that drive sentinel-based
+ * infinite scroll. Install with
+ * ``vi.stubGlobal("IntersectionObserver", MockObserver)``; reset
+ * ``MockObserver.instances`` in ``beforeEach`` and fire crossings via
+ * ``instance.trigger(isIntersecting)``.
+ */
+export class MockObserver {
+  static instances: MockObserver[] = [];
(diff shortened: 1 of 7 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1314, opened 2026-07-19
Name the empty or commented-out block behind 'expected a dictionary.'

bdraco:

When a config has a section key whose options are all commented out, like `advanced:` with `# minimum_chip_revision: "3.1"` under it, the save dialog and the editor squiggle only said `expected a dictionary.` ESPHome strips the config path from the message before it reaches us, so the user had no idea which line was wrong or what to do about it. A real user hit exactly this after commenting out `minimum_chip_revision` to work around an OTA failure, and needed Discord support to learn the fix was removing the `advanced:` line too.

`describeValueTypeCause` now recognizes a value-less `key:` with nothing (or only comments) under it and names the line, the key, and the repair. The dialog and the inline tooltip both read: `expected a dictionary. Line 8: everything under "advanced" is commented out, so it has no options. Uncomment an option, or comment out "advanced" too.` A key with nothing under it at all gets a matching wording that suggests adding an option or removing the line. The hint is gated on the `expected a dictionary` message so an unrelated error anchored on a legal empty key (`logger:`) never picks up remove-the-line advice.

- fixes N/A (reported via Discord support thread)

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/translations/en.json
+++ b/src/translations/en.json
@@ -1249,6 +1249,8 @@
     "error_nested_list_hint": "The \"- \" items below \"{key}\" are indented as its value. If they should be separate list entries, dedent them to line up with \"- {key}\".",
     "error_indent_hint": "YAML indentation problem near line {line}. Make sure the line and its neighbours are indented consistently with their block; an over-indented or under-indented line, or a stray \":\" in a value, causes this.",
     "error_missing_colon_hint": "Line {line}: \"{key}\" looks like an option name without a value. Add \":\" and a value after it, or remove the line.",
+    "error_commented_block_hint": "Line {line}: everything under \"{key}\" is commented out, so it has no options. Uncomment an option, or comment out \"{key}\" too.",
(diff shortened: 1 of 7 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Suggestion 1 applied in 7db871c1, the disjunct folds to lineKeyToken alone since the marker capture is non empty whenever it matches.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Applied the simplify pass in 1e902593, the message argument is now required so a future call site cannot silently drop the gated hints, and the duplicated unit test fixture is trimmed to the minimal shape.

@esphbot review

esphbot (approved):
> [!TIP]

PR #1318, opened 2026-07-20
Preselect the trigger when only one is offerable in the add-automation dialog

### LINKED ISSUE #2214: New-automation dialog makes the user hand-pick the only available trigger
### What happened?

Adding an automation from the Template Button section opens the "New automation" dialog with an empty "Which trigger?" dropdown, even when exactly one trigger exists for the target (Button → On Press). Getting past it takes three interactions (open dropdown, pick the only entry, Continue) where zero choices are actually being made. This is the very first step of the very first automation most users build, so the friction lands exactly where first impressions form.

### Suggested improvement

Preselect the trigger when the filtered list has exactly one entry and keep Continue enabled, so the dialog is a single Continue click (with the dropdown still editable). Preselecting only on a unique match keeps the multi trigger case unchanged.

### Dashboard version

dev (explicit-id-signal, 2026.7.0)

### How did you install it?

Dev checkout

### Where is it running?

macOS

bdraco:

When the filtered trigger list for the chosen target has exactly one entry, the New automation dialog now preselects it, so the dialog is a single Continue click. The dropdown stays editable and the trigger description renders immediately; the multi trigger case is unchanged. The preselect runs at every point the selection is reset (load, kind change, component change), so it also applies when only one trigger remains after already used ones are filtered out.

- fixes https://github.com/esphome/device-builder/issues/2214

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/add-automation-dialog.ts
+++ b/src/components/device/add-automation-dialog.ts
@@ -164,6 +164,7 @@ export class ESPHomeAddAutomationDialog extends LitElement {
         this._componentId =
           this._available.devices.find((d) => d.parent_id === container.id)?.id ?? "";
       }
+      this._preselectUniqueTrigger();
     } catch (err) {
       this._error = getErrorMessage(err);
     } finally {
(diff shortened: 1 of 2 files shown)


bdraco:
@esphbot review

PR #1319, opened 2026-07-20
Offer an indent auto-fix when a stray top-level key belongs to the section above

### LINKED ISSUE #2215: bad key at top level could be an auto fix since we know its not a top level component
in this case its an `id` we could suggest indenting and offer an autofix. but anything that is not a valid component but is a valid key under the parent should have a better error message + autofix offer

Image

bdraco:

A stray top level key (`id: mylogger` at column 0 below a `logger:` block) only got esphome's raw "Component not found: id." squiggle and banner. We can do better: when the key is not a component but the component catalog confirms it is a valid option of the section above, the error now carries a plain language hint ("Indent line 13 by 2 spaces to move it inside \"logger\".") and a one click Try auto-fix that indents the line.

The new `describeComponentNotFoundFix` analyzer mirrors the merged invalid option fix: message regex, buffer shape checks (column 0 key line, a valueless section opener above, map shaped body), an AST agreement check, then the catalog gate, failing closed at every step so a genuinely unknown component gets no fix rather than a wrong one. The indent delta follows the section's own child indent. No new fix kind; the kind-less positive indent path in `applyAutoFix` (staleness guards, validate before apply, undo) delivers it unchanged, as does the banner button.

- fixes https://github.com/esphome/device-builder/issues/2215

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/translations/en.json
+++ b/src/translations/en.json
@@ -1263,6 +1263,7 @@
     "error_go_to_line": "Go to line {line}",
     "error_dash_space_fix": "Missing space: line {line} \"-{key}\" is a list item missing the space after its \"-\". Add the space so it reads \"- {key}\".",
     "error_nest_under_fix": "Indentation mismatch: line {line} \"{key}\" is not an option of the block it sits in, but it is a valid option of \"{target}\". Indent line {line} by {spaces, plural, one {# space} other {# spaces}} to nest it under \"{target}\".",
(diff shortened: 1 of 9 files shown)


bdraco:
@esphbot review

bdraco:
@esphbot review

PR #1324, opened 2026-07-20
Treat a bare platform-domain section as empty, not external

### LINKED ISSUE #2218: Empty platform-domain section renders as an external component
### What happened?

An empty platform domain section (`switch:` with nothing under it yet) renders in the structured pane as "External component / switch / This section doesn't have a structured editor" with a Delete section button. `switch` is not an external component; it is a platform domain waiting for its first `- platform:` entry. The pane decides "unknown" purely by a catalog lookup of the bare key, and the catalog only carries dotted ids (`switch.gpio`), so every bare domain header lands in the external component branch.

### Suggested improvement

When the bare key is a known platform domain (catalog carries dotted ids under it), render an empty platform section state instead: name the domain, explain that entries under it each pick a platform, and offer an Add button that deep links into the add component dialog filtered to that domain (the `request-add-component` flow the id reference pickers already use). Genuinely unknown keys keep the external component treatment.

### Dashboard version

dev (2026.7.0)

### How did you install it?

Dev checkout

### Where is it running?

macOS

bdraco:

An empty platform domain section (`switch:` with nothing under it yet) rendered as "External component / switch / This section doesn't have a structured editor" with a red Delete section button. The pane decides unknown by looking the bare key up in the catalog, and the catalog only carries dotted ids (`switch.gpio`), so every bare domain header fell into the external branch.

The miss branch now also checks the catalog's platform domains (derived from categories and dotted id prefixes, the same source the top level completions use). A bare domain renders as "Platform section" with an explainer and an Add button that deep links the add component dialog filtered to that domain, through the same `request-add-component` event the id reference pickers use. Delete stays available, and a genuinely unknown key keeps the external component treatment.

- fixes https://github.com/esphome/device-builder/issues/2218

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Copilot's nit fixed in 8ef41e53, platformDomains now returns a ReadonlySet so the memoized instance can't be mutated by a caller. Kōan's file size suggestion is real but pre-existing; tracking the split as a follow up issue so this stays a focused bugfix.

@esphbot review

PR #1327, opened 2026-07-20
Simplify the device-section-config split seams

bdraco:

Follow up cleanup pass over the #1326 split, from a four angle review (reuse, simplification, efficiency, altitude) plus the Copilot comments on that PR. The three one line `_renderXxxBranch` delegates are gone; `render()` calls the branch functions directly, matching how it already called the header and dialogs. The opts objects dissolved into plain signatures: `renderStructuredFormBranch` derives `renderEntries` itself (the memoized `resolveSectionEntries` makes the repeat free and restores the substitutions override test pin to one file), and `render-header.ts` derives `catalogMiss` / `headerTitle` where they are consumed. The notice shell and delete row moved into `render-branches.ts`, their only consumer, so `render-tables.ts` is purely the three manage lists; the api actions table now routes through a new `selectApiActionRows` in `automation-rows.ts` like its two siblings, with a unit test. Five hand rolled bubbling `CustomEvent` dispatches collapsed to the existing `fireEvent` helper, and the stranded `_apiListFlashKey` field moved next to the other non reactive fields.

One deliberate behavior tweak, from the Copilot review on #1326: the docs link is now gated through `isSafeLinkHref` and carries `rel="noopener noreferrer"`, consistent with the other safe link call sites; a catalog `docs_url` without an http(s)/mailto scheme renders no link instead of a live anchor. Everything else is identical, verified in the live dashboard against all three render branches and the dialogs. Main file is 524 lines.

- follow up to #1326 (issue device-builder-frontend#1325)

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/device/device-section-config.ts
+++ b/src/components/device/device-section-config.ts
@@ -62,7 +62,6 @@ import {
   renderPlatformDomainBranch,
   renderStructuredFormBranch,
   renderYamlOnlyBranch,
-  type StructuredFormOpts,
 } from "./device-section-config/render-branches.js";
 import {
   renderAddAutomationDialog,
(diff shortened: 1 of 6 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Fixed the comment in 71794b01: it now states the actual identity guarantees (constant or catalog array for map/default sections, fresh allocation for list sections) instead of claiming memoization.

@esphbot review

esphbot (approved):
> [!TIP]

PR #1333, opened 2026-07-20
Fetch ESPHome Web deploy translations in the lokalise environment

jesserockz:

The `deploy-web.yml` build job read the Lokalise secrets directly, but they live in the `lokalise` GitHub environment, so they resolved empty on every run — the "Lokalise secrets unset; deploying English-only" guard fired and web.esphome.io shipped English-only.

This mirrors the pattern already in `release.yml`: a dedicated `translations` job holds the credentials (`environment: lokalise`), downloads the locales, and hands them to the build job as the `translations` artifact, so the bundler job never sees the API token. `build:web` regenerates the language manifest from the restored `src/translations/`, so every locale ships.

It also tightens the failure mode: on the canonical repo, missing secrets now hard-fail (`::error::` + `exit 1`) instead of silently degrading the live site to English-only; forks still warn and build English-only.

Workflow-only change — no source or backend changes, and the diff vs `main` is just the job split (the Pages action pins bumped by #1321/#1322 are preserved).

- N/A

## Types of changes

- [x] CI / workflow change — `ci`

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/.github/workflows/deploy-web.yml
+++ b/.github/workflows/deploy-web.yml
@@ -11,8 +11,8 @@ on:
     branches: [main]
   workflow_dispatch:
 
-# Build job just reads the repo and uploads the artifact; the deploy job below
-# gets the Pages-write + OIDC scope.
+# The translations + build jobs only read the repo (contents: read, inherited
+# below); the deploy job alone gets the Pages-write + OIDC scope.
 permissions:
   contents: read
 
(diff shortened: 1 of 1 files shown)


esphbot (approved):
> [!TIP]

PR #1335, opened 2026-07-21
Reveal the fresh clone: clear the search and highlight the new device

### LINKED ISSUE #2246: Cloning a device takes you back to the search page with the device you searched for originally to clone selected instead of the new device
We should clear the search and take them to the newly clone device with the highlight

bdraco:

After a clone, the dashboard kept the search that matched the source device, so the user landed back on the filtered list with the original device instead of the new one. The clone response already carries the new configuration filename; `executeClone` was discarding it.

Now a successful clone hands the returned configuration to a new `_onCloned` on the dashboard page: it clears the search (which by definition matched the source, not the clone), leaves facet filters alone (parity with the adopt flow), and arms the same highlight plus deferred scroll adopt uses. The scroll is also tried immediately because the rescan's ADDED push can land before the command reply. The search box is deliberately not refocused; attention belongs on the new card. Failures keep the old behavior, the search stays put.

Verified live in both card and table view: search cleared, list scrolled to the clone, 4 second highlight ring shown.

- fixes https://github.com/esphome/device-builder/issues/2246

## Types of changes

## Checklist

- [x] `npm run lint` passes.
- [x] `npm run test` passes.

### DIFF
--- a/src/components/dashboard/actions-ui.ts
+++ b/src/components/dashboard/actions-ui.ts
@@ -60,9 +60,10 @@ export async function executeClone(
   const device = host._actionDevice;
   if (!device) return;
   const { newName, newFriendlyName } = e.detail;
+  let result: Awaited<ReturnType<ESPHomeAPI["cloneDevice"]>>;
   try {
     const friendly = newFriendlyName.length > 0 ? newFriendlyName : undefined;
-    await host._api.cloneDevice(device.configuration, newName, friendly);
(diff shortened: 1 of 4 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1337, opened 2026-07-21
Ask desktop installs how they'll be used before the experience step

stvncode:

Implements the desktop-app onboarding changes agreed on the call: under the ESPHome Desktop app (detected via the handshake's `desktop_version`, the repo-wide "is desktop" signal), the wizard asks **"How do you want to use this installation?"** right after Welcome, before the experience-level question.

- **Standalone installation** — continues the flow as-is (experience level, tour offer). The old existing-server orientation step is subsumed by this question, so it no longer shows on desktop; non-desktop installs keep it unchanged.
- **Use as remote builder** — onboarding ends on the spot: the wizard persists `remote_compute_only: true` + `hide_device_builder: true` and closes. The dashboard reacts to the preference flip (no reload) and shows only the remote-building screens, with the pairing wizard first — verified live end-to-end in the dev server.

Detection-driven defaults, per the spec:

- No existing installation found on the network → **Standalone** preselected with the RECOMMENDED badge.
- Existing installation found (mDNS-discovered Device Builder) → **Use as remote builder** preselected with the badge.
- The default + badge are pinned when the screen is entered (same philosophy as the existing-server freeze) so a host arriving mid-choice can't flip them under the user.
- Picking **Standalone while an installation was detected** shows a warning banner naming the found install ("…creates a second, separate Device Builder — continue only if that's what you want"). The banner reads the live host map on purpose, so a host discovered mid-screen still raises the flag.

A standalone pick also writes `hide_device_builder: false`, so re-running onboarding (version bump) can back out of a previous remote-builder setup. `hide_device_builder` is only ever written when the usage question was shown.

New copy lives under `onboarding.wizard.usage.*` in `en.json` only, per the translations policy.

No backend changes needed — the flow uses the existing `config/set_preferences` fields (`remote_compute_only`, `hide_device_builder`) and the existing remote-build dashboard mode.

- N/A (spec from maintainer call)

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/onboarding/choice-card-styles.ts
+++ b/src/components/onboarding/choice-card-styles.ts
@@ -77,6 +77,12 @@ export const choiceCardStyles = css`
     flex-shrink: 0;
   }
 
+  /* Image variant (brand logo) sized to match the mdi glyph box. */
+  .choice-image {
+    width: 24px;
+    height: 24px;
+  }
+
   .choice-text {
     display: flex;
     flex-direction: column;
--- a/src/components/onboarding/choice-card.ts
+++ b/src/components/onboarding/choice-card.ts
@@ -1,8 +1,12 @@
 import { type TemplateResult, html, nothing } from "lit";
 
 export interface ChoiceCardProps {
-  /** mdi icon name; the caller is responsible for registering it. */
-  icon: string;
+  /** mdi icon name; the caller is responsible for registering it.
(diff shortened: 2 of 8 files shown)


esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Remote-builder path silently overwrites the saved experience level

PR #1338, opened 2026-07-21
Run the leave guard before Escape pops history in the device editor

### LINKED ISSUE #2259: [Bug] Escape key bypasses prompt for "Save changes" etc.
### What happened?

1. Make changes to an existing device's file without saving them.
2. Use the mouse to click on a dark area around the outside edges inside of the ESPHome ingress (your cursor cannot be in the editor textarea).
3. Hit the escape key -- and find that the ESPHome editor didn't prompt for saving the unsaved changes.

When hitting the back arrow button in the top left of the ESPHome ingress if there are changes to the file it prompts whether to "Cancel", "Save", or "Discard" the unsaved changes. I would hope that hitting the escape key would offer the same prompt. :)

I just spent about 20 minutes making changes without saving them and tried to exit the search/replace box by hitting escape and this bug trashed me. I thought I'd better check to see if anyone had reported it because I'm sure it'll find other devs and destroy their work, as well. 👎 

### Dashboard version

2026.7.1

### How did you install it?

Home Assistant App

### Where is it running?

Home Assistant App

### Relevant logs / errors

```shell
None
```

### Anything else?

_No response_

bdraco:

Pressing Escape in the device editor with focus outside the YAML surface called raw `window.history.back()`, relying on the page's popstate listener to intercept and show the unsaved changes prompt. That listener never wins; popstate handlers fire in registration order, so the router (registered first) commits the navigation and unmounts the page before the veto runs, discarding the dirty buffer with no prompt. The header back arrow already handles this by running the leave guard before popping history.

This extracts that guarded back navigation into a shared `goBackOrHome()` in `util/navigation.ts` and routes both the header back arrow and the Escape handler through it. Escape now shows the same Cancel / Discard / Save prompt as the back button, and Escape on a deep linked editor (null `history.state`) falls back to `navigate("/")` instead of leaving the site. Verified in the live dashboard: dirty buffer plus Escape opens the prompt, Cancel keeps the buffer, Discard leaves, and Escape inside the CodeMirror search box still just closes the search panel.

Also checked the secrets page per the issue triage: it is not affected. It registers no Escape handler at all, so there is no bypass path; its back arrow and in-app navigation already run the guard (re-verified in the browser after this refactor).

- fixes https://github.com/esphome/device-builder/issues/2259

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/esphome-layout.ts
+++ b/src/components/esphome-layout.ts
@@ -15,7 +15,7 @@ import { espHomeStyles } from "../styles/shared.js";
 import { textStyles } from "../styles/text.js";
 import { stripBase, withBase } from "../util/base-path.js";
 import { deviceBuilderChannel } from "../util/device-builder-channel.js";
-import { navigate, runLeaveGuard } from "../util/navigation.js";
+import { goBackOrHome } from "../util/navigation.js";
 import { registerMdiIcons } from "../util/register-icons.js";
 import {
   desktopDocsUrl,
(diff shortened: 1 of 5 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1342, opened 2026-07-21
Show substitution references in long-form pin fields

bdraco:
# What does this implement/fix?

With the advanced pin schema (`pin: { number: ${testpin1}, inverted: true }`) the visual editor's pin field rendered blank. A scalar `pin: ${testpin1}` already works because the form's leaf gate edits any `${var}` scalar as text, but the long form is an object, so it lands in the pin renderer, where a `${var}` in `number` parses to no board GPIO and the picker selects nothing.

The picker cannot represent a substitution reference, and resolving it into the dropdown would let one pick clobber the reference with a literal. So the pin renderer now catches a long-form pin whose `number` looks like a substitution and edits `number` as text, with the same resolves-to hint the scalar path shows (previewing `GPIO10` from the file's `substitutions:`, or the resolved-at-build-time marker for package defined ones). Edits route to `pin.number` so `mode` and `inverted` are preserved, and the Advanced disclosure renders unchanged. Verified in the live editor with the issue's exact config: both `binary_sensor.gpio` and `output.gpio` now show the reference plus its resolved pin, and a long-form pin with a literal `number` still gets the board picker.

**Related issue or feature (if applicable):**

- fixes https://github.com/esphome/device-builder/issues/2268

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [ ] New feature (non-breaking change which adds functionality) — `new-feature`
- [ ] Enhancement to an existing feature — `enhancement`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [ ] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `docs`
- [ ] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Checklist

- [x] The code change is tested and works locally.
- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.
- [x] Tests have been added to verify that the new code works (where applicable).

PR #1344, opened 2026-07-21
Edit a pin mode substitution shorthand as text

### LINKED ISSUE #1343: Pin mode substitution shorthand renders read-only in the visual editor
A long-form pin's `mode` written as a scalar substitution shorthand renders read-only in the visual editor:

```yaml
substitutions:
  my_mode: INPUT_PULLUP

binary_sensor:
  - platform: gpio
    pin:
      number: GPIO5
      mode: ${my_mode}
```

The Advanced disclosure shows "value set in YAML" for `mode` instead of the editable text-with-preview treatment `pin.number` now gets (#1342, backend issue esphome/device-builder#2268).

Cause: `renderLongFormChild` in `src/components/device/config-entry-pin-renderer.ts` intercepts a string `mode` before `ctx.renderEntry`, so the form's scalar `${var}` leaf gate (`config-entry-form.ts`, the one that renders substitution scalars as text with the resolves-to hint) never runs; `expandPinModeShorthand("${my_mode}")` returns null and it falls through to `renderNestedField`'s scalar-at-NESTED read-only shell.

This is safe today (no clobber, no blank widget) but inconsistent with the `number` treatment. Fix would be to route a string `mode` matching `looksLikeSubstitution` through `ctx.renderEntry` so the leaf gate gives text plus hint.

Found during review of #1342; out of scope there.

bdraco:

A long-form pin's scalar `mode: ${my_mode}` rendered as the read-only "value set in YAML" shell in the visual editor. The shorthand expansion wrapper can't expand a substitution reference (`expandPinModeShorthand` returns null), so it fell through to the nested renderer's scalar notice, inconsistent with the editable text treatment `pin.number` got in #1342.

`renderLongFormChild` now excludes substitution looking strings from the shorthand branch, so they take the normal `renderEntry` dispatch where the form's scalar `${var}` gate already renders a text input with the resolves-to hint. Literal shorthands (`mode: OUTPUT`) keep the flag checkbox expansion, unknown ones (`mode: BOGUS`) keep the read-only fallback, and the object form is unchanged. Verified in the live editor: `mode: ${my_mode}` shows an editable text input with the resolved `INPUT_PULLUP` preview, and a literal `mode: INPUT_PULLUP` still shows the Input and Pullup checkboxes ticked.

- fixes https://github.com/esphome/device-builder-frontend/issues/1343

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/config-entry-form.ts
+++ b/src/components/device/config-entry-form.ts
@@ -52,7 +52,7 @@ import {
 import { registerMdiIcons } from "../../util/register-icons.js";
 import { nearestScrollContainer } from "../../util/scroll-container.js";
 import { SessionBlobCacheController } from "../../util/session-blob-cache-controller.js";
-import { looksLikeSubstitution, parseSubstitutions } from "../../util/substitutions.js";
+import { isSubstitutionString, parseSubstitutions } from "../../util/substitutions.js";
 import {
   _isStructuralType,
   filterRenderable,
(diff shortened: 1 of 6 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1347, opened 2026-07-22
Round-trip substitution references in numeric list rows

### LINKED ISSUE #1346: Numeric list rows clobber a substitution reference on first edit
A `${var}` item in a numeric multi-value list clobbers the reference on first edit:

```yaml
substitutions:
  ch: "3"

some_component:
  channels:
    - ${ch}
    - 5
```

In the visual editor, `renderMultiValueField`'s numeric rows (`src/components/device/config-entry-renderers/lists.ts`, around lines 114-152) feed each INTEGER/FLOAT item into `<input type="number">` via `.value=${item}`. A `${var}` string blanks the number input, and `updateAt` writes `Number(value)` back, so the first keystroke replaces the token with a bare number (or NaN churn).

Text-typed rows are fine and already render `renderSubstitutionHint` per item. The numeric branch needs the same per-item carve-out the scalar paths have (`config-entry-form.ts` leaf gate, `primitives-numeric.ts`, pin `number`/`mode` in #1342/#1344): an item matching `isSubstitutionString` (added in #1344, `src/util/substitutions.ts`) should render as a text row with the resolves-to hint and round-trip the string. The form's scalar gate can't help here because it checks the value at the array path, not per item.

Found during the #1344 review sweep; out of scope there.

bdraco:

A `${var}` item in an INTEGER or FLOAT multi value list fed `<input type="number">`, which blanks a non numeric value, and the edit handler wrote `Number(value)` back, so the first keystroke clobbered the reference. Text typed lists already handled substitutions per row.

The numeric decision is now per row: an item that looks like a substitution renders as a text input and round trips the string verbatim on edit, no coercion; the per row resolves-to hint the list renderer already carries starts showing for it. Sibling literal rows keep the number input and numeric coercion, and dropping the `$` flips the row back to numeric on the next render, the same transient string typed number the single value `${var}` gates write. Verified in the live editor with an `lcd_pcf8574` `user_characters.data` list: the `${glyph_row}` row is editable text with the resolved preview while the literal rows stay number inputs.

Builds on the `isSubstitutionString` predicate #1344 added (now merged, so this targets `main` directly).

- fixes https://github.com/esphome/device-builder-frontend/issues/1346

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/config-entry-pin-renderer.ts
+++ b/src/components/device/config-entry-pin-renderer.ts
@@ -575,7 +575,9 @@ function renderLongFormChild(
   // A scalar shorthand (``mode: OUTPUT``) needs the display-expansion
   // wrapper; a ``${var}`` scalar goes through renderEntry so the form's
   // substitution gate edits it as text with the resolves-to hint (#1343);
-  // the object form goes through the normal nested dispatch.
+  // the object form goes through the normal nested dispatch. Deliberately
+  // ``looksLikeSubstitution``, not ``isSubstitutionString`` — the typeof
+  // guard is load-bearing (an object mode must not hit the wrapper).
   return typeof modeValue === "string" && !looksLikeSubstitution(modeValue)
     ? renderPinModeField(scoped, modePath, ctx)
     : ctx.renderEntry(scoped, modePath);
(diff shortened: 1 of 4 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1374, opened 2026-07-22
Fall through to the flag group when a nested scalar shorthand is cleared

### LINKED ISSUE #1373: Clearing a nested scalar shorthand strands the group on an empty Set to in YAML notice
Repro: give a pin a `mode: ${my_mode}` shorthand so the Mode group renders the editable text field (#1344), then delete the field's contents.

The empty emit correctly drops `mode:` from the YAML, but the form's in memory values still hold `pin.mode = ""`, and `renderNestedField`'s scalar shorthand branch (`src/components/device/config-entry-renderers/nested.ts`, around line 41) treats any string, including empty, as an unmodellable shorthand. The user sees the read only notice with a blank interpolation, "Set to  in YAML; edit it in the YAML pane on the right.", and the Mode group is stranded until a page reload re parses the YAML and the flag group editor returns.

Fix shape: exclude the empty string from the scalar shorthand branch (`raw !== ""`) so a cleared value falls through to the normal collapsible flag group, which is the recoverable editor. Found while testing #1370; unrelated to it (the branch and the emit path predate).

bdraco:

Clearing a nested scalar shorthand's text field (a pin `mode: ${my_mode}` made editable by #1344) correctly drops the key from the YAML, but the form's in memory values keep the empty string, and `renderNestedField`'s scalar shorthand branch treated any string as an unmodellable shorthand. The user saw the read only notice with a blank interpolation, "Set to  in YAML; edit it in the YAML pane on the right.", and the group was stranded until a reload re parsed the YAML.

The scalar branch now excludes the empty string, so a cleared value falls through to the normal collapsible flag group, matching what a reload already produced. An audit of the other editable text fallbacks (FLOAT, float with unit, time period, boolean, pin number) confirmed they all already recover from a cleared value; this branch was the only read only notice reachable from an in memory empty string.

Verified in the live editor: clearing the mode field's `${my_mode}` now renders the flag group immediately instead of the stranded notice.

- fixes https://github.com/esphome/device-builder-frontend/issues/1373

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/config-entry-renderers/nested.ts
+++ b/src/components/device/config-entry-renderers/nested.ts
@@ -37,9 +37,12 @@ function _enableStash(ctx: RenderCtx): Map<string, Record<string, unknown>> {
 export function renderNestedField(entry: ConfigEntry, path: string[], ctx: RenderCtx) {
   // A scalar at a NESTED key (an unmodellable shorthand the user set in
   // YAML) renders read-only with its value, not as an empty flag group.
+  // Empty string is a cleared value, not a shorthand — fall through to the
+  // flag-group editor or the notice strands the group until reload (#1373).
   const raw = ctx.getAt(path);
   if (
     !entry.multi_value &&
+    raw !== "" &&
     (typeof raw === "string" || typeof raw === "number" || typeof raw === "boolean")
   ) {
     return html`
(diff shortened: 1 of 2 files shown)


bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

Issue #1365, opened 2026-07-22
Float with unit field silently clears non finite input and strands junk strings read only


bdraco:
The float with unit magnitude input in `src/components/device/config-entry-renderers/primitives-numeric.ts` (the `@input` around line 340) maps non finite input to `null`:

```ts
const next = raw === "" ? null : Number(raw);
emit({ value: Number.isFinite(next) ? next : null, unit });
```

A typed `1e309` becomes `null`, which `serializeFloatWithUnit` turns into an empty value, silently clearing what the user typed. That is the opposite of the policy #1364 establishes for the other numeric emit paths (ship verbatim so the validator flags it).

The read side has the matching half of the gap: `renderFloatWithUnitField`'s unparseable bail still uses `looksLikeSubstitution` only, so a stored junk string like `21X` or `1e309` under a FLOAT_WITH_UNIT field is stranded in the read only YAML notice, while the plain FLOAT renderer now keeps any string editable as text (#1364). Fixing the emit without the bail would trade silent clearing for stranding, so the two belong together, probably as a shared string editable / primitive shell helper beside `renderYamlOnlyFallbackIfNonPrimitive` in `config-entry-renderers-shared.ts` used by both renderers.

`validateEntry`'s FLOAT_WITH_UNIT branch already flags such values via `parseFloatWithUnit`, so only the renderer halves need work.

PR #1376, opened 2026-07-22
Bail color fields to editable text when the value is not #rrggbb

### LINKED ISSUE #1371: Color field normalizes named colors and substitutions to black and clobbers on first pick
`config-entry-form.ts` routes `ConfigEntryType.COLOR` to `renderStringField(entry, "color", ...)` (around line 968). An `<input type="color">` cannot represent a named color or a `${substitution}`; the browser normalizes any non `#rrggbb` value to `#000000`, so the stored value renders as a plausible looking black swatch and the first pick emits hex, silently replacing it. Same data loss class as #1365/#1368, without even the unchecked switch tell.

Fix shape, following #1370: bail to `renderUnparseableScalarField` (`config-entry-renderers-shared.ts`) when the value fails `/^#[0-9a-f]{6}$/i`, so named colors and substitutions stay editable as text.

bdraco:

The COLOR entry type routed straight to a `<input type="color">`, which can only represent `#rrggbb`; the browser normalizes anything else to `#000000`, so a named color or `${substitution}` would render as a plausible looking black swatch and the first pick would silently replace it. Same data loss class as #1365/#1368, without even the unchecked switch tell.

A new `renderColorField` bails any present value that is not `#rrggbb` to `renderUnparseableScalarField` from #1367: named colors, substitutions, short hex, and junk stay editable as text, a non string value gets the YAML only shell, and `#rrggbb` and empty keep the native picker. The dispatch case calls the new renderer; nothing else changes.

No catalog field carries `type: color` today (noted on the issue), so this is not reachable in app to demo; the guard exists so the first schema sync that emits the type does not ship the clobber. All three branches are pinned in the bail suite.

- fixes https://github.com/esphome/device-builder-frontend/issues/1371

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/config-entry-form.ts
+++ b/src/components/device/config-entry-form.ts
@@ -94,6 +94,7 @@ import {
   renderMultiValueField,
   renderNestedField,
   renderNestedListField,
+  renderColorField,
   renderNumberField,
   renderPinField,
   renderRegistryListField,
@@ -966,7 +967,7 @@ export class ESPHomeConfigEntryForm extends LitElement {
       case ConfigEntryType.PIN:
         return renderPinField(entry, path, ctx);
       case ConfigEntryType.COLOR:
-        return renderStringField(entry, "color", path, ctx);
+        return renderColorField(entry, path, ctx);
       case ConfigEntryType.MAC_ADDRESS:
         return renderStringField(entry, "text", path, ctx);
       case ConfigEntryType.LAMBDA:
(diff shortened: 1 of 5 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1380, opened 2026-07-22
Re-emit edited flow lists in flow style with their trailing comment

### LINKED ISSUE #1378: Editing a flow list converts it to block style and drops its trailing comment
Follow up from #1377's review. A flow list's inline comment is captured in `ParsedSection.comments` (the reader splits it before the bracket test), but an edited flow list re serializes through the array branch as multi line block style, so the #1235 re append never fires (it is gated on `fresh.length === 1`): the style silently converts flow to block and the trailing comment drops.

Reachable today: the display `user_characters` `data: [..] # note` shape. Fix candidates: a flow aware re emit (single line via `formatYamlFlowScalar`, comment re appended), or extend `ListItemSource` with a flow marker so the splice keeps the authored style.

bdraco:

A flow list's trailing comment is captured at parse (`data: [10, 12] # note` splits before the bracket test), but an edited flow list re serialized through the array branch as multi line block style: the authored style silently flipped and, because the re emit was no longer a single line, the #1235 comment re append never fired, so the comment dropped too.

The reader now records flow authored keys in a `flowListKeys` set on `ParsedSection` (cleared on duplicate key re assignment, the same hygiene #1377 established for `listSources`). When such a key changes and the new value is a non empty all scalar array, the splice re emits it as the same single flow line via the existing `formatYamlFlowScalar` (now exported; it is the flow indicator aware quoter that keeps `"${glyph_row}"` protected), and the existing single line comment re append restores the trailing comment with no further code. An emptied list still drops the key, and non scalar values fall through to the normal re emit.

Verified in the live editor: editing one row of a seeded `libraries: [Wire, SPI] # deps` keeps the single flow line and the comment. Pinned for the edit, the substitution quoting, and the emptied list drop; the unchanged flow list byte verbatim pin and the #1377 block list pins stay green.

- fixes https://github.com/esphome/device-builder-frontend/issues/1378

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/util/yaml-section-reader.ts
+++ b/src/util/yaml-section-reader.ts
@@ -93,12 +93,14 @@ export function parseSectionCore(
   const spans = new Map<string, KeySpan>();
   const comments = new Map<string, string>();
   const listSources = new Map<string, ListItemSource>();
+  const flowListKeys = new Set<string>();
   // leadStart is finalised by the post-loop pass below.
   const recordSpan = (key: string, start: number, end: number): void => {
     spans.set(key, { start, end, leadStart: start });
-    // A duplicate key re-assigns the value; a stale list source pointing
-    // at the first occurrence's lines must not survive it.
+    // A duplicate key re-assigns the value; stale list metadata pointing
+    // at the first occurrence must not survive it.
     listSources.delete(key);
+    flowListKeys.delete(key);
   };
   const startIdx = findSectionStart(lines, sectionKey, fromLine);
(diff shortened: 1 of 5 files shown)


bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1391, opened 2026-07-23
Gate experience_level off the remote-builder onboarding path

### LINKED ISSUE #1390: Remote-builder onboarding silently overwrites experience_level with BEGINNER
Flagged by both Copilot (inline) and the Kōan CHANGES_REQUESTED review on #1337, but merged unfixed: `_persistChoices` in `src/components/onboarding/onboarding-wizard-dialog.ts` always sends `experience_level: this._experience`, and `open()` resets `_experience` to BEGINNER. On the desktop "remote builder" path the experience screen is never shown, so re running onboarding (which the PR explicitly supports, and which writes `hide_device_builder`/`remote_compute_only` on that re run) silently downgrades a prior EXPERT/ADVANCED user to BEGINNER — dropping them out of expert mode across the app from a screen that never asked about experience.

`updatePreferences` takes a `Partial`, so the fix is the same gate already used for `hide_device_builder`: omit the field when the experience screen was not shown, e.g. `...(remoteBuilder ? {} : { experience_level: this._experience })`.

While in there, the same PR's other unaddressed review threads are worth sweeping: the `wizard-screens.ts` docstring still claims "Welcome and experience are always mandatory" despite the remote builder early exit; the "experience is at index 1" comment (and possibly the adjacent `this._index <= 1` freeze guard) doesn't account for the usage screen shifting the index mapping on desktop — the guard may pin the existing server tail one screen early there, needs verification; and the standalone warning banner doesn't name the discovered install(s) as the PR description promised.

bdraco:

Completing onboarding on the desktop remote builder path persisted `experience_level: BEGINNER` even though the experience screen was never shown; `open()` resets the field and `_persistChoices` always sent it, so re running onboarding to switch into remote build mode silently downgraded a returning EXPERT user out of expert mode. The field is now omitted on that path, the same gate shape `hide_device_builder` already uses; the standalone and non desktop flows keep sending the chosen level. Verified in the live editor: with a stored `expert` level, completing the remote builder path wrote `remote_compute_only` and `hide_device_builder` and left the level untouched.

This also sweeps the other unaddressed review threads from #1337: the standalone warning banner now names the discovered install(s), reusing the existing_server screen's name formatting through one hoisted helper (unique names, cap of three, overflow count); the `wizard-screens.ts` docstring no longer claims experience is always mandatory; and the `_screens` freeze guard comment now says why index 1 is correct, the step only exists in the non desktop flow where experience sits at index 1 (verified, no logic change needed).

- fixes https://github.com/esphome/device-builder-frontend/issues/1390

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/onboarding/onboarding-wizard-dialog.ts
+++ b/src/components/onboarding/onboarding-wizard-dialog.ts
@@ -179,9 +179,11 @@ export class ESPHomeOnboardingWizardDialog extends LitElement {
 
   /** Ordered screens for the current environment. */
   private get _screens(): WizardScreen[] {
-    // existing_server sits after experience (index 1). While the user is still
-    // on Welcome or experience the tail can grow as mDNS hosts arrive; freeze it
-    // once they advance past experience so a late host can't shift the flow.
+    // existing_server can appear only in the non-desktop flow, where experience
+    // is index 1 (wizardScreens ignores the flag when the usage screen shows).
+    // While the user is still on Welcome or experience the tail can grow as
+    // mDNS hosts arrive; freeze it once they advance past experience so a late
+    // host can't shift the flow.
     const showExistingServer =
       this._index <= 1 ? this._computeShowExistingServer() : this._existingServerPinned;
     return wizardScreens({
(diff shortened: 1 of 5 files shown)


bdraco:
@esphbot review

bdraco:
Empirical before/after on the live dashboard, identical walk both times (stored level set to `expert` over the WS API, then welcome, usage with remote build preselected, Continue to complete):

Pre fix (origin/main dialog in the working tree): read back `{"experience_level":"beginner","remote_compute_only":true,"hide_device_builder":true}`; the stored expert was silently downgraded.

With this branch: read back `{"experience_level":"expert","remote_compute_only":true,"hide_device_builder":true}`; the level survives.

How the bug lands in production, traced through `app-shell` → `onboarding-gate.ts` → the backend's `onboarding.py`: the wizard only auto pops when no experience level is stored, so the mainline victim is an existing ESPHome user on a fresh desktop install, exactly who the remote build card recommends itself to. Picking remote build stamped `beginner` from a flow that never asked, and that stamp is doubly sticky: the `!isExperienceChosen` gate means the wizard never auto pops again, and the backend's `migrate_preexisting_install` (which defaults installs that completed onboarding to EXPERT) is a no op once a level exists, so the bogus beginner blocks the very migration that would have corrected it. Backing out of remote only mode later lands them in the builder locked to beginner with nothing ever asking. With the fix the level stays null on the remote path, and on the next backend start the pre existing install migration classifies them EXPERT, matching the card's "recommended for existing users" framing. A stored EXPERT also stays safe against a mid wizard Settings write and against any future onboarding re run surface (the backend explicitly plans `ONBOARDING_VERSION` bump re prompts).

esphbot (approved):
> [!TIP]

PR #1416, opened 2026-07-24
Freeze collision checks while the create is in flight

bdraco:

Follow-up visual fix to #1411: submitting the create wizard flashed a red self-collision error on the hostname field for a moment before the dialog closed.

`createDevice`'s success is announced by a devices push that lands while the dialog is still open, so the live collision check (added in #1411) briefly flagged the name being created as "already exists". The setup step now passes an empty taken-hostnames set to the name inputs while `submitting` is true, so the in-flight create can't collide with itself; the real set returns if the create fails and the dialog stays open.

Test pins it: with `submitting` set, a push containing the device's own slug produces no error and leaves `canSubmit` true.

- visual regression from the live collision detection in #1411, observed on first real device create

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/wizard/wizard-step-setup.ts
+++ b/src/components/wizard/wizard-step-setup.ts
@@ -27,6 +27,11 @@ import "@home-assistant/webawesome/dist/components/checkbox/checkbox.js";
 import "@home-assistant/webawesome/dist/components/spinner/spinner.js";
 import "../shared/device-name-inputs.js";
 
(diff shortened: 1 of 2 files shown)


bdraco:
@esphbot review

bdraco:
@esphbot review

PR #1418, opened 2026-07-24
Group GPIO pin logic into dedicated pin directories

### LINKED ISSUE #1414: Reorganize all pin logic into its own dir where it makes sense
Its already big, but https://github.com/esphome/device-builder-frontend/pull/1412 made bigger.

Pins are the things users struggle with them most when getting started so its important to get it right, but also we need to make it easier to maintain

bdraco:

The pin surface had grown to eight source modules plus their tests, sitting flat across `src/util/` and `src/components/device/`. This groups the GPIO pin logic into two dedicated directories so it is easier to find and maintain: `src/util/pin/` holds the DOM-free logic and `src/components/device/pin/` holds the render helpers, with the tests mirrored. The redundant `pin-` and `config-entry-pin-` filename prefixes are dropped now that the directory carries the namespace.

While mapping the files I found the `pin-` prefix actually covers two unrelated subsystems; the SHA-256 pairing-fingerprint files (`pin-emoji`, `pin-format`, `pin-emoji-grid`, `pin-hex`) only share the substring, so they stay exactly where they are rather than getting swept into a GPIO directory.

No behaviour change. Imports across the five external consumers and the moved tests are retargeted to the new paths; `pnpm run lint`, `pnpm run test` (6208 tests), and prettier all pass.

- fixes https://github.com/esphome/device-builder-frontend/issues/1414

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1419, opened 2026-07-24
Lock the GPIO select on board-preconfigured pins

### LINKED ISSUE #1417: Board preconfigured pins should lock the GPIO select too
PR #1412 made the wiring section of a board preconfigured pin view only: when a pin sits on a GPIO the board designates for the section (a featured component's locked_pins, a fields preset, or suggestions), the mode and inverted fields render read only under a "Preconfigured for this board" row, and the tooltips point at the YAML editor as the last resort override.

The GPIO select itself is still editable in that state. Moving a preconfigured pin to another GPIO is the same class of almost certain mistake as changing its mode flags, so the select should get the same treatment: disabled while the guard is active, with the same "Preconfigured for this board. Edit the YAML to change it." tooltip.

Notes for the implementation:

- Hard locked pins (entry.locked) already disable the select through effectiveDisabled. This issue is about the softer board designations that arm the wiring guard (boardPresetPin in config-entry-pin-renderer.ts) without locking the entry.
- A disabled wa-select whose current value has no matching option renders blank; the renderer already re-adds the active pin for that reason, keep that working.
- Test coverage belongs beside the "pin wiring board-preset guard" describe in test/components/device/config-entry-pin-wiring.test.ts.

bdraco:

PR #1412 made the wiring section of a board preconfigured pin view only, but the GPIO select itself stayed editable. Moving a preconfigured pin to another GPIO is the same class of mistake as changing its mode flags, so the select now disables in lockstep with the wiring guard. The guard hint is discoverable: hovering a disabled control never shows a native title tooltip, so the select carries a wa-tooltip (the same pattern visit-web-ui-link uses) and the label shows the hard lock's icon chrome retargeted at the YAML editor.

The add-component dialog had the same gap because it never passed a section key to the shared form, so board designations could not match. It now threads the resolved component id through, which arms the pin guard (and the #1412 wiring guard) when adding a featured component like the ESK-1 RGB LEDs whose pin preset is not hard locked. renderPinWiring now receives the computed guarded flag instead of rederiving it, so the select and the panel stay in lockstep by construction.

Hard locked pins are unchanged; a pin the user moved off the board's designation stays fully editable. The renderer already forces the active pin's option into the dropdown, so the disabled select shows the real pin instead of blanking.

- fixes https://github.com/esphome/device-builder-frontend/issues/1417

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/add-component-form.ts
+++ b/src/components/device/add-component-form.ts
@@ -298,6 +298,7 @@ export class ESPHomeAddComponentForm extends LitElement {
           .errors=${this._errors}
           .board=${this.board}
           .yaml=${this.yaml}
+          .sectionKey=${this._sectionKey}
           .presentComponents=${presentComponents}
           ?disabled=${disabled}
           ?required-only=${true}
@@ -475,11 +476,15 @@ export class ESPHomeAddComponentForm extends LitElement {
     if (this._localBlockMessage) this._localBlockMessage = "";
   }
 
+  /** Section key the block resolves to — featured ids are synthetic
+   *  (``featured.<board>.<local>``); the underlying component id keys the
+   *  committed YAML and scopes the board's featured designations. */
+  private get _sectionKey(): string {
+    return resolveFeaturedComponentId(this.component.id, this.board);
+  }
+
   private _generateYamlPreview(): string {
-    // Featured ids are synthetic (`featured.<board>.<local>`); preview the
-    // underlying component the block resolves to, matching the committed YAML.
-    const key = resolveFeaturedComponentId(this.component.id, this.board);
-    const lines: string[] = [`${key}:`];
+    const lines: string[] = [`${this._sectionKey}:`];
     lines.push(...serializeYamlValues(this._values, "  "));
     return lines.join("\n");
   }
--- a/src/components/device/config-entry-form.ts
+++ b/src/components/device/config-entry-form.ts
@@ -243,9 +243,10 @@ export class ESPHomeConfigEntryForm extends LitElement {
    *  against the parent component's domain — without it a
    *  binary_sensor's filter picker offers sensor-only filters
    *  and a non-addressable light's effects picker offers
-   *  addressable-only effects. Empty when the form is mounted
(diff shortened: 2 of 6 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

bdraco:
Checked the title-on-disabled concern empirically in the running dashboard: a disabled wa-select host keeps pointer-events auto and no shadow part sets pointer-events none, so hover still reaches the host and the title tooltip surfaces (wa-select is a custom element; it does not get the native disabled pointer-event suppression). The wiring panel guard row also carries the same hint as a fallback.

Also consolidated the guard: renderPinWiring now receives the computed guarded flag instead of re-deriving it from boardPreset, so the select and the panel lock in lockstep by construction.

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Correction to my earlier note: hands-on hover showed the native title does NOT surface on the disabled select after all (not-allowed cursor, no tooltip), so the bot's concern was right. Replaced it with the pattern the codebase already uses for this (visit-web-ui-link): a wa-tooltip anchored on the select, plus the label now carries the same lock icon chrome a hard-locked pin gets, retargeted at the YAML editor.

esphbot (approved):
> [!TIP]

bdraco:
On suggestion 1: not a real gap in practice, since no board manifest designates a pin via a substitution. A ${var} that resolves onto a designation is the user's own hand-authored setup, and the reference input edits the substitution name, not the wiring the panel guards, so locking it would only strand the escape hatch. Added the one-line comment at the input noting the intentional scope.

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1437, opened 2026-07-24
Re-bind the logs/install method picker to the live device object

### LINKED ISSUE #1431: Logs are harder to access than they need to be after an OTA
If you click the logs button and the dialogue pops up, the over the network option is grayed out when the device is not online. It's common that the device will be off-line for a few moments after OTA, but the option stays greyed out even when the device comes back online.

bdraco:

On the dashboard page the install/logs method picker snapshots the device object into `_installMethodDevice` when it opens, but device state events replace the object in the reactive list, so the snapshot never sees the OFFLINE to ONLINE transition. In logs mode the "Connect wirelessly" row gates on the device being online, so after an OTA reboot the row stayed grayed out until the dialog was closed and reopened.

The fix mirrors the drawer's existing live re-bind in `willUpdate`: on every `_devices` change the snapshot is swapped for the live object (the OTA row, address seed, bootloader and never-flashed bindings all refresh), and a device that leaves the list closes the picker, same as the drawer. As a side effect the method selected from the picker now acts on the live device object instead of the stale snapshot.

The device detail page was not affected; its controller reads state through a getter into the context-subscribed device list. The other dashboard snapshots (`_cardContextDevice`, `_actionDevice`) back transient menus and confirm dialogs with no online gating, so they are left alone.

Both re-bind blocks now share a `relinkLive` lookup in `live-rebind.ts`. Test side, the new regression test also asserts the dialog's `deviceState` prop through the DOM, and the dashboard page tests' duplicated child mock block is hoisted into `test/pages/_mock-dashboard-children.ts` (used by the two existing page tests too).

- fixes https://github.com/esphome/device-builder-frontend/issues/1431

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/dashboard/live-rebind.ts
+++ b/src/components/dashboard/live-rebind.ts
@@ -1,0 +1,14 @@
+import type { ConfiguredDevice } from "../../api/types/devices.js";
+
+/**
+ * The live ``_devices`` object matching *current*'s configuration, or ``null``
+ * when the device left the list. Device state events replace list objects
+ * wholesale, so any snapshot held across renders must re-link through this on
+ * every ``_devices`` change or it silently stops seeing updates.
+ */
+export function relinkLive(
+  devices: ConfiguredDevice[],
+  current: ConfiguredDevice
+): ConfiguredDevice | null {
+  return devices.find((d) => d.configuration === current.configuration) ?? null;
+}
--- a/src/pages/dashboard.ts
+++ b/src/pages/dashboard.ts
@@ -601,6 +601,20 @@ export class ESPHomePageDashboard extends LitElement {
         this._drawerOpen = false;
       }
     }
+    // Same re-bind for the install/logs method picker: its OTA row gates on
+    // the device being online, and a device rebooting after an OTA must
+    // un-gray the row while the picker sits open (#1431).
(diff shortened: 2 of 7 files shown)


bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Applied both Kōan suggestions in f6c6fbea: the drawer and picker re-bind blocks now share `relinkLive` in `src/components/dashboard/live-rebind.ts` (the `!` assertions are gone), and the regression test mounts with `_devicesLoaded` so it asserts the dialog's `deviceState` prop through the DOM, not just the page state field. PR body updated to mention the test-mock extraction.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Added the missing unit half in a41145ad: `test/components/dashboard/live-rebind.test.ts` pins the match rule directly (same configuration returns the live object, gone from the list and empty list return null), so the drawer half no longer rides only on the picker's end to end test.

@esphbot review

PR #1460, opened 2026-07-25
Type the section-editor event contract

### LINKED ISSUE #1458: Type the section-editor event contract
Follow up to #1453 / #1456, which typed the member half of the section editor contract. The event half stays duck typed: `section-mount` / `section-unmount` with `{ node }` (fired by `device-section-config` and `AutoApplyController`) and `dirty-change` with `{ dirty }`, consumed by `device.ts`. A rename or detail shape change on either firer is still a silent runtime break, the same drift class the interface eliminated for members.

Proposal: declare the detail types alongside `SectionEditor` in `src/components/device/section-editor.ts` and augment `HTMLElementEventMap` so both `fireEvent` call sites and the page's `@...` bindings type check against them.

bdraco:

Types the event half of the section editor contract, finishing what #1456 did for the member half. The trio the page depends on (`section-mount` / `section-unmount` with `{ node }`, `dirty-change` with `{ dirty }`) was still duck typed between its two firers and `device.ts`.

`section-editor.ts` now declares the detail interfaces and a `SectionEditorEventMap`, both firers dispatch through a `fireSectionEvent` wrapper that checks the name and the detail shape at compile time, and a global `HTMLElementEventMap` augmentation types the consumer side, letting `device.ts` drop its three event casts. The generic `fireEvent` stays untyped on purpose: call sites across the repo already use names that collide with built in DOM map keys (`change`, `cancel`, `close`), so a repo wide typed overload would break them; the wrapper scopes enforcement to the trio that has two independent firers.

`AutoApplyHost` now extends `SectionEditor`, which is truthful (every real host exposes those members by delegating to the engine) and is what lets the controller's `{ node: this._host }` dispatch type check. The engine test's host stub gains the three members. Enforcement is `tsc`; no behavior change.

- fixes https://github.com/esphome/device-builder-frontend/issues/1458

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/automation-editor/auto-apply-controller.ts
+++ b/src/components/device/automation-editor/auto-apply-controller.ts
@@ -7,10 +7,10 @@ import type {
 } from "../../../api/types/automations.js";
 import type { LocalizeFunc } from "../../../common/localize.js";
 import { fireEvent } from "../../../util/fire-event.js";
-import type { SectionEditor } from "../section-editor.js";
-import { fireSectionEvent } from "../section-editor.js";
 import { formatApiError } from "../../../util/format-api-error.js";
 import { notifyError } from "../../../util/notify.js";
+import type { SectionEditor } from "../section-editor.js";
+import { fireSectionEvent } from "../section-editor.js";
 import { applyYamlDiff, emptyAutomationTree } from "./serialise.js";
 
 /** Debounce window between a value change and the auto-apply upsert.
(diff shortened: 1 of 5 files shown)


bdraco:
@esphbot review

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Import placement breaks `format:check` (organize-imports sort)

bdraco:
The blocker and two of the three suggestions are in the latest push: the two section-editor imports now sit where organize-imports sorts them (below notify.js), the HTMLElementEventMap augmentation derives every detail from SectionEditorEventMap so the two ends cannot drift, and the shared lifecycle detail is renamed SectionLifecycleDetail so the sharing reads as deliberate. Deferring the central registry: the augmentation stays feature scoped until a second claimant exists, at which point hoisting one project wide map next to fireEvent (the HA HASSDomEvents shape) is the right move; noted in this thread rather than the code so the first mover does not build unused infrastructure.

@esphbot review

esphbot (approved):
> [!TIP]

PR #1469, opened 2026-07-25
Restructure AutoApplyController's flags into explicit apply and delete modes

### LINKED ISSUE #1461: Restructure AutoApplyController's flag choreography into an explicit state machine
`AutoApplyController` now coordinates five interacting control flags (`_applyTimer`, `_applyInFlight`, `_applyDirty`, `_deleting`, `_applySuppressed`, plus the `_connected` environment mirror), and their legal combinations are enforced only by guard ordering. The last three defects fixed in #1455 were all flag disagreements about who owns the section: the suppressed requeue, the failed delete re-arm, and the mid flush retarget.

Proposal, as a behavior free refactor after #1455 merges:

1. Replace the apply pipeline trio with an explicit state union, `idle | debounced | applying | applying-queued`, so illegal combinations (a timer armed while applying, a queued re-run with no apply in flight) become unrepresentable.
2. Model delete as an owning mode with its own transitions (`delete-flushing | delete-writing`), absorbing the suppression and re-arm rules that currently live as flag choreography in `delete()`'s finally.
3. Keep `_connected` and the UI facing `dirty` orthogonal: one is an environment signal, the other a projection.

The engine's unit suite (23 tests as of #1455, including ordering, suppression, re-arm, retarget, and teardown cases) pins every transition the machine must preserve, so the refactor is well guarded.

bdraco:

Behavior free refactor of `AutoApplyController`'s flag choreography into explicit modes, per the follow up filed off the #1455 review. The apply pipeline trio (`_applyInFlight`, `_applyDirty`, plus their guard ordering) becomes one `ApplyPhase` union, `idle | applying(queued)`, so a queued re-run with no apply in flight is unrepresentable. Delete becomes an owning `DeleteMode` object that carries the suppression record, so a suppression cannot exist without a delete and cannot outlive the one that captured it; the four outcome branches in `delete()`'s finally move to an extracted `_settleDelete`, which also takes the extraction Kōan suggested on #1455's final round. `_connected` stays an environment mirror and `dirty` a UI projection, both orthogonal.

One deliberate deviation from the issue sketch: the debounce timer stays its own cancellable handle rather than a union arm. A timer armed while an apply is in flight is a legal state today (typing during a round trip), and teardown must cancel it without touching the in flight call; folding it into the union would either drop that coalescing or change teardown semantics. The `delete-flushing | delete-writing` sub phases are also omitted since nothing reads them yet; the mode object is the place to add them when something does.

No test changes: the engine's 27 unit tests (ordering, suppression, re-arm, retarget, teardown) pass unmodified, which is the point, they pin every transition the modes must preserve.

Chained on #1455 (base `fix-delete-upsert-race`); retarget to `main` after it merges.

- fixes https://github.com/esphome/device-builder-frontend/issues/1461

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/automation-editor/auto-apply-controller.ts
+++ b/src/components/device/automation-editor/auto-apply-controller.ts
@@ -79,8 +79,8 @@ interface DeleteMode {
  * Shared auto-apply / delete / dirty-tracking engine for the three
  * automation-section editors (automation, script, api-action).
  *
- * Owns the debounce timer, the in-flight/dirty coalescing state, the
- * self-written-YAML echo marker, and the ``section-mount`` /
+ * Owns the debounce timer, the apply phase (in-flight + queued
+ * re-run), the self-written-YAML echo marker, and the ``section-mount`` /
  * ``section-unmount`` lifecycle announcements the device page uses to
  * hold a direct ref to the active section (see device.ts
  * ``_onSectionMount``). The controller shape gives proper teardown:
--- a/src/components/device/automation-editor/auto-apply-controller.ts
+++ b/src/components/device/automation-editor/auto-apply-controller.ts
@@ -53,6 +53,28 @@ export interface AutoApplyOptions {
   setError(message: string): void;
 }
 
+/** Apply pipeline phase. A queued re-run is only representable
+ *  while a call is in flight, so the flag pair that used to drift
+ *  apart (in-flight false with a stale queue bit, and vice versa)
+ *  is now one value. The debounce timer stays a separate handle:
+ *  typing during an in-flight round trip legitimately arms it
+ *  while applying, and teardown cancels it without touching the
+ *  in-flight call. */
+type ApplyPhase = { kind: "idle" } | { kind: "applying"; queued: boolean };
+
+/** A delete owning the section. Carrying the suppression record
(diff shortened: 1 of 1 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Addressed in 38e54b79:

1. Agreed on the gap and that closing it is a behavior change that does not belong in this refactor; filed as #1472 with your fall-through shape.

2. Taken, the field is now `_deleteMode` with `_setDeleteMode`, matching the type and the prose.

3. Dropped the guard, `const queued = this._apply.queued;` compiles clean under the pinned tsc exactly as you verified, so the finally now reads as the single owner transition it is.

Engine suite 27/27 and full suite green at 38e54b79.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Taken in 4b54757d, the class docstring now says apply phase (in-flight + queued re-run) instead of the retired in-flight/dirty pair.

@esphbot review

esphbot (approved):
> [!TIP]

PR #1474, opened 2026-07-25
Hold flushPending until the queued re-run settles

### LINKED ISSUE #1472: flushPending resolves without awaiting the queued re-run when a debounce and an in-flight apply overlap
Noted during the #1469 review; pre-existing on main (the branch tables are identical before and after the refactor). When a debounce timer and an in-flight apply overlap, the legal state the #1469 design note calls out (a keystroke landing inside the 200ms window while a round trip is outstanding), `flushPending()` takes the debounce branch: it cancels the timer and awaits `autoApply()`, which immediately queues a re-run on the in-flight call and resolves. The queued re-run is never awaited.

`src/pages/device.ts` awaits `flushPending()` before the global save specifically so the save commits exactly what the user typed. In the overlap state the save commits the pre keystroke YAML, and the queued re-run then fires a `yaml-draft` right after, re dirtying the buffer the user just saved.

The fix is small once #1469's `ApplyPhase` union lands, fall through to the settle poll instead of returning after the debounce branch:

```ts
async flushPending(): Promise<void> {
  if (this._debounce) {
    this._cancelDebounce();
    await this.autoApply();
  }
  while (this._apply.kind === "applying") {
    await new Promise((r) => setTimeout(r, 20));
  }
}
```

This is a behavior change with its own timing sensitive tests (the macrotask barrier notes in `delete()` apply), so it stays out of the behavior free refactor.

bdraco:

When a debounce timer and an in-flight apply overlapped (a keystroke landing inside the 200ms window while a round trip was outstanding), `flushPending()` cancelled the timer and awaited `autoApply()`, which merely queued a re-run on the in-flight call and resolved. The caller was released while the keystroke it came to flush was still unwritten, so the device page's global save could commit the pre keystroke YAML and the queued re-run then re dirtied the buffer the user just saved.

The debounce branch now falls through to the settle poll instead of returning, so `flushPending()` resolves only once nothing is applying, including the queued re-run. The #1469 `ApplyPhase` union is what makes the poll condition cover both.

The new test pins the overlap: the flush must not release while the first round trip is outstanding, and the queued re-run must have run before it does. Verified the test fails on the pre fix code.

- fixes https://github.com/esphome/device-builder-frontend/issues/1472

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/automation-editor/auto-apply-controller.ts
+++ b/src/components/device/automation-editor/auto-apply-controller.ts
@@ -208,20 +208,26 @@ export class AutoApplyController implements ReactiveController {
    * Force a pending debounced auto-apply to flush immediately. The
    * device page calls this (via the host) on the active section
    * before its global save so the YAML buffer is fully caught up.
-   * Resolves only once nothing is applying — including the queued
-   * re-run a debounce cancelled into an in-flight call.
+   * Resolves only once nothing is pending — neither a debounce
+   * (even one armed mid-poll by a late keystroke) nor an in-flight
+   * call, including the queued re-run a cancelled debounce leaves
+   * on one.
    */
   async flushPending(): Promise<void> {
-    if (this._debounce) {
-      this._cancelDebounce();
-      // With a call already in flight this only queues a re-run —
-      // fall through to the poll so the caller isn't released while
-      // the keystroke it came to flush is still unwritten.
-      await this.autoApply();
-    }
-    // Wait for the in-flight call (and any queued re-run) to settle.
-    while (this._apply.kind === "applying") {
-      await new Promise((r) => setTimeout(r, 20));
+    // Terminates: during a delete, ``scheduleAutoApply`` refuses to
+    // arm (the delete mode suppresses instead), so ``delete()``'s
+    // call site can't spin; outside one, each iteration retires the
+    // pending work that gated it.
+    while (this._debounce || this._apply.kind === "applying") {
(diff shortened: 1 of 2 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Both taken in 67fc237a:

1. Folded the branches into your single loop, so a debounce armed during the poll is retired the same way; the docstring and a termination comment carry the delete-mode argument. A new test pins the mid-poll keystroke case (flush enters the poll with no debounce, keystroke lands during it, release only after its upsert).

2. Both flush tests now use the order-array pattern and assert `["draft", "draft", "flushed"]`, pinning that the drafts reached the page before the caller was released rather than the call-count proxy.

Suite 29/29, full run green at 67fc237a.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Keeping the loop unbounded intentionally. Hold-until-quiet is the flush contract, the save must commit exactly what the user typed, and a cap re-opens a variant of the same bug (the save commits behind whatever landed after the Nth retire, just with a bigger N). Your own analysis shows it is self limiting: one upsert per round trip regardless of keystroke rate, and it drains the moment input stops since nothing but user handlers re-arm the debounce. A Save spinner that outlasts active typing is the honest presentation of that.

PR #1484, opened 2026-07-25
Ride the section-unmount announcement on the mount-time parent

### LINKED ISSUE #1483: section-unmount dispatches from a detached host and never reaches the page
Noted during the #1475 review; pre-existing. `disconnectedCallback` runs after the node leaves the tree, so the `section-unmount` dispatched there fires from a detached element and never reaches the page's `.layout-grid` listener — the identical bubbles-nowhere mechanism #1475 fixed for the delete paths' `yaml-updated`. The mirror `section-mount` lands fine (the element is already inserted in `connectedCallback`), so the asymmetry is easy to miss.

Two emitters share the shape: `device-section-config`'s `disconnectedCallback` and `AutoApplyController.hostDisconnected` (which the three automation editors ride). The receiver `_onSectionUnmount` clears `_activeSection` and `_sectionDirty`; today the miss is mostly masked because the incoming section's `section-mount` re-points `_activeSection`, and `_sectionDirty` is OR'd with `_isYamlDirty`. The exposed shape is leaving a section with no replacement while `_sectionDirty` is the only dirty signal, which strands the unsaved-changes guard armed (or the reverse: a stale `_activeSection` whose flushPending the save path still awaits).

With `fireFromAnchor` on main this is a one liner per emitter: snapshot `parentNode` in `connectedCallback` / `hostConnected` and dispatch the unmount announcement through it with `connected: false`.

bdraco:

`disconnectedCallback` runs after the node has left the tree, so the `section-unmount` both section editors announce there dispatched from a detached element and never reached the page's `.layout-grid` listener, the same bubbles nowhere mechanism #1475 fixed for the delete paths' `yaml-updated`. The page's `_onSectionUnmount` clears `_activeSection` and `_sectionDirty`; the miss is mostly masked because the incoming section's `section-mount` re points `_activeSection`, but leaving a section with no replacement strands a stale `_sectionDirty` (arming the unsaved changes guard) and a stale `_activeSection` whose `flushPending` the save and switch paths still consult.

Both emitters (`device-section-config.disconnectedCallback` and `AutoApplyController.hostDisconnected`) now snapshot their mount time parent at connect and dispatch the unmount announcement through it, `fireSectionEvent(anchor ?? host, ...)`. The typed helper form is used rather than `fireFromAnchor` because connectedness is always false in a disconnect hook, so the helper's parameter would be a constant; the never mounted fallback dispatches on the host, a harmless no op.

Two tests pin it, one per emitter, each detaching the element first and asserting the announcement lands on the shadow host with the right node in the detail.

- fixes https://github.com/esphome/device-builder-frontend/issues/1483

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/automation-editor/auto-apply-controller.ts
+++ b/src/components/device/automation-editor/auto-apply-controller.ts
@@ -117,6 +117,8 @@ export class AutoApplyController implements ReactiveController {
   // Whether the host element is on screen; a failure-path re-arm must not
   // schedule a write for a torn-down section.
   private _connected = false;
+  /** Mount-time parent carrying the detached-host unmount dispatch. */
+  private _mountAnchor: ParentNode | null = null;
 
   constructor(
     private readonly _host: AutoApplyHost,
@@ -130,6 +132,10 @@ export class AutoApplyController implements ReactiveController {
    *  device-section-config's section-mount event. */
   hostConnected(): void {
     this._connected = true;
+    // The unmount announcement must ride the mount-time parent:
+    // disconnectedCallback order means the host has already left the
+    // tree, and a detached dispatch bubbles nowhere (#1483).
+    this._mountAnchor = this._host.parentNode;
     fireSectionEvent(this._host, "section-mount", { node: this._host });
   }
 
@@ -138,7 +144,9 @@ export class AutoApplyController implements ReactiveController {
     // Cancel the pending debounced upsert — a write scheduled by a
     // section that's no longer on screen must not fire.
     this._cancelDebounce();
-    fireSectionEvent(this._host, "section-unmount", { node: this._host });
+    fireSectionEvent(this._mountAnchor ?? this._host, "section-unmount", {
+      node: this._host,
+    });
   }
 
   /** Brief-window dirty flag covering the debounce gap so the global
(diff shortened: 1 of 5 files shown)


bdraco:
@esphbot review

bdraco:
DRY pass at d876c03f: both emitters now call a shared `announceSectionMount(host)` in section-editor.ts, which announces the mount and returns the matching unmount announcer holding the captured anchor — the snapshot field, the rationale comment, and the `anchor ?? host` dispatch live once. Each emitter keeps only the returned closure. Behavior identical; both suites and the full run green at d876c03f.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
All three handled at a85028d1:

1. Both call sites null `_announceUnmount` right after firing, with a one-shot comment, so the contract lives in the code.

2. `setup()` now takes an optional parent so the anchor is in place for the first `hostConnected()`; the test no longer double-connects or double-fires `section-mount`.

3. The unguarded `dirty-change` is filed as #1486 — an identity guard needs `node` added to `SectionDirtyChangeDetail` (a contract change across both emitters and the payload tests), so it gets its own PR rather than riding this one.

Suites 32/32, full run 6425 green at a85028d1.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Both taken at dbb5207f: the anchor test now sits in the host-lifecycle describe directly below its counterpart, and that counterpart carries the note that its parentless setup exercises the `anchor ?? host` fallback while the anchored production path is pinned by the test below it. Full run green at dbb5207f.

@esphbot review

esphbot (approved):
> [!TIP]

PR #1489, opened 2026-07-26
Identity-guard dirty-change like the rest of the lifecycle trio

### LINKED ISSUE #1486: dirty-change is the one unguarded section lifecycle event
Noted during the #1484 review. Of the section lifecycle trio, `section-mount` and `section-unmount` now deliver reliably (identity guarded at the page, anchor carried past detachment per #1484), but `dirty-change` remains unguarded: `_onSectionDirtyChange` writes `_sectionDirty = e.detail.dirty` with no idea which editor spoke. Its detail carries only `{ dirty }`, so an identity guard needs a contract change, adding `node` to `SectionDirtyChangeDetail` and guarding on `this._activeSection === e.detail.node` like the unmount handler.

The exposed shape is narrow: a late `dirty-change` from a previous editor (an in-flight upsert settling `dirty=false` after a same-kind element retarget, or a re-armed sibling timer) can overwrite the incoming section's freshly announced dirty state, briefly disarming or arming the page's unsaved-changes guard against the wrong section. A detached emitter is harmless (the dispatch bubbles nowhere), which is why #1484 deliberately left this one on the host.

Contract change touches `SectionEditorEventMap`, both emitters' `fireSectionEvent(..., "dirty-change", ...)` call sites, and the page handler; the payload-contract tests pin the shape.

bdraco:

The race in #1486 is temporal: an older apply's settle cleared the brief window dirty flag while a newer keystroke — possibly for a sibling the reused editor element was re pointed at — was still sitting in the 200ms debounce, disarming the page's unsaved changes guard against edits that came after it. The fix is at the source: `autoApply`'s finally now clears the dirty flag only when no debounce is armed, so a stale settle cannot speak for newer edits; their own settle clears the flag. This also removes the same section fast typing flicker, which had the identical shape. An engine test pins it (keystroke mid round trip, the stale settle leaves dirty true, the newer edit's settle clears it).

As hardening, `dirty-change` also joins the identity guarded lifecycle trio: `SectionDirtyChangeDetail` carries `node`, both emitters pass their host, and the page ignores a flip whose emitter is not the active section. Node identity cannot catch the reused element case above (the emitter is the same element), but it closes the no active window between unmount and mount — a flip dropped there is re synced by the next mount from `node.dirty`, which a page test pins along with the stale emitter case.

- fixes https://github.com/esphome/device-builder-frontend/issues/1486

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.


bdraco:
@esphbot review

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Identity guard can't catch the retarget case it's built for — emitter node is the *same* element

bdraco:
You were right about the guard, and thank you for tracing it properly — node identity cannot filter a reused element, and the reachable race is temporal, not spatial: an older apply's settle clears the brief-window flag while a newer keystroke (possibly the retargeted sibling's) is still in the debounce. Addressed at 6462cd9f (rebased over the branch update):

1. The real fix landed at the source per your second path: `autoApply`'s finally now clears the dirty flag only when no debounce is armed, so a stale settle cannot disarm edits that came after it — same-section fast typing included, which had the same flicker. An engine test pins it (keystroke mid round trip, settle leaves dirty true, the newer edit's own settle clears it).

2. The node contract and page guard stay, reframed as hardening: they close the no-active window and any future distinct-element emitter, and the null-active branch is now pinned (flip dropped between unmount and mount, next mount re-syncs from `node.dirty`).

3. The section-config emitter's payload is pinned symmetrically in a new suite.

4. The guard test uses the typed-view pattern from device-escape-leave and the inert eslint-disable is gone.

Kept `fixes #1486` since the described race is now genuinely closed by (1). Suites 35/35, full run 6428 green.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Body updated post merge per the review note so the record describes the real fix (the settle/debounce gate) with the node guard framed as hardening — the squash message predates it, so the PR page is the authoritative description.

Issue #1451, opened 2026-07-25
Delete can race an in-flight auto-apply upsert and resurrect the deleted section


bdraco:
`AutoApplyController.delete()` guards on `_deleting` but not on `_applyInFlight`. If the user confirms a delete while a debounced auto apply upsert is still round tripping, the delete computes and writes its YAML, then the late upsert resolution dispatches `yaml-draft` with a buffer computed against the pre delete YAML, which can resurrect the just deleted section in the page's draft state.

The editors' delete buttons only disable on `engine.deleting`, so nothing blocks entering this window. This predates #1447 and #1449 (the same `disabled = this._engine.deleting` shipped before both); flagged by Copilot during #1449 review and deferred because that PR is a behavior free refactor.

Two candidate fixes, probably combined:

1. In `AutoApplyController.delete()`, treat an in flight apply as a barrier: either await its settlement before computing the delete diff, or invalidate its completion (sequence token) so a late `yaml-draft` from a superseded apply is dropped.
2. Optionally widen the footer's disabled window to `engine.inFlightWrite` for belt and braces, accepting the brief flicker during each debounce flush.

Fix 1 is the load bearing one; the button disable alone still races the gap between click and dispatch.

PR #1494, opened 2026-07-26
Replace flushPending's 20ms poll with a settled promise on ApplyPhase

### LINKED ISSUE #1481: Replace flushPending's 20ms poll with a settled promise owned by ApplyPhase
Retrospective on #1474. `flushPending()` settles on a 20ms `setTimeout` poll over `_apply.kind`. The poll is simple and its macrotask granularity is load bearing (the `delete()` comment documents why a microtask resolution would reintroduce #1451), but it is still a poll: a settle can be observed up to 20ms late, every waiter spins its own timer, and the termination argument lives in a comment rather than the types.

The #1469 `ApplyPhase` union makes an explicit completion signal cheap: the `applying` variant can own a `settled` promise created at entry and resolved in the finally, after the queued re run decision, so a waiter chains `await phase.settled` per round instead of polling. The macrotask requirement stays explicit as a single `await new Promise((r) => setTimeout(r))` hop after the chain drains, with the #1451 rationale attached to that one line instead of to poll timing.

Net effect: no timers while idle, settlement observed at the actual boundary, and the "resolves only once nothing is pending" contract enforced by construction rather than by loop shape. Behavior change is timing only; the engine suite's ordering tests (draft before updated, the queued re run order arrays) are the gate. Not urgent, the poll is correct today, but worth doing next time the controller is open.

bdraco:

`flushPending()` settled on a 20ms `setTimeout` poll over the apply phase: a waiter observing an in-flight upsert it did not start could be released up to 20ms after the actual settle, every waiter spun its own timer, and the termination story lived in comments around the poll timing.

The `applying` phase variant now owns its completion signal: a `settled` promise created at entry and resolved in the finally, after the queued re-run decision. Because the queued re-run's phase is installed synchronously before the old promise resolves, a waiter woken by the settle re-checks against the new round and cannot slip through a gap. Waiters observe each round at its actual boundary and idle costs zero timers.

The load bearing macrotask requirement stays, now as one explicit zero delay hop after each settled wait: nested Lit update cycles carry the settling upsert's `yaml-draft` back into `.yaml` on microtasks, so resolving straight off the settle would release the caller against a stale buffer (the #1451 regression the old poll's timing prevented implicitly). The rationale now sits on that one line, and `delete()`'s comment points at it.

Timing is the only behavior change; the engine suite's ordering cells (draft before updated, queued re-run order arrays) gate it unchanged. A new cell pins the boundary itself: a waiter parked on an in-flight upsert stays parked through a full second of timer advance, then releases on a single zero delay hop after the settle, where a poll would still be asleep in its interval.

- fixes https://github.com/esphome/device-builder-frontend/issues/1481

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/automation-editor/auto-apply-controller.ts
+++ b/src/components/device/automation-editor/auto-apply-controller.ts
@@ -67,8 +67,12 @@ export interface AutoApplyOptions {
  *  is now one value. The debounce timer stays a separate handle:
  *  typing during an in-flight round trip legitimately arms it
  *  while applying, and teardown cancels it without touching the
- *  in-flight call. */
-type ApplyPhase = { kind: "idle" } | { kind: "applying"; queued: boolean };
+ *  in-flight call. ``settled`` resolves in the finally, after the
+ *  queued re-run decision (its phase is installed first), so a
+ *  waiter observes each round at its actual boundary. */
+type ApplyPhase =
+  | { kind: "idle" }
+  | { kind: "applying"; queued: boolean; settled: Promise<void>; settle(): void };
 
 /** A delete owning the section. Carrying the suppression record
  *  inside the mode means it cannot exist without a delete and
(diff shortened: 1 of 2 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Both taken in the new commit:

1. The live descriptions dropped the poll vocabulary: the `flushPending` doc now says "armed mid-wait", the settle-wait test renamed to "already settle-waiting" with its comments updated, and the delete-ordering mock's comment says "the flush's macrotask hop". The new cell's own "not a poll tick" phrasing stays as the deliberate contrast with history, and the `AUTO_APPLY_DEBOUNCE_MS + 20` advances are left alone; they are generous margins now rather than poll-sized, and changing timing values in passing buys nothing.
2. The hop comment now states the invariant's scope: only the settled branch hops, and `delete()` relies on entering with `_debounce === null` so its wait lands there.

bdraco:
@esphbot review

bdraco:
The follow-up commit raced the merge, the merge carried only 2bc1d1a8. The two comment nits land unchanged via https://github.com/esphome/device-builder-frontend/pull/1495.

Issue #1479, opened 2026-07-25
Evaluate anchor-dispatched drafts to make the section switch synchronous again


bdraco:
Retrospective on the #1468/#1477 arc. The section switch guard now carries five interacting mechanisms: the awaited flush barrier, the `_switchSeq` supersede token, the `compose` exemption for Back, the `isConnected` bail, and post flush line re resolution (`resolveCurrentSectionLine` at three call sites). Each is individually justified and tested, but all five exist to serve one root requirement: the switch must not unmount an automation editor while its upsert is in flight, because the editor's `yaml-draft` dispatches from a detached element and never reaches the page.

#1475 built the tool that dissolves that root requirement: `fireFromAnchor` dispatches through the mount time parent when the host is detached, and it already covers all three disk writing delete paths. If `AutoApplyController.autoApply` dispatched `yaml-draft` through the same anchor, a mid upsert unmount would no longer lose the draft, and `_guardSectionSwitch` could return to a synchronous swap. That deletes the barrier, the token, `compose`, the `isConnected` bail, the post flush re resolution (the coordinates would be read synchronously again), and closes #1478 (no wait, no affordance needed) as designed out rather than mitigated.

What has to be verified before committing to it:

1. The draft diff basis. `yaml-draft` carries `applyYamlDiff(host.yaml, diff)` where `host.yaml` is the detached element's stale prop. If the page's `_yaml` advanced after the unmount (the new section's own draft), applying the late draft clobbers it, the same class as #1476. The dispatch likely needs to carry its basis (`basedOn`) so the page can drop or re apply a draft computed against a superseded buffer.
2. The global save's flush stays. `_saveYaml` must still await `flushPending()`; only navigation stops waiting.
3. The suppression and re arm choreography in `delete()` assumed navigation waits; the mid flush retarget tests in the controller suite pin what must keep working.

If 1 proves tractable, the net deletion is large (the guard shrinks to `action()` plus the dedupe returns) and a whole class of deferral window bugs (#1470, #1478, the double click and caret yank cases) becomes structurally impossible instead of individually guarded. If it does not, this issue documents why the barrier is the chosen design.

bdraco:
[...]

PR #1496, opened 2026-07-26
Carry the emitter's configuration on yaml-draft and identity-guard it

bdraco:

First step of the #1479 arc, and a standing hazard fix on its own: `yaml-draft` had no device identity, so a draft resolving after a device switch spliced the previous device's section into the new device's buffer. The router reuses one page element across device routes, which is the same cross device window #1491 closed for `yaml-updated`; the draft path was still open.

`yaml-draft` now has a typed detail in the section editor contract, `YamlDraftDetail { configuration, yaml }`, wired through `SectionEditorEventMap` and the global `HTMLElementEventMap` so every firer type checks. All five emitters (the auto apply engine, the component editor's debounced splice, the add component dialog, the api action dialog, and the shared add automation and add script dispatch helper) carry their snapshotted target, and the page's `_onYamlDraft` drops a mismatched announcement before touching the buffer.

The `basedOn` basis extension the sync switch needs is deliberately not in this PR; it lands with the consumer that gives it semantics.

- part of https://github.com/esphome/device-builder-frontend/issues/1479

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/add-api-action-dialog.ts
+++ b/src/components/device/add-api-action-dialog.ts
@@ -31,6 +31,7 @@ import { renderMarkdown } from "../../util/markdown.js";
 import { registerMdiIcons } from "../../util/register-icons.js";
 import { parseYamlAutomations } from "../../util/yaml-sections.js";
 import { applyYamlDiff, sectionKeyFromLocation } from "./automation-editor/serialise.js";
+import { fireSectionEvent } from "./section-editor.js";
 
 import "@home-assistant/webawesome/dist/components/icon/icon.js";
 import "../base-dialog.js";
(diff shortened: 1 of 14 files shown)


bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Post-await read of this.configuration defeats the new guard here
> - Same post-await live read as the add-automation dialog
> - _dispatchDraft reads live this.configuration; every call site is post-await
> - Import placement fails format:check (organize-imports moves it)

bdraco:
Findings 1 to 3 taken in 81f1e3e5; finding 4 declined with evidence.

1. and 2. Both wizard dialogs now hoist `configuration` and `yaml` above the upsert await and pass the locals through to `dispatchAutomationAdded`, mirroring the api-action dialog, with the snapshot rationale in the comment. This also closes the same-device stale-basis half you noted.
3. `_dispatchDraft` now takes `configuration` as a parameter so the pairing is type-checked; the bundle flow snapshots once after its guard and threads it through the `addComponent` calls and all five dispatch sites, and `_submitComponent` does the same for its flow.
4. The import placement is a false positive from the off-pin prettier your caveat anticipated: this worktree runs the repo's pinned toolchain (prettier 3.9.6 via pnpm), and both the pre-commit hook's `format:check` at commit time and a fresh `pnpm run format:check` now pass with the import where it sits. No change made.

On the description's over-claim: it said all five emitters carry their snapshotted target while three read live props post-await; with this push the claim is true. Thanks for the prop-chain verification on the flushDraft path.

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Suggestion 1 taken in the new commit: `add-dialog-target-snapshot.test.ts` drives all three dialogs' private submit handlers with a deferred backend stub, reassigns `configuration` and `yaml` mid round trip, and asserts the emitted detail keeps the pre-switch target and (for the wizards) a basis derived from the pre-switch buffer. Inlining any of the snapshots back to a live prop read now fails these cells instead of passing silently.

The `automation-added` sibling window is noted for the next step of the #1479 arc, which touches the same dispatch path when drafts move to anchor dispatch. Thanks for the plainly stated retraction on finding 4.

bdraco:
@esphbot review

bdraco:
@esphbot review

PR #1497, opened 2026-07-26
Anchor-dispatch auto-apply drafts and carry their basis

bdraco:

Second step of the #1479 arc: an automation editor unmounted mid upsert no longer loses its draft. Today the dispatch fires from the detached element and bubbles nowhere, which is the root requirement the whole section switch barrier serves; Back's composed switch is the live loss window.

`prepareYamlUpdated` is generalized into `prepareSectionEvent(host, name)` (the three delete emitters migrate mechanically), and the auto apply engine captures a `yaml-draft` announcer before its await, so the draft rides the mount time anchor when the element is gone. `YamlDraftDetail` gains `basedOn` (the buffer the draft was computed against) and `node` (the emitter, the #1489 identity pattern), and the page's `_onYamlDraft` now applies the active section's drafts regardless (its yaml prop legitimately lags its own last draft by a render), applies a late anchored draft whose basis still matches, and drops a late draft whose basis the pane moved past with an "Edit discarded" toast instead of silently clobbering the newer edit. Dropping rather than recomputing is deliberate: for the anchored automation case today's behavior loses the draft silently in every case, so conservative visibility is a strict improvement there. For the dialog emitters (never the active section) the same guard is a trade rather than a strict win: pre-#1497 their merge always landed and silently clobbered any concurrent edit; now the rarer moved-buffer case drops the add with the toast instead. The recompute can be layered on later if the toast shows up in practice.

The engine also reads its buffer exactly once now (the upsert argument and the diff application shared two separate live reads, so a YAML pane keystroke during the round trip could mangle the splice), and the other emitters carry their basis from the snapshots #1496 introduced.

Behavior is otherwise unchanged and the switch barrier stays; removing it is the arc's next step once this holds.

- part of https://github.com/esphome/device-builder-frontend/issues/1479

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
All four taken in b613ce0d:

1. The description now names the dialog trade honestly (their merge previously always landed and silently clobbered the concurrent edit; now the moved-buffer case drops the add visibly), the copy is artifact neutral ("Change not applied" / "It finished after the buffer changed and no longer applies. Redo it if needed."), and a dialog-labeled page cell pins the newly gated path.
2. The queued re-run is skipped when the host is disconnected, with the debounce-cancel rationale extended to it; a new cell pins that an unmount with a queued re-run delivers exactly one anchored draft and issues no second upsert.
3. The switch-barrier doc now says the anchor delivers the late draft and that the await's remaining value is a matching basis, so the arc's next step reads the true justification.
4. `_dispatchDraft` takes a single object, removing the positional string hazard.

bdraco:
@esphbot review

PR #1507, opened 2026-07-26
Replace source-text assertions with behavioral coverage

### LINKED ISSUE #1504: Replace source-text assertions with behavioral coverage in two suites
Flagged in the #1502 review: `test/util/section-entry-overrides.test.ts` and `test/components/device/automation-editor/catalog-picker-dialog.test.ts` read component source files from disk with `fs.readFileSync` and assert on the source text. Those assertions pass whenever the string is present and fail on an innocuous rename or reformat, while never exercising that the wiring works, which is the opposite of the coverage they appear to give.

Replace them with behavioral assertions: mount the component (or call the resolver) and check the `.entries` / picker contents the tests claim to be about.

bdraco:

Two suites asserted on component source text read from disk with `fs.readFileSync`, which passes on string presence and fails on innocuous renames while never exercising the wiring. Both are now behavioral.

`catalog-picker-dialog.test.ts` mounts the real dialog under happy-dom and drives it through its rendered surface: search typed into the actual input filters across id, name, and description case insensitively; "By type" groups `switch.template` and `switch.gpio` under one `switch` header and skips core items; "Building blocks" lists only core items; clicking a row on "By target" emits `catalog-picked` with the id param pre filled for a declared id device and left empty for a synthesized one (#2208); the tab strip drops "By target" for conditions. The `open()` default tab contract already lives in the sibling open contract suite.

The two source scans in `section-entry-overrides.test.ts` moved to a new `section-config-entries-wiring.test.ts` that mounts `device-section-config` and pins the #160 regression class directly: the form's `.entries` prop is the resolver's output (a MAP entry for substitutions), not the raw catalog, and `flushDraft` validates the user keyed values against that same resolver schema, so a bogus flat catalog produces no phantom required errors while pass through sections still enforce theirs.

- fixes https://github.com/esphome/device-builder-frontend/issues/1504

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/device-section-config/draft-and-delete.ts
+++ b/src/components/device/device-section-config/draft-and-delete.ts
@@ -20,6 +20,10 @@ import { fireSectionEvent, prepareSectionEvent } from "../section-editor.js";
 // catalog. MAP_SECTIONS (substitutions / packages) carry an irrelevant flat
 // catalog schema that doesn't match what the user actually edits in the form —
 // using it would surface phantom "missing required" errors per keystroke.
+// Pre-save backend lint (validateYaml) is deliberately not wired in here: the
+// YAML pane's squiggles (yaml-lint-backend.ts) provide the same lint
+// continuously, and the explicit Validate button runs the full ESPHome
+// compile against the saved file.
 export function flushDraft(host: ESPHomeDeviceSectionConfig): string | null {
   host._draftTimer = null;
   if (!host._config) return null;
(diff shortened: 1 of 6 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
All four taken in e0677fe3. mountDialog now switches tabs by clicking the rendered .picker-tab button (the @click wiring is covered and one as any is gone), pickFirstRow is sync with the assertion moved to each cell so a silent non-pick fails with the cell's own diff, the open contract sibling's docstring describes the new division of labour instead of the vanished node-env scan, and the no-pre-save-lint architecture note now lives in draft-and-delete.ts's flushDraft header where the question would come up. Chained #1508 rebased on top.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Both remaining nits are in 4582e6aa: the two stale _onSave references now name flushDraft as the section path's validation call site, and the tab helper throws a named error when a kind/tab pair has no rendered button instead of failing on undefined. (Push was delayed by a local signing outage.)

@esphbot review

Issue #1480, opened 2026-07-25
Adopt using-based disposables for paired acquire/release state in async flows


bdraco:
Retrospective on the recent async work. Several flows now hold manually paired acquire/release state across awaits, and the release side lives in a `finally` (or is implicit) at some distance from the acquire:

- `device-labels-editor.ts`: `_pendingSaves++` before the chained task, `--` in its finally. The pairing is load bearing enough that a review round added a comment plus a dedicated test warning that resetting the counter in the swap branch would drive it negative and permanently disable the same device clear.
- `auto-apply-controller.ts`: `_setDeleteMode(mode)` at delete start, `_setDeleteMode(null)` in the finally; `_apply = { kind: "applying" }` at entry, `{ kind: "idle" }` in the finally.
- `device-section-config.ts` `_onDeleteRow`: `_deletingRow = key` / `""` in a finally.

The pinned TypeScript supports `using` declarations (Symbol.dispose), which turn each of these into scope bound ownership the compiler enforces, the shape RAII gives C++ and `defer` gives Go:

```ts
// util/scoped.ts
export function acquire(release: () => void): Disposable {
  return { [Symbol.dispose]: release };
}

// call site
using _pending = acquire(() => this._pendingSaves--);
this._pendingSaves++;
await task;
// release runs on every exit path, no finally to forget
```

Candidates in rough order of value: the pending saves counter (the invariant a test currently guards by convention), the delete mode and apply phase transitions, and `_deletingRow`. Behavior free refactor; the existing suites (labels 11, engine 30) pin every transition. Worth checking whether vitest's transform and the pinned tsc both handle `using` before starting; if not, a `withScope(release, fn)` higher order helper gets the same pairing guarantee without the syntax.

bdraco:
Explored in #1509 and closed there. The toolchain works (tsc 7 with ESNext.Disposable lib, vitest, and the Rspack build all handle using), but the survey showed the candidate family is ~35 adjacent try/finally resets where the pairing is already clear, and the engine conversions made ordering implicit (using-in-try disposes before the finally) where it used to be the finally's explicit first line. Not easier to read or less error prone at the sites that exist, so not adopting.

PR #1509, opened 2026-07-26
Adopt using-based disposables for paired acquire/release state

bdraco:

Adopts `using` declarations for the manually paired acquire/release state the async work left behind, per the retrospective. A new `util/scoped.ts` exports `acquire(release)`, which returns a `Disposable` so `using _x = acquire(...)` runs the release on every exit path with the pairing enforced by scope instead of a distant `finally`.

Four sites convert. The labels editor's `_pendingSaves` counter (the invariant a test guards by convention) is now a scope bound hold in `_persist`, released when the call settles. The engine's apply phase and delete mode reset inside their `try` blocks, which matches the old order exactly: `using` disposes at try exit before the `finally`, so the queued re run still installs its own phase unclobbered and `_settleDelete` still re arms with the delete mode already released. The section editor's `_deletingRow` lock releases on every exit of `_onDeleteRow`, and its `finally` is gone.

Toolchain was verified up front as the issue asked: the pinned tsc 7.0.2 needs `ESNext.Disposable` added to the tsconfig lib (included here), vitest's transform lowers `using` fine, and the production build compiles. `scoped.ts` seeds the `Symbol.for` fallback that TypeScript's own downlevel emit uses, so runtime and emit agree on engines without the well known symbol. A new suite pins the release on every exit path, reverse declaration order, and the dispose before finally ordering the engine conversions rely on.

Chained on #1506 because the engine sites overlap its diff.

- fixes https://github.com/esphome/device-builder-frontend/issues/1480

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/automation-editor/auto-apply-controller.ts
+++ b/src/components/device/automation-editor/auto-apply-controller.ts
@@ -9,6 +9,7 @@ import type { LocalizeFunc } from "../../../common/localize.js";
 import { fireEvent } from "../../../util/fire-event.js";
 import { formatApiError } from "../../../util/format-api-error.js";
 import { notifyError } from "../../../util/notify.js";
+import { acquire } from "../../../util/scoped.js";
 import { writeAutomationDelete } from "../apply-removal.js";
 import type { SectionEditor } from "../section-editor.js";
 import {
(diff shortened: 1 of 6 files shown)


bdraco:
@esphbot review

bdraco:
Closing after review discussion. The survey found the paired acquire/release family is real (35 finally-reset sites in src/) but almost all are adjacent try/finally resets where the pairing is already impossible to miss, so using buys nothing. The two engine conversions actually traded an explicit ordering (reset as the finally's first line) for a spec-implicit one (using-in-try disposes before the finally), which is harder to review, not easier. The one genuinely improved site (the labels counter's distant, invariant-critical pairing) is too little to carry a new idiom.

PR #1522, opened 2026-07-27
Intercept dirty Back before the router commits the route

### LINKED ISSUE #1520: Browser Back on a dirty editor leaves silently; the popstate guard never intercepts in real dispatch
With unsaved changes in the device editor (or secrets page), pressing the browser's Back button leaves the page silently: no unsaved-changes prompt, the buffer is discarded, and the dashboard renders. The in-app leave paths all work — nav links and the command palette run the guard through `navigate()`, and the header back arrow and Escape run it through `goBackOrHome()` — so the gap is specifically the raw browser Back/Forward chrome.

The page-level popstate interception (now `PopLeaveGuardController` after #1518, previously the per-page `_onPopState` handlers) is supposed to catch this: consume the rollback suppression, dirty-check, `stopImmediatePropagation`, re-push the page URL, prompt, replay on proceed. The handler logic is correct — invoking it directly intercepts and prompts — but in a real popstate dispatch it never does. Instrumented against the dev stack (probe listener registered before all app listeners sees the popstate with the page mounted and dirty; no `stopImmediatePropagation` is ever called; the router commits the popped URL and unmounts the page). Verified byte-identical behavior on merged `main` (dd6e320f) and on the #1518/#1519 refactor branches, so this predates the refactor and was not introduced by it.

Likely mechanism: `popstate` targets `window`, and at-target dispatch runs listeners in registration order regardless of the capture flag. The app shell (and its router controller) connects before any routed page, so the router's listener runs first and its `goto()` commits the popped route; the page's capture-phase listener never gets to veto in time. If that holds, the fix belongs in the interception design rather than the pages — e.g. the centralized guard (`PopLeaveGuardController` in `src/util/navigation.ts`) could be armed by the shell before the router registers, or the router's popstate handling could consult `runLeaveGuard` the way `navigate()` and `goBackOrHome()` already do.

Repro: dev stack, open a device config, type anything in the YAML editor, press browser Back. Expected: unsaved-changes prompt with the URL held. Actual: dashboard, edit gone. `beforeunload` still guards tab close, so the loss window is SPA back/forward only.

Related: the interception logic is pinned by `test/util/pop-leave-guard-controller.test.ts` and `test/pages/device-escape-leave.test.ts`, but both drive the handler directly; no test pins the real dispatch ordering against the router, which is why this survived.

bdraco:

With unsaved changes, pressing the browser's Back button left the page silently: no prompt, buffer discarded, dashboard rendered. The in app leave paths were fine because they run the guard explicitly; the raw browser chrome path relied on the popstate interception, and that interception was structurally dead, since before #1518 too.

The mechanism, verified against the shipped `@lit-labs/router` source and live probes: popstate targets `window`, and for a window targeted event every listener runs in registration order, the capture flag confers no priority. The router registers its popstate listener at app shell connect, first in the whole app, and its handler calls `goto(location.pathname)` synchronously; the dashboard route has no `enter`, so the route commit happens inside the router's handler before the page guard's listener (registered last, at page connect) ever runs. `stopImmediatePropagation` cannot suppress a listener that already ran. Three comments in the tree already described this exact ordering as the reason the in app paths pre run the guard.

The fix registers one interception listener ahead of the router: `installLeaveGuardInterceptor()` in `navigation.ts` adds a single popstate listener that delegates to the active `PopLeaveGuardController`, and `createRouter()` installs it before constructing the `Router`, so the ordering holds by construction. The controller stops adding its own listener and instead claims and releases delivery on connect and disconnect, with the same ownership check the guard registry uses. The handler itself is unchanged; its `stopImmediatePropagation` now genuinely blocks the router, so a dirty Back holds the URL, prompts, and replays on proceed, and the page no longer unmounts mid confirm. `goBackOrHome` keeps its prompt before pop path (better UX than intercept then re push); the comments that documented the dead veto are rewritten to describe the working design.

New suite coverage closes the gap that let this stay green: a router style bubble listener registered after the interceptor is never called on a dirty pop and is called on allowed, suppressed, and clean pops; interceptor installation is idempotent; delivery follows the active controller across page swaps. Verified live against the dev stack with a real Edit click navigation and an in page `history.back()`: before the fix the probe shows the silent leave, after it the URL holds with the prompt pending, Cancel stays, and Discard replays to the dashboard without a bounce. The in app back arrow flow is unchanged.

- fixes #1520

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/app-shell/router.ts
+++ b/src/components/app-shell/router.ts
@@ -141,9 +141,7 @@ export function createRouter(
   host: ReactiveControllerHost & HTMLElement,
   hooks: RouterHooks
 ): Router {
-  // Before the Router exists: its popstate listener must register after
-  // the leave-guard interceptor's, or a dirty Back commits the route
-  // change ahead of the guard's veto (#1520).
+  // Ordering is load-bearing; see installLeaveGuardInterceptor's doc.
   installLeaveGuardInterceptor();
   return new Router(host, [
     {
--- a/src/components/app-shell/router.ts
+++ b/src/components/app-shell/router.ts
@@ -3,6 +3,7 @@ import { html, type ReactiveControllerHost } from "lit";
 import type { LocalizeFunc } from "../../common/localize.js";
 import { withBase } from "../../util/base-path.js";
 import {
+  installLeaveGuardInterceptor,
   isFreshNavigatePush,
   isSameDocumentPush,
   popPushedEntrySilently,
@@ -140,6 +141,10 @@ export function createRouter(
   host: ReactiveControllerHost & HTMLElement,
   hooks: RouterHooks
 ): Router {
+  // Before the Router exists: its popstate listener must register after
+  // the leave-guard interceptor's, or a dirty Back commits the route
+  // change ahead of the guard's veto (#1520).
+  installLeaveGuardInterceptor();
   return new Router(host, [
     {
       path: withBase("/"),
--- a/src/pages/device.ts
+++ b/src/pages/device.ts
@@ -633,10 +633,9 @@ export class ESPHomePageDevice extends LitElement {
       return;
     }
     /* Otherwise leave the editor via the header back arrow's guarded
-       path. A raw history.back() here loses the dirty buffer: the
-       router's popstate listener commits the leave before this page's
-       ``_onPopState`` can veto it, so the unsaved-changes prompt must
-       run before the navigation, not after. */
+       path, which prompts before the pop — nicer than a raw
+       history.back(), which the popstate interceptor would catch only
+       after the URL changed, via re-push. */
     e.preventDefault();
     void goBackOrHome();
   };
--- a/src/pages/device.ts
+++ b/src/pages/device.ts
@@ -632,10 +632,7 @@ export class ESPHomePageDevice extends LitElement {
       this._drawerOpen = false;
       return;
     }
-    /* Otherwise leave the editor via the header back arrow's guarded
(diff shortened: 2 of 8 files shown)


bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
All three nits applied in bd45106c.

1. The idempotency cell now uses a clean pop, so a duplicated listener genuinely fails it (nothing stops propagation on that path).
2. New cell "releases delivery when the last controller disconnects" pins the stale-slot hazard you described — an unmounted page can no longer veto dashboard pops.
3. The three test-file headers asserting the dead-veto ordering as current fact are rewritten; the #2259 reference now dates itself to the pre-interceptor era.

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Both applied.

1. New `test/components/app-shell/router-interceptor.test.ts` pins the production wiring: a real `createRouter` + fake shell, Router listener registered at connect, then a dirty pop asserts `router.goto` was never consulted while a clean pop falls through. The guard suite self-installs, so this file alone catches the `createRouter` install call going missing.

2. The third dead-veto header (device-escape-leave) is reframed to the pre-interceptor era — you were right that my "all three" claim was off by one; the grep I used matched a different phrasing.

bdraco:
@esphbot review

PR #1526, opened 2026-07-28
Gate the caret-follow reveal on the whole render unit

### LINKED ISSUE #2408: [Bug] "Show advanced settings" still flips on for empty members of an already-inline advanced cluster
### What happened?

Follow-up from the review of esphome/device-builder-frontend#1525 (which fixed #2398 for plain leaves). The same over-reveal survives one shape narrower: exclusive groups and constraint clusters.

Under `gateAdvanced` the form's placement rule (`unitPrefilled`, `src/components/device/config-entry-form.ts`) treats a whole exclusive group / constraint cluster as inline when any member has a value, and `_advancedUnitClassifier` only puts a group in the advanced bucket when every member is advanced. The caret-follow reveal predicate (`pathIsAdvanced`, `src/util/config-entry-tree.ts`) checks only the target entry's own material value.

So for an all-advanced cluster with member A filled and member B empty, the unit already paints inline atomically, yet clicking B's YAML line still flips "Show advanced settings" on, surfacing every unrelated empty advanced field.

Suggested direction from the review: resolve the focus target's unit (via `buildFormRenderPlan`'s `clusterByFirstKey` / `memberKeys`) before the value check, so `pathIsAdvanced` and `unitPrefilled` cannot disagree. Both reveal call sites are affected (`device-section-config/reveal.ts` and `config-entry-form.ts` `_maybeRevealForFocus`).

### Dashboard version

1.7.1b1 (main after device-builder-frontend#1525)

### How did you install it?

Dev checkout

### Where is it running?

Local

### Anything else?

bdraco:

Clicking the YAML line of an empty advanced field inside an exclusive group or constraint cluster still flipped "Show advanced settings" on, even though the whole unit already paints inline when any member has a value. The reveal predicate judged the target entry alone while the form places units atomically, so the two could disagree.

The plan module now owns the unit rule once: `unitAllAdvanced` and `unitHasMaterialValue` are the shared halves, `unitAdvancedGate` composes them into a per-key verdict, and the form's `unitPrefilled` and `_advancedUnitClassifier` are rewired onto the same halves. `pathIsAdvanced` takes the gate as an injected predicate (avoiding a util-to-components cycle) and both reveal paths pass it; when the top-level key belongs to a unit, the gate's verdict replaces the per-entry rule. A wholly empty all-advanced unit still reveals so the group's control surfaces, and mixed units, which always render in the basic bucket, no longer trigger a reveal either.

- fixes https://github.com/esphome/device-builder/issues/2408

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/config-entry-form-plan.ts
+++ b/src/components/device/config-entry-form-plan.ts
@@ -49,6 +49,31 @@ export function buildFormRenderPlan(
 }
 
 /**
+ * Resolver from a top-level entry key to its render unit's members (exclusive
+ * group or constraint cluster), undefined for plain entries. Feeds
+ * ``pathIsAdvanced`` so reveal decisions match unit placement.
+ */
+export function unitMembersByKey(
+  entries: ConfigEntry[],
+  requiredGroups: RequiredGroup[]
+): (key: string) => ConfigEntry[] | undefined {
+  const byKey = new Map<string, ConfigEntry[]>();
+  const { clusters } = buildConstraintClusters(entries, requiredGroups);
+  for (const cluster of clusters) {
+    for (const member of cluster.members) byKey.set(member.key, cluster.members);
+  }
+  const groups = new Map<string, ConfigEntry[]>();
+  for (const entry of entries) {
+    if (!entry.exclusive_group) continue;
+    const members = groups.get(entry.exclusive_group) ?? [];
+    members.push(entry);
+    groups.set(entry.exclusive_group, members);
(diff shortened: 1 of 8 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1530, opened 2026-07-29
Move hasMaterialValue into the util layer

### LINKED ISSUE #2409: Move hasMaterialValue into the frontend util layer
Follow-up from the review of esphome/device-builder-frontend#1525.

`src/util/config-entry-tree.ts` now takes a runtime (value) import of `hasMaterialValue` from `src/components/device/config-entry-render-filter.ts`, giving the util layer a runtime edge into the component layer. Existing util-to-components imports are predominantly `import type`, so this is the first real runtime inversion. No import cycle exists today; this is a maintainability note, not a defect.

`hasMaterialValue` is DOM-free (depends only on `api/types`, `util/nested-values`, `util/yaml-serialize`), so it can move into `src/util/` with `config-entry-render-filter.ts` re-exporting it so its existing consumers don't churn.

bdraco:

`hasMaterialValue` is a DOM-free predicate that lived in the component layer, and `src/util/config-entry-tree.ts` importing it gave the util layer a runtime edge into components. It now lives in `src/util/material-value.ts`; `config-entry-render-filter.ts` re-exports it so its existing consumers don't churn, and `config-entry-tree.ts` imports it from util. Function body and doc comment move verbatim, no behavior change.

- fixes https://github.com/esphome/device-builder/issues/2409

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/config-entry-render-filter.ts
+++ b/src/components/device/config-entry-render-filter.ts
@@ -25,8 +25,10 @@ import type { BoardCatalogEntry } from "../../api/types/boards.js";
 import type { ConfigEntry } from "../../api/types/config-entries.js";
 import { ConfigEntryType } from "../../api/types/config-entries.js";
 import { isEntryVisible } from "../../util/config-validation.js";
-import { asMappingList, asRecord, isPlainObject } from "../../util/nested-values.js";
-import { YamlRawValue } from "../../util/yaml-serialize.js";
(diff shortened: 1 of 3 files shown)


bdraco:
@esphbot review

PR #1533, opened 2026-07-29
Adopt under the factory broadcast name and auto-start the rename flow

### LINKED ISSUE #2412: Renamed adopt can strand the device: first OTA targets a hostname the factory firmware never broadcasts
## The gap

The Take control dialog lets the user edit the device name (hostname) during adoption, but adoption only writes the YAML; nothing renames or reflashes the running device. The factory firmware keeps broadcasting its MAC-suffixed name until the first flash bakes in the new one, so a renamed adopt's first OTA has to reach a device that does not answer to the configured hostname.

## What bridges it today, and where it breaks

At adopt, the cross-name probe (`probe_device(name, service_name=mdns_name)` in `controllers/devices/importable.py`) resolves the factory service live and applies its addresses against the chosen name; `on_ip_change` persists the primary to `device.ip` in the metadata sidecar. At first install, `_build_address_cache_args` misses the mDNS cache for the new name, falls back to `device.ip`, and the upload runs with `--mdns-address-cache <new-name>.local=<factory-ip>`, so espota2 flashes the device at its real IP. After reboot the device announces under the new name and everything converges.

That bridge has three holes, and in each the renamed device is stranded until a USB flash (the new name is not broadcast yet, so there is no mDNS self-heal):

1. **Device asleep or un-resolvable at adopt time.** A deep-sleep device adopted from a stale discovery banner entry (A records carry a 120s TTL) never gets an IP persisted; the first OTA resolves `<new-name>.local` and fails.
2. **DHCP re-lease between adopt and first install.** The persisted IP goes stale and the upload targets the wrong host. The device is still broadcasting its factory name the whole time, but that name was dropped with the `import_result` row at adopt.
3. **First-attempt race on `loaded_integrations`.** `get_address_cache_args` gates on "api"/"web_server" in `loaded_integrations`, which is compile-derived; the chained upload can start before the post-compile device refresh lands, so the very first install of a renamed adopt may get no cache args at all. A retry works, but the first impression is a failed flash.

## Options

- **Persist the factory broadcast name at adopt** (metadata sidecar field, cleared on the first successful flash) and have the install path resolve it live at flash time. The factory firmware keeps broadcasting that name right up until it is reflashed, so this closes holes 1 and 2 completely, backend-only, no UX change. Preferred.
- **Drop the rename field from the adopt dialog** (frontend) and route renames through the existing rename flow, which compiles and flashes targeting the old identity (`esphome rename` handles exactly this).
- **Auto-run the rename flow when the dialog name differs from the broadcast** — robust, but changes adoption from "write YAML" to "write YAML plus immediate flash".

Related: #2389 fixed the reachability side of the renamed adopt (the un-withdrawable mdns claim); this issue is the flash-targeting side of the same scenario.

bdraco:

The Take control dialog lets the user edit the device hostname, but adoption only wrote a YAML under the chosen name — nothing renamed the running device. The factory firmware keeps broadcasting its MAC-suffixed name until first flash, so a renamed adopt's first OTA targeted a hostname nobody answers to and could strand the device (asleep at adopt, DHCP re-lease, or the cache-args race fixed backend-side in esphome/device-builder#2417).

The dialog UX is unchanged; the orchestration behind it now composes two existing commands:

- `devices/import` is always called with the **factory broadcast name** — the only hostname the running device answers to — so the adopted config is reachable from the moment it exists.
- When the dialog name was edited, the `adopted` event carries `renameTo` and the dashboard immediately starts the existing rename flow (`devices/rename`, OTA path). Its compile + flash chain targets the factory identity, the command dialog follows the job ("Rename — old → new"), and on success the swap re-keys the card to the chosen name.
- If the rename fails at any point (device unreachable, compile error), the backend's revert keeps the device adopted under its factory name — nothing is stranded, and the user can rename later from the card. A failed rename enqueue surfaces through the existing rename-failed toast.

`performRename` now takes `(configuration, currentName)` instead of a `ConfiguredDevice` so the adopt path can invoke it before the device row exists in frontend state; both existing callers are updated. The orchestration lives in `adoptFollowUp` (actions-ui) with the payload typed once as `AdoptedDetail`, and the import response's `configuration` is carried through the event as the authoritative YAML path.

The dialog's name fields are now the shared `esphome-device-name-inputs` pair the clone and create flows use: friendly name primary, hostname behind the disclosure, with the shared validation (charset, esphome's 31-char hostname cap, edge hyphens, taken-hostname collisions from the same `takenHostnameSet`) — all failures that would otherwise surface only after the dialog closed, as a rename error on an already-adopted device. The hostname seeds to the factory broadcast verbatim (`reset` gained a `prefillHostname` seed) and derivation stays live, matching the create/rename idiom: typing a friendly name recalcs the hostname (maintainer-directed), the rename hint announces the escalation (`role="status"`), and clearing the friendly name falls back to the factory broadcast instead of stranding an empty hostname. The factory name itself is exempt from the taken set since its own importable row is in it. A `dashboard.adopt_rename_hint` line under the fields tells the user at decision time that changing the hostname builds and installs new firmware over the network.

Verified live against a Bonjour-registered fake importable, through the final shared-pair UI: the dialog opens friendly-name-first with "Hostname: athom-fake-7e11aa" collapsed behind the disclosure; editing the hostname to `kitchen-co2` shows the install hint; submit landed the YAML under the factory name and the command dialog followed the chain ("Rename — Athom Fake CO2 (athom-fake-7e11aa → kitchen-co2)") with the compile streaming. The failure path was verified end-to-end in an earlier pass: the tail flashed at the factory device's resolved address, failed against the fake device, and the backend reverted — factory YAML kept.

- fixes https://github.com/esphome/device-builder/issues/2412

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/adopt-dialog.ts
+++ b/src/components/adopt-dialog.ts
@@ -1,11 +1,12 @@
 import { consume } from "@lit/context";
 import { LitElement, css, html, nothing } from "lit";
-import { customElement, property, state } from "lit/decorators.js";
+import { customElement, property, query, state } from "lit/decorators.js";
 import type { ESPHomeAPI } from "../api/esphome-api.js";
 import type { AdoptableDevice } from "../api/types/devices.js";
 import type { LocalizeFunc } from "../common/localize.js";
 import { apiContext, localizeContext } from "../context/index.js";
 import type { AdoptedDetail } from "./dashboard/actions-ui.js";
+import type { ESPHomeDeviceNameInputs } from "./shared/device-name-inputs.js";
 import {
   dialogActionButtonStyles,
   dialogActionsRowStyles,
@@ -13,21 +14,18 @@ import {
 import { dialogChromeStyles } from "../styles/dialog-chrome.js";
 import { inputStyles } from "../styles/inputs.js";
 import { espHomeStyles } from "../styles/shared.js";
-import type { ValidationError } from "../util/config-validation.js";
-import { validateDeviceName } from "../util/config-validation.js";
 import { DialogOpenController } from "../util/dialog-open-controller.js";
 import { EnterController } from "../util/enter-controller.js";
 import { fireEvent } from "../util/fire-event.js";
 import { formatApiError } from "../util/format-api-error.js";
 import { markJustCreated } from "../util/just-created.js";
 import { previewPackageImportUrl } from "../util/package-import-url.js";
-import { renderInlineError } from "../util/render-error.js";
 import { fetchSecretKeys, hasSharedWifiSecret } from "../util/secrets-cache.js";
-import { HOSTNAME_MAX_LEN } from "../util/slugify-hostname.js";
 import { wifiFieldsStyles } from "./onboarding/wifi-fields-styles.js";
 import { isWifiPasswordTooShort, renderWifiFields } from "./onboarding/wifi-fields.js";
 
 import "./base-dialog.js";
+import "./shared/device-name-inputs.js";
 
 @customElement("esphome-adopt-dialog")
 export class ESPHomeAdoptDialog extends LitElement {
@@ -44,9 +42,10 @@ export class ESPHomeAdoptDialog extends LitElement {
   // rename step, after the import already landed.
   @property({ attribute: false }) takenHostnames: ReadonlySet<string> = new Set();
 
+  @query("esphome-device-name-inputs")
+  private _nameInputs?: ESPHomeDeviceNameInputs;
+
   @state() private _device: AdoptableDevice | null = null;
-  @state() private _name = "";
-  @state() private _friendlyName = "";
   // Default on — the legacy dashboard added an encryption key to
   // every adopted device unconditionally and the lack of a way to
   // opt out was the actual annoyance. Keep the secure-by-default
@@ -235,15 +234,6 @@ export class ESPHomeAdoptDialog extends LitElement {
 
   open(device: AdoptableDevice) {
     this._device = device;
-    /* Default to the discovered hostname verbatim — including the
-       MAC-suffix factory firmware appends. The import always lands
-       under this broadcast name (the only hostname the running
-       device answers to); an edited name is applied afterwards via
-       the rename flow, which flashes the device before the old name
-       is dropped. Defaulting to a stripped form silently dropped
-       the disambiguator on devices like ``apollo-plt-1-983300``. */
-    this._name = device.name;
-    this._friendlyName = device.friendly_name || "";
     this._encryption = true;
     this._busy = false;
     this._error = null;
@@ -251,6 +241,18 @@ export class ESPHomeAdoptDialog extends LitElement {
     this._password = "";
     this._hasWifiSecrets = undefined;
     this._dialog.open = true;
+    /* Seed the shared name pair after the open render mounts it.
+       The hostname seeds to the discovered broadcast verbatim —
+       including the MAC-suffix factory firmware appends — and
+       latched, so friendly-name edits don't silently become a
+       rename. The import always lands under this broadcast name
+       (the only hostname the running device answers to); an edited
+       hostname is applied afterwards via the rename flow, which
+       flashes the device before the old name is dropped. */
+    void this.updateComplete.then(() => {
+      this._nameInputs?.reset(device.friendly_name || "", device.name);
+      this.requestUpdate();
+    });
     // A Wi-Fi device whose install has no shared wifi_ssid/wifi_password
     // would import a config referencing an undefined !secret and fail to
     // validate (esphome/device-builder#1742). Probe the shared secrets so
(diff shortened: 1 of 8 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
All five addressed in 0ea311f1:

1. **Authoritative configuration**: the import response's `configuration` now rides the `adopted` event and feeds `markJustCreated`, the highlight, and the rename's `configuration` argument — no more re-derived `${name}.yaml` strings.
2. **Taken-hostname gate**: the adopt dialog now takes `takenHostnames` (wired from `takenHostnameSet` like the clone/create dialogs), renders the `naming.hostname_taken` inline error, disables submit, and re-checks in `_submit` for the Enter path. The unedited factory default is exempt since its own importable row is in the set — pinned by two new tests.
3. **Factory-name validation**: took the comment route — the factory name is assumed valid because esphome enforces the same hostname charset on `name:` at compile time, so a genuine factory image can't broadcast a non-conforming name; recorded at the assignment.
4. **Orchestration coverage**: the `_onAdopted` body moved into `adoptFollowUp` in actions-ui (matching the file's delegation pattern), and two new tests pin the argument order (`configuration`, factory name, chosen name, `configOnly: false`), the follow-job call, and the `renameTo: null` no-op.
5. **Decision-time copy**: a `dashboard.adopt_rename_hint` line ("Changing the hostname will build and install new firmware over the network.") renders under the name field only while the name differs from the broadcast.

Also rebased onto the branch-update merge; the dialog tests carry the new `configuration` field in the event detail.

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
All four addressed in a22e2ac2:

1. **Stricter edited-name rules**: the name gate is now a single `_nameError` helper shared by render and `_submit` — `validateDeviceName` plus, for an edited name only, the 31-char `HOSTNAME_MAX_LEN` cap (`validation.max_length`), the hard edge-hyphen error (`naming.hostname_edge_hyphen`), and the taken check. Two new tests pin the cap and the hyphen refusal.
2. **One payload shape**: `AdoptedDetail` is exported next to `adoptFollowUp` and referenced by the dialog's `fireEvent` (typed const), `_onAdopted`'s annotation, and `adoptFollowUp`'s parameter.
3. **Config-only exclusion recorded**: the `adoptFollowUp` comment now states the offline config-only route is deliberately excluded because a config-only rewrite would strand a hostname the running factory firmware never broadcasts.
4. **Call-count assertion**: the second dialog test asserts `toHaveBeenCalledTimes(1)` before indexing.

PR body updated to cover the taken-hostname gate, the stricter name rules, and the `adopt_rename_hint` string.

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Both nits resolved structurally in 18fe5281 (maintainer-directed): the adopt dialog now uses the shared `esphome-device-name-inputs` pair the clone and create flows use, so the hostname rules live in exactly one place — `_nameError` and the raw fields are deleted, and the file drops to 571 lines. The only shared-component change is a `prefillHostname` seed on `reset()` that latches the hostname (adopt seeds the factory broadcast, so friendly-name edits never silently become a rename); the factory name's taken-set exemption moved to a memoized set-minus-factory passed as `takenHostnames`. The dialog tests now drive the shared pair through the DOM like the clone tests, including a new pin that a friendly-name-only edit keeps `renameTo: null`.

bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Hostname is not latched — a friendly-name edit still turns adopt into a compile + OTA rename

bdraco:
Addressed in 882ba355:

- **Blocking (latch-vs-derive mismatch)**: live derivation is the intended behavior — a maintainer correction after trying the latched build ("when I fill in the new Name it doesn't recalc the new hostname"), which is what 94f8f952 implemented. The stale "latched" wording was mine; the PR body now states the live-derivation contract, and the re-derive test remains the pin.
- **Suggestion 1 (cleared friendly strands an empty hostname)**: the `reset` seed now doubles as the derivation fallback — when the friendly name slugs to nothing, both derivation paths return to the factory broadcast instead of "", so submit never dead-ends. Empty for the create/clone flows, whose behavior is unchanged. Pinned by `test returns the hostname to the factory broadcast when the friendly name is cleared` (submit proceeds, `renameTo: null`).
- **Suggestion 2**: the rename hint carries `role="status"`.
- **Second-open stale frame**: `open()` now seeds the pair synchronously when it is already mounted (and again post-render for the first open), so a re-open never renders a frame of the previous device's state.
- **Nits**: `AdoptedDetail` stays beside `adoptFollowUp` — that placement was the round-2 suggestion this PR adopted, and the emitter imports the type without a cycle. The doubled `takenHostnameSet` calls resolve to one memoized set (WeakMap keyed on the `_devices` reference), so the second call is a cache hit.

bdraco:
@esphbot review

PR #1534, opened 2026-07-29
Read the legacy api services block spelling

bdraco:

Companion to esphome/device-builder#2420. The api actions manage list and the navigator use the frontend's synchronous YAML parser, which only looked for an `actions:` block; actions under the legacy `api: services:` spelling rendered as "No API actions defined yet." The block lookup now falls back to `services` (the item-level `service:` fallback already existed), and the caret-follow anchor accepts the legacy block path so cursor sync works before the backend's first canonicalizing write.

Replaces the closed #1527, which grew far past the problem.

- related to esphome/device-builder#2396

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/automation-editor/automation-focus.ts
+++ b/src/components/device/automation-editor/automation-focus.ts
@@ -53,12 +53,18 @@ export function automationRelativePath(
 ): YamlPathSegment[] | null {
   const anchor = location && handlerAnchor(location);
   if (!anchor) return null;
-  let at = path.indexOf(anchor.keys[0]);
-  while (at >= 0 && !anchor.keys.every((k, i) => path[at + i] === k)) {
-    at = path.indexOf(anchor.keys[0], at + 1);
+  let rest: YamlPathSegment[] | null = null;
+  for (const keys of anchor.keyPaths) {
+    let at = path.indexOf(keys[0]);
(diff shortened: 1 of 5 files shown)


bdraco:
@esphbot review

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Legacy `services:` items can take their name from a nested `action:` key

bdraco:
Addressed in d16f4aa2:

- The mislabel is fixed with the general form you suggested: `_readKeyOnLine`'s sibling scan is now pinned to the item's own content column (the dash line's content column, or the first body line's for a bare dash), so a same named key nested deeper in the body can never win, in either lookup order. Pinned by a test using your exact repro, a legacy `service: start_va` item whose `then:` calls `homeassistant.action:`.
- Added the anchor tests: the legacy `["api", "services", 1, "then", 0, "logger.log"]` slice, plus the bare `["api", "services"]` negative that pins the `rest === null` sentinel against a future `!rest` rewrite.
- Exported `API_ACTIONS_BLOCK_KEYS` from `yaml-automations.ts` and consumed it at all three sites, the block lookup, the focus anchor, and `reveal.ts`.

bdraco:
@esphbot review

PR #1540, opened 2026-07-29
Retire the section-level deprecation-nudge machinery

### LINKED ISSUE #2437: Retire the section-level deprecation-nudge machinery
Follow-up to #2426. With `editor/migrate_config` established as the single migration engine (one-click, whole-file, backend rules plus a frontend detection mirror), the section-level deprecation-nudge machinery in the frontend is legacy:

1. `src/components/device/deprecation-notice.ts` — the `DEPRECATED_OPTIONS` registry is now empty (the ethernet `clk_mode` entry moved into the backend engine via esphome/device-builder-frontend#1538), so the component never renders. The `renamedOption` helper, its shared `renamed_option_*` copy keys, the registry tests, and the mount in `device-section-config/render-branches.ts` are all dormant.
2. Future renames (the pending sgp4x/sen5x/sen6x `voc`/`nox` pairs) land as `controllers/migrations.py` rules plus `src/util/config-migrations.ts` detection, not as section-level entries — decided when consolidating the ethernet duplication.

Retire the mechanism when convenient: delete the component, helper, copy keys, tests, and mount; note in the migrate-command docs that section-scoped nudges are intentionally not a surface anymore. Not urgent — the dead registry is harmless — but the longer it sits the more likely someone adds an entry to the wrong side.

bdraco:

Retires the section-level deprecation-nudge machinery. Its registry has been empty since the ethernet migration moved into `editor/migrate_config`, and future renames land as backend rules behind the config-migration nudge, so the component could never render again. Deletes `deprecation-notice.ts` (the registry, `renamedOption`, the component), its mount in the section-config render branch, the orphaned `renamed_option_*` / `deprecation_applied` copy, and its tests. The host-side `applySectionValues` test survives reframed as `section-config-apply-values.test.ts` since it pins the generic draft-splice path, not the nudge.

- fixes https://github.com/esphome/device-builder/issues/2437

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1539, opened 2026-07-29
Carry the full field path in edit-action-field routing

bdraco:
# What does this implement/fix?

Second of three parts for esphome/device-builder#2392 (nested action lists cannot route to the automation editor). The TRIGGER field renderer emitted `edit-action-field` with only the leaf key, so an action list nested below a component's top level, sprinkler's `repeat_number.set_action` or a per valve `run_duration_number.set_action`, could not reach the automation editor.

The event now carries the field's full path (the renderer already threads it; segments are schema keys and stringified list indices, never containing dots), and the section host joins it into the dotted `component_action` field the backend addresses nested fields with (esphome/device-builder#2433). The section key format is unchanged, dots live inside the existing fourth segment and round-trip through `locationFromSectionKey`. Cursor driven focus follows the whole chain to the leaf, with decimal segments mapped to the numeric indices the cursor path carries. The YAML fallback parser's top level only `component_action` pass is deliberately unchanged; nested fields do not surface in the manage table at all (it reads that parser directly), and the in form edit button is their sole route. A nested field's header label prefixes the container chain, `Valves #1 → Run duration number → Set action`, so sibling `set_action` fields stay distinct.

No visible change until the catalog ships nested TRIGGER entries (the third part, the sync gate lift); direct child fields behave exactly as before.

**Related issue or feature (if applicable):**

- part of https://github.com/esphome/device-builder/issues/2392

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [ ] New feature (non-breaking change which adds functionality) — `new-feature`
- [x] Enhancement to an existing feature — `enhancement`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [ ] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `docs`
- [ ] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Checklist

- [x] The code change is tested and works locally.
- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.
- [x] Tests have been added to verify that the new code works (where applicable).

PR #1545, opened 2026-07-29
Emit nested component action fields as dotted sections

### LINKED ISSUE #1543: Nested action-list carets route to the component section, not the automation editor
From the esphbot review of #1539. sectionAtLine resolves a caret inside a nested action list (sprinkler's valves[i].run_duration_number.set_action) to the enclosing component section, because parseYamlAutomations's component_action pass is top level only. The widened component_action focus anchor in automation-focus.ts already handles the full dotted chain, so cursor driven focus goes live for nested fields once the fallback parser emits nested sections; until then the in form edit button is the sole route, which is the documented scope of #1539. Teaching the parser nested fields also surfaces the manage table rows for them.

bdraco:

The fallback YAML parser's `component_action` pass was gated to direct children of a component instance, so a nested action field (sprinkler's `repeat_number.set_action`, a per valve `run_duration_number.set_action`) produced no section: no navigator row, and a caret inside its body routed to the enclosing component section instead of the automation editor, even though #1539's routing and labels already speak the dotted encoding.

The depth gate is replaced by a path reconstruction: when the line scan hits a `*_action:` key inside an instance, a frame stack walk over the instance body rebuilds the concrete path, mapping keys by indent and list items as counted dash indices, and the section is keyed on the backend's dotted `component_action` encoding (`automation:component_action:lawn:valves.1.run_duration_number.set_action`). A field inside a trigger or automation body (an `on_*`, another `*_action`, or a `then` family ancestor) is excluded, which preserves the old exclusions by meaning rather than by depth; direct children keep byte identical keys, so pinned selections survive. `sectionAtLine`'s smallest span rule then routes carets to the new sections with no changes, the navigator labels them through the dotted aware `actionFieldLabel` ("Valves #2 → Run duration number → Set action"), and edit and delete round trip through the merged backend support (esphome/device-builder#2433).

Verified against the live dashboard with a two valve sprinkler: both nested fields list in the Automations navigator with container chain labels, clicking valve #2's row opens the automation editor with that valve's actions hydrated and the YAML pane highlighting exactly its `set_action:` block, and the flat cover fields and `on_press` rows are unchanged.

- fixes https://github.com/esphome/device-builder-frontend/issues/1543

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/util/action-field-path.ts
+++ b/src/util/action-field-path.ts
@@ -12,7 +12,7 @@ import type { YamlPathSegment } from "./yaml-ast.js";
  */
 
 /** Dotted `field` string for a config-entry path. */
-export function joinActionFieldPath(path: string[]): string {
+export function joinActionFieldPath(path: Array<string | number>): string {
   return path.join(".");
 }
 
--- a/src/util/yaml-ast.ts
+++ b/src/util/yaml-ast.ts
@@ -26,6 +26,7 @@
  * caller doesn't need a fallback for "the parse failed".
  */
 import { syntaxTree } from "@codemirror/language";
+import { isAutomationKey } from "./yaml-section-lexer.js";
 import type { EditorState } from "@codemirror/state";
 import type { SyntaxNode } from "@lezer/common";
 
@@ -126,15 +127,7 @@ export function isUnderThenItem(state: EditorState, pos: number): boolean {
  *  ``close_action``, lock ``unlock_action``, …). Covers every
  *  list-of-actions body in the ESPHome schema; broader than the
  *  legacy ``then``-only carve-out. */
-const RE_AUTOMATION_KEY = /^(?:then|else|on_[a-z0-9_]*|[a-z0-9_]+_action)$/;
-
-/** True for pair keys whose value is a list of actions: ``then`` /
- *  ``else`` / ``on_*`` / ``*_action``. The indent-based completion
- *  fallback uses this when the Lezer parse mis-nests an empty trailing
- *  ``- `` and ``isUnderAutomationItem`` can't see the enclosing key. */
-export function isAutomationKey(key: string): boolean {
-  return RE_AUTOMATION_KEY.test(key);
-}
+export { isAutomationKey } from "./yaml-section-lexer.js";
 
 /**
  * True when *pos* is inside a list ``Item`` whose enclosing Pair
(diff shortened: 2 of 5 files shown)


bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Zero-indent nested sequences lose their parent key from the dotted path

bdraco:
@esphbot review

bdraco:
All four findings taken in a5daf195. The zero indent sequence fix is the one token you validated (a dash keeps any same indent frame, so the parent key whose column the dashes share survives), pinned by a two valve zero indent regression asserting both full dotted paths. The instance dash line's bare inline first key is now seeded as a frame before the walk, with its own pin. The exclusion test now reaches isAutomationKey for real: the cover fixture nests a stop_action key inside open_action's body via then, so a mutation of the check fails the assert. And endsBlockAtIndent has its JSDoc back, with the hoisted predicate moved above the pair.

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

bdraco:
Guard taken in d989ddb1: an index first path (a container key the bare key regex cannot name, quoted or hyphenated) now refuses instead of emitting an unroutable row, restoring main's no row fallback, pinned with a quoted container fixture. The unanchored _COMPONENT_ACTION_FIELD_RE note is acknowledged as pre existing and main identical; happy to tighten it in a follow up if it ever bites.

esphbot (approved):
> [!TIP]

bdraco:
Both remaining notes taken in the new head: the index guard generalises to any index without a preceding key segment, so a mid path unnameable container with a list under it refuses instead of emitting a wrong but routable path (a quoted container over a mapping stays the documented best effort limit), and the walk now consumes block scalars via BLOCK_SCALAR_RE plus _blockScalarBodyEnd, refusing a target inside one and skipping past it otherwise, so *_action shaped text in a lambda body no longer emits a phantom row while a real field after the scalar still resolves. Both pinned.

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
All three taken in the new head: the instance dash line block scalar is consumed before the walk starts (refusing a target inside it), a container key containing a dot refuses under the widened encoding guard so the lossless join invariant holds, and the file size crossing is covered by #1548, whose core move of the walk takes _actionFieldPath out of this module entirely. Both new behaviours pinned.

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
The quadratic walk note is deferred to #1548 with your measurements recorded there; its single forward pass per host is exactly the fix and also takes the walk out of this file for the size cap. The anchored container key joins the documented best effort class alongside the quoted and hyphenated spellings. No code change, so the head stays 0aad7626.

PR #1547, opened 2026-07-29
Expand maybe_simple_value shorthands at section load

### LINKED ISSUE #2397: [Bug] “microphone” keyword accepts either a single value or a list, but the structured editor accepts only a list
### What happened?

The “microphone” keyword accepts either a single value or a list. However, the structured editor accepts only a list.

[Docs](https://esphome.io/components/voice_assistant/)

[YAML](https://github.com/esphome/media-players/blob/main/onju-voice/onju-voice.yaml)

Image

### Dashboard version

1.7.1b1

### How did you install it?

Desktop app

### Where is it running?

Local — Windows

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

Frontend half of esphome/device-builder#2397. esphome accepts `voice_assistant.microphone` as either a bare id or a list of source mappings (`cv.ensure_list(cv.maybe_simple_value(..., key="microphone"))`), but the structured editor rendered the field list-only: for `microphone: onju_microphone` it showed "No items yet — click + to add one", and clicking Add would have replaced the scalar with `[{}]`.

The backend companion PR (esphome/device-builder#2443) carries the schema's shorthand marker through the catalog as `ConfigEntry.maybe_key` (the child key a bare scalar expands into; with `multi_value`, list items may also be bare scalars). This PR consumes it: a new `normalizeMaybeValues` pass runs once at section load, next to the existing hex normalization, expanding shorthands into the canonical mapping/list shape the renderers, dotted-path reads, and edits all address. Expanding at load (rather than per-renderer) is load-bearing: child fields read and write through `ctx.getAt`/`emitChange` against the stored values, so a display-only coercion would leave the child inputs blank and the first edit would clobber the scalar.

The walk descends nested entries so shorthands on descendants (sprinkler's `valves[].valve_switch: "Name"`) expand too, keeps identity when nothing changes, preserves the parser's null-prototype defence when cloning, and passes parser wrappers (`YamlRawValue`) through untouched so the raw-block bail-outs keep seeing them. The YAML keeps its shorthand until the section is next written; a section edit canonicalizes to the equivalent block-list form (same canonicalize-on-edit policy as legacy key respelling).

Verified live against a dev backend running the companion branch: `microphone: onju_microphone` renders as a populated "Microphone 1" group; the untouched YAML keeps the scalar; editing a sibling field and saving writes `microphone:` / `- microphone: onju_microphone`, which `esphome config` validates.

Scope note: the automation editor feeds catalog entries into the same form without this normalization, so maybe_simple_value shorthands in action/trigger params keep the old behavior there; that surface is a follow-up.

- fixes https://github.com/esphome/device-builder/issues/2397
- companion backend PR: esphome/device-builder#2443

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/api/types/config-entries.ts
+++ b/src/api/types/config-entries.ts
@@ -132,6 +132,14 @@ export interface ConfigEntry {
   unit_options: string[] | null;
   /** When True the field accepts a list of values. */
   multi_value: boolean;
+  /**
+   * For NESTED entries validated by esphome's `maybe_simple_value`: the
+   * child key a bare scalar value expands into
+   * (`microphone: mic_id` == `microphone: {microphone: mic_id}`).
+   * With `multi_value` each list item may also be a bare scalar. Null
+   * (or absent) = no scalar shorthand.
+   */
+  maybe_key?: string | null;
   /** When True accepts either a literal value OR a !lambda block. */
   templatable: boolean;
   /**
--- a/src/components/device/device-section-config/loading.ts
+++ b/src/components/device/device-section-config/loading.ts
@@ -1,6 +1,7 @@
 import type { ConfigEntry, RequiredGroup } from "../../../api/types/config-entries.js";
 import { fetchComponent } from "../../../util/component-name-cache.js";
 import { normalizeHexValues } from "../../../util/hex-int.js";
+import { normalizeMaybeValues } from "../../../util/maybe-values.js";
 import { loadCatalog } from "../../../util/yaml-completion-catalog.js";
 import { platformDomains } from "../../../util/yaml-completion-items.js";
 import { parseYamlSectionValues } from "../../../util/yaml-section-reader.js";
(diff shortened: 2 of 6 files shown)


bdraco:
@esphbot review

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Import order fails `prettier --check`, which CI enforces

bdraco:
Addressed the round in 4efc4245.

On the blocking formatting warning: it does not reproduce. `pnpm run format:check` passes locally on the flagged file, and the PR head's own Prettier + ESLint CI job (which runs format:check) is green — the repo's prettier keeps the import where it is, so nothing gates merge here.

Taken:
- `makeConfigEntry` now lists `maybe_key: null`, so the factory shows the full `ConfigEntry` shape again.
- The wrapper-passthrough tests construct real `YamlRawValue` instances, pinning the renderer bail-out's actual `instanceof` contract.
- The module docstring now distinguishes the two prior behaviors (empty-list clobber under `multi_value`, read-only value-set-in-YAML notice for single groups), and I ran the suggested live check on the single-nested path: a sprinkler config with `valve_switch: Front Lawn` and `main_switch: Main Lawn` renders both groups populated with no phantom required errors (the catalog marks no required children under either, so none can surface).
- The automation-forms gap is now called out in the PR body as an explicit follow-up.

Declined: relaxing the `maybeKey &&` guard to wrap bare mappings on every `multi_value` entry. Same bug class, but this PR is deliberately driven by the catalog's marker; blanket wrapping changes entries the backend hasn't vouched for and belongs in its own change.

@esphbot review

esphbot (approved):
> [!TIP]

PR #1552, opened 2026-07-29
Prefill the entity name from the catalog title in the add-component dialog

### LINKED ISSUE #2450: A name should always be pre-filled and suggested so users don't have to figure it out
Image

Image

bdraco:

Adding an entity platform through the add-component dialog left the Name field empty; the user had to invent one, and an untouched Add produced an id-only entity that ESPHome treats as internal (never exposed to Home Assistant).

The form now seeds the Name field from the component's catalog title ("GPIO Switch", "A02YYUW Waterproof Ultrasonic Sensor"), editable and clearable like the auto-generated id. The suggestion is deduped against the `name:` values already in the component's own top-level section: ESPHome hard-fails on duplicate entity names, comparing `sanitize(snake_case(name))` per device/platform (`core/entity_helpers.py`), so the new `suggestEntityName` helper mirrors that exact key and walks a numeric suffix ("GPIO Switch 2") on collision, while a same-named entity on another platform doesn't force a suffix. The line scan still over-collects within the section (nested sub-entity names, sub-device duplicates ESPHome would allow), costing at most an unneeded suffix; a name behind a block scalar or in a `packages:`/`!include` file is invisible to it and can still collide, and the user sees the seeded value before pressing Add.

Scope of the seed: platform-shaped ids only (id contains `.`), mirroring `generateDefaultComponentId`'s judgment. A full-catalog survey shows top-level `name` string entries split into 290 dotted entity platforms (all entity domains) and 6 undotted components (`esphome`, `esp32_ble`, `ble_client`, `esp32_camera`, `serial_proxy`, `sprinkler`) where `name` means something else (device hostname, BLE advertised name, ...), so undotted ids never seed. Featured wraps are also excluded: their name presets already arrive through `seedDefaults`, and their display title can be the "SPI Bus (lcd_spi)" fallback shape.

Behavior notes:

- The seed sits beside the id seed in `buildInitialValues`, so detour-restored values, `prefillReference`, and `prefillFields` all still override it, and a `default_value`/preset name wins via the `undefined` guard.
- The dialog's skip-the-form fast path cannot flip: `name` is in `ALWAYS_SHOWN_KEYS`, so any component this seeds already painted an unlocked field.
- Clearing the field before Add still writes id-only YAML (`coerceFields` drops an empty string on a non-required field), preserving the internal-entity opt-out.
- No backend/catalog change needed; the title comes from `ComponentCatalogEntry.name`, which the dialog already holds.

- fixes https://github.com/esphome/device-builder/issues/2450

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/add-component-form-seed.ts
+++ b/src/components/device/add-component-form-seed.ts
@@ -12,10 +12,7 @@ import {
   collectExistingIds,
   generateDefaultComponentId,
 } from "../../util/default-component-id.js";
-import {
-  collectExistingNames,
-  suggestEntityName,
-} from "../../util/default-entity-name.js";
+import { suggestEntityName } from "../../util/default-entity-name.js";
 import { resolveEntryLabel } from "../../util/entry-label.js";
 import { isFeaturedId } from "../../util/featured-id.js";
 import { getIn, setIn } from "../../util/nested-values.js";
@@ -205,13 +202,7 @@ export function buildInitialValues(ctx: SeedContext): Record<string, unknown> {
     (e) => e.key === "name" && e.type === ConfigEntryType.STRING
   );
   if (nameEntry && next["name"] === undefined) {
-    // Names collide per platform (unlike globally-unique ids), so the
-    // pool is scoped to the component's own top-level section.
-    const seededName = suggestEntityName(
-      component.id,
-      component.name,
-      collectExistingNames(yaml, component.id.split(".", 1)[0])
-    );
+    const seededName = suggestEntityName(component.id, component.name, yaml);
     if (seededName !== null) next = { ...next, name: seededName };
   }
 
--- a/src/components/device/add-component-form-seed.ts
+++ b/src/components/device/add-component-form-seed.ts
@@ -205,10 +205,12 @@ export function buildInitialValues(ctx: SeedContext): Record<string, unknown> {
     (e) => e.key === "name" && e.type === ConfigEntryType.STRING
   );
   if (nameEntry && next["name"] === undefined) {
+    // Names collide per platform (unlike globally-unique ids), so the
+    // pool is scoped to the component's own top-level section.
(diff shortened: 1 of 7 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
All four addressed in b33e1430, plus a scope correction from review here: ESPHome's duplicate check is per device/platform, so `collectExistingNames` now collects only under the component's own top-level section instead of the whole file (a switch named like a sensor no longer forces a suffix).

1. Reworded the collector's doc comment (and the PR body) to state the line scan's real limits: block scalars and `packages:`/`!include` names are invisible and can still collide.
2. Extracted `collectInstanceScalars(yaml, key, section?)` next to `readInstanceScalar`; `collectExistingIds` / `collectExistingNames` are now thin wrappers.
3. Extracted `isPlatformComponentId` into `featured-id.ts`; both seeds share the one judgment.
4. The fixture now covers a real `- name:` list-item line.

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
All three addressed in edfafd89:

1. `suggestEntityName(componentId, title, yaml)` now owns both halves: it gates on `isPlatformComponentId`, derives the section via `parseCatalogId(componentId).domain`, and runs the scan itself, so an ineligible id never walks the YAML and the scope judgment can't drift from the eligibility judgment.
2. `collectInstanceScalars` reuses `TOP_LEVEL_KEY_RE`.
3. The scoping fixture now has a column-0 banner comment inside the sensor section with a name collected after it.

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1560, opened 2026-07-30
Share the widened id pool with top-level seeding

bdraco:

Follow-up to #1553, which widened the uniqueness pool for nested-list row ids from `id:` lines to every id-naming key (`id`, `*_id`) after a review turned up `tca9548a.channels[].bus_id` — a declaring id that mints from the same base slug as `usb_uart.channels[].id`. That widening only landed on the nested-row path; the top-level component seeding still scanned `id:` alone, so a hand-typed `bus_id: uart_1` could still collide with a freshly minted `uart_1` from the add dialog. Both minting paths now share `collectTakenIds`.

The review suggested repointing all three `collectExistingIds` call sites, but the other two are presence checks rather than minting: `_missingRequiredPrereqs` asks whether a featured prerequisite is already in the config, and the bundle-add loop skips members that are already present. For a minted id, over-inclusion is harmless since it only bumps the numeric suffix; for a presence check it would change behaviour, letting a mere reference (`uart_id: my_prereq`) read as the prerequisite being installed and suppress the warning or silently skip a bundle member. Those two keep the narrow scan.

- fixes <link to issue>

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/add-component-form-seed.ts
+++ b/src/components/device/add-component-form-seed.ts
@@ -10,7 +10,7 @@ import {
   resolveSoleCandidate,
 } from "../../util/config-entry-yaml-scan.js";
 import {
-  collectExistingIds,
+  collectTakenIds,
   generateDefaultComponentId,
 } from "../../util/default-component-id.js";
 import { suggestEntityName } from "../../util/default-entity-name.js";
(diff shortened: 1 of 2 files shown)


bdraco:
Follows up the non-blocking suggestion on #1553. Note it repoints only the *minting* call site, not all three — reasoning in the PR body.

@esphbot review

PR #1567, opened 2026-07-30
Flip the canonical rp2040 platform key to rp2

bdraco:

Frontend half of https://github.com/esphome/backlog/issues/155 (companion to esphome/device-builder#2463): the canonical rp2040 platform key flips to `rp2`. The swap of `RP2_CANONICAL_KEY` / `RP2_ALIAS_KEY` in `component-presence.ts` does the work; `canonicalComponentKey` now folds `rp2040` onto `rp2`, and the alias map, target platform set, wizard platform rows, navigator icons, board reselect matching, and the add component deps `provides` lookup all follow. Hand edits: a `wizard.tag.rp2` label in en.json for the chip badge's platform fallback (the `rp2040` key stays because it is the chip series tag beside `rp2350`, unrelated to the platform key), and the crash report keeps both spellings, mapping either to the bug form's RP2040 dropdown value, since 2026.7+ firmware reports the renamed platform while older builds report the legacy one.

Kept as is, deliberately: the wizard chips' `mcu: "rp2040"` filter values (chip series, not the platform key), the `rp2040_ble` icon key (a real component id), and the `esphome-web-rp2040.uf2` asset name. The three tests that pinned the old canonical output inverted to exercise the legacy fold instead.

Lockstep pair with the backend regen (catalog id and `provides` index re-key to `rp2`); neither side ships alone.

- implements https://github.com/esphome/backlog/issues/155 (cross-repo, close manually on merge)

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Took the round in 3d747a71. Verified against esphome 2026.7.3 that the crash handler now tags rp2.crash, so log-doc-links carries both spellings with the same firmware skew rationale as the crash report. The raw-key catalog lookups need no fold: every path goes through fetchComponent, which is components/get_component_bodies, and the backend normalizes the id server side (its own flip test pins a raw rp2040 request hydrating the rp2 body), so an unmigrated rp2040: block keeps its form. The stale platform-key prose is swept, the browser-flash gate tests gained the rp2 rows and the corrected comment, and the PR body's reason for keeping wizard.tag.rp2040 is fixed, it is the chip series tag; the new rp2 key covers the chip badge's platform fallback for a board without mcu or variant. @esphbot review

esphbot (approved):
> [!TIP]

bdraco:
All four taken in the new push: the rp2.crash it.each row, assertions behind both crash-report table entries (plus a first test for platformFromIntegrations), the remaining platform-key prose in the two test files and pin/gpio.ts and boards.ts, and the three fixtures now stand in the platform the catalog actually produces. @esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Import hoisted to sorted position and both greens taken in the new push, the filters fixture now exercises the legacy rp2040 block against a rp2 catalog entry. On the 🟡: format:check does not actually go red here, and the reason is worse than the prediction, prettier-plugin-organize-imports silently no-ops under this repo's typescript 7 (no JS language service for the plugin to drive), so import order is unenforced everywhere and the check is green only because the sorter is dead. Verified with an unsorted probe file passing --check untouched. Filed #1572 to enforce import order through a live mechanism. @esphbot review

PR #1575, opened 2026-07-30
Enforce import order with eslint-plugin-perfectionist

### LINKED ISSUE #1572: prettier-plugin-organize-imports silently no-ops under typescript 7
The Koan review on #1567 flagged a misplaced import in crash-report.ts, predicting the format:check CI step would go red because prettier-plugin-organize-imports hoists imports. It does not go red, and the reason is worse than a missing CI step: the plugin is configured in .prettierrc but silently does nothing under our toolchain.

Repro on main: write a file with obviously unsorted imports, then

    npx prettier --check src/util/__probe.ts   # passes
    npx prettier --write src/util/__probe.ts   # leaves the order untouched

The plugin resolves against typescript 7.0.2 (the native preview), which lacks the JS LanguageService API the plugin drives, so it no-ops without an error. Import order is therefore unenforced everywhere, format:check in CI and the husky hook are green only because the sorter is dead, and reviewers reasonably assume it runs.

Options, roughly in order of preference: enforce import order through eslint instead (the eslint chain already gates CI and does not depend on the ts language service), pin a plugin-compatible typescript for prettier only, or drop the plugin declaration so the config stops promising a sort that never happens.

bdraco:

prettier-plugin-organize-imports has silently no-opped since the typescript 7.0.2 bump (#1160): the TS 7 native preview lacks the JS LanguageService API the plugin drives, so import order has been unenforced while format:check stayed green. This takes the issue's preferred option and moves enforcement to eslint, which already gates CI and pre-commit.

Two perfectionist rules in eslint.config.mjs encode the convention organize-imports had established: `sort-imports` (packages before relative paths, type and value imports interleaved by specifier) and `sort-named-imports`. Two structures the old plugin preserved stay untouched: side-effect imports (element registration, polyfills, test mocks) hold their exact statement positions, and blank-line-separated import paragraphs sort only internally via `partitionByNewLine`, which keeps vitest-environment pragmas at the top of their files and deliberate orderings like styles/theme.ts intact.

perfectionist itself probes `require("typescript")` for `isExternalModuleNameRelative` and crashes under TS 7, so `.pnpmfile.cjs` gives it the same private TS ~6.0.3 pin typescript-eslint already uses. The dead plugin is removed from .prettierrc and package.json. The one job organize-imports did that perfectionist does not (pruning unused imports) is already covered twice: `noUnusedLocals` in the tsc gate and `@typescript-eslint/no-unused-vars`.

The bulk of the diff is one mechanical `eslint --fix` sweep over src/ and test/ (pure reorders, verified: every side-effect import keeps its position, zero net line change). Where the sorter would have displaced a file-header comment or vitest-environment pragma by dragging it along with its adjacent import, the header is re-anchored at the top (blank-line separated, which `partitionByNewLine` then respects), so pragmas and file docblocks stay where they were.

Verified: the unsorted-probe repro from the issue now fails `pnpm run lint:eslint` and is repaired by `--fix`; full suite (6823 tests), tsc, prettier, and both builds pass.

- fixes https://github.com/esphome/device-builder-frontend/issues/1572

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1586, opened 2026-08-01
Keep a stack of dependency detours in the add-component dialog

bdraco:

A dependency can itself be missing a dependency; `climate.anova` needs `ble_client`, which needs `esp32_ble_tracker`. The detour return address was a single slot, so the second "+ Add <dep>" click overwrote the first. Adding the tracker returned to the BLE Client form with no return banner, and adding the client closed the dialog outright; the user never got back to the Anova Cooker form and the values they had typed there were gone.

`_returnTo` / `_depDomain` become a `_detourStack` of frames, each carrying the suspended component, the dep domain it was sent off to satisfy, its in-progress values and its prefill. `navigateToDep` pushes, the new `popDetour` restores one level, and both the submit return and Back unwind a single frame at a time. `_returnValues` turns into a strict hand off slot (push parks it on the frame, pop hands it back), which is why the `_restoredValuesForMount` gate is gone; the naive "any detour in flight" version withheld the inner level's own values once the stack was deeper than one.

No cycle guard or depth cap: each submit merges the component into the YAML and the form recomputes its missing dep banner, so a cycle stops offering the button, and every frame costs a deliberate click.

Verified in the browser as well as in tests: anova, then add BLE Client, then add ESP32 BLE Tracker Hub, and the chain returns through BLE Client to the Anova Cooker form with its name still filled in.

One pre existing quirk this preserves rather than fixes: Back during a detour drops an in flight bundle queue while the submit return keeps it.

- N/A, reported directly

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/add-component-dialog-dep-nav.ts
+++ b/src/components/device/add-component-dialog-dep-nav.ts
@@ -7,16 +7,28 @@ import {
 } from "../../util/bus-constraint-prefill.js";
 import { fetchComponent } from "../../util/component-name-cache.js";
 
+/** One suspended level of a "+ Add <dep>" detour. */
+export interface DetourFrame {
+  component: ComponentCatalogEntry;
+  depDomain: string;
+  values: Record<string, unknown> | null;
+  prefill: BusPrefill | null;
+}
+
+/** Slice of ``ESPHomeAddComponentDialog`` state a suspended level restores into. */
+export interface DetourHost {
+  _selected: ComponentCatalogEntry | null;
+  _detourStack: DetourFrame[];
+  _returnValues: Record<string, unknown> | null;
+  _depPrefill: BusPrefill | null;
+}
+
 /** Slice of ``ESPHomeAddComponentDialog`` state ``navigateToDep`` reads / writes. */
-export interface DepNavHost {
+export interface DepNavHost extends DetourHost {
   readonly _api: ESPHomeAPI;
   platform: string;
   board: { id: string; featured_components?: FeaturedComponent[] } | null;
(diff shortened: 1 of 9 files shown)


bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

bdraco:
@esphbot @Copilot thanks, pushed c78a25c4 for the first two.

1. Prefill on pop: agreed the nested return could revert an edit, so a restored level now keeps the frame prefill's `required` / `optionOverrides` but drops its `fields`; the snapshot already carries whatever the prefill seeded, edits included. A level that had no snapshot still gets its prefill whole.

2. Catalog-fallback bleed: fixed at the source rather than in the fallback branch; `_depPrefill` is nulled right after the frame captures it, so both branches start clean and the direct branch sets its own. Test added for the fallback path.

3. Prerequisite / bundle pick mid-detour: left as is, out of scope for this fix; I extended the comment to say the clear now drops every suspended level rather than one.

On the TypeScript note about `entries.filter((e) => e !== null)`: that narrows on its own in TS 5.5+ via inferred type predicates, and this repo builds with 7.0.2, so `pnpm run lint` (`tsc --noEmit`) passes as written.

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1587, opened 2026-08-01
Prefill the name for non-platform components that carry a label

bdraco:

Adding a BLE Client left the Name field empty. The catalog-title seed from #1552 only ran for platform-shaped ids (`switch.gpio`), and bailed on every undotted component because on some of those `name` means something other than a label.

I went through the catalog: exactly six undotted components have a top-level string `name`. Two are identities the firmware derives elsewhere, `esphome` (the node hostname) and `esp32_ble` (the advertised BLE name, which defaults to that hostname); those still return null. The other four are labels and now get seeded: `ble_client`, `esp32_camera` (whose schema extends `ENTITY_BASE_SCHEMA`, so it is a real Home Assistant entity name), `serial_proxy` and `sprinkler`. The collision suffix works the same way, scoped to the component's own top-level section.

- follow-up to #1552

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/util/default-entity-name.ts
+++ b/src/util/default-entity-name.ts
@@ -1,5 +1,5 @@
 import { parseCatalogId } from "./config-entry-yaml-scan.js";
-import { isFeaturedId, isPlatformComponentId } from "./featured-id.js";
+import { isFeaturedId } from "./featured-id.js";
 import { collectInstanceScalars } from "./yaml-instance-scalars.js";
 
 /**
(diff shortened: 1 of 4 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot thanks, all three applied in 4204fd67.

1. Dropped the `!isPlatformComponentId` conjunct; the identity set alone decides now, so a dotted id added to it later would still take effect.

2. The doc comment records that the enumeration is against the 2026.7.3 catalog and says a later upstream identity field needs adding to the set. Agreed a positive signal from the backend would be better than a denylist; that is a coordinated catalog change, not this PR.

3. Reworded the `collectExistingNames` docstring: per-platform scope for a dotted id, the component's own section for an undotted one, whose entities may well be named under a different domain.

Also worth recording here: I drove the whole thing in the browser against a live backend, and the BLE Client form now mounts with Name prefilled.

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Good catch on the stale docstring, that drift was mine; 8699a4e4 drops the "shared by the id and name seeds" clause. `isPlatformComponentId` still has two live callers (default-component-id, draft-and-delete), so only the claim went.

@esphbot review

esphbot (approved):
> [!TIP]

PR #1598, opened 2026-08-01
Fix unreadable warning text in dark mode

bdraco:

Warning text was unreadable in dark mode: dark brown text on the near-black dark-theme warning fill. The cause is a phantom token; `--wa-color-warning-text-quiet` does not exist in the WebAwesome theme (the text-on-fill tokens are named `--wa-color-warning-on-quiet` / `-on-normal` / `-on-loud`), so the color always fell through to the hard-coded light-mode fallback while the background token resolved per theme.

Replaces the phantom token with `--wa-color-warning-on-quiet` at its three call sites: the shared `warningBannerStyles`, the update-all dialog's summary note, and the constraint-cluster header. In light mode the shared banner keeps the tone its fallback approximated; the other two sites had no working fallback, so they gain a proper warning tone in light mode too (verified visually, both look right). Dark mode gets the theme's readable warning tone.

Also adds a guard test asserting every `--wa-color-*` token referenced under src/ exists in the WebAwesome styles tree, so the next phantom fails CI instead of shipping silently. The guard surfaced the full extent of this class: ten more phantom tokens (`danger-text-normal`, `brand-text-quiet`, `border`, `danger-border`, `neutral-500`, `success-quiet`, `surface-subtle`, `text`, `text-subtle`, `warning-quiet`) are allowlisted in the test with a note pointing at the follow-up sweep; those degrade to inherited colors rather than becoming unreadable, so they stay out of this fix.

- noticed while reviewing esphome/device-builder-frontend#1597

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/config-entry-renderers/constraint-cluster.styles.ts
+++ b/src/components/device/config-entry-renderers/constraint-cluster.styles.ts
@@ -29,7 +29,7 @@ export const constraintClusterStyles = css`
     color: var(--wa-color-text-quiet);
   }
   .constraint-cluster-header.unsatisfied {
-    color: var(--wa-color-warning-text-quiet, currentColor);
+    color: var(--wa-color-warning-on-quiet, currentColor);
   }
   .constraint-cluster-header wa-icon {
     flex-shrink: 0;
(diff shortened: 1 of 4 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

PR #1602, opened 2026-08-01
Fix section splice dropping keys from CRLF documents

### LINKED ISSUE #1601: Section splice drops keys from CRLF documents
The YAML section machinery silently drops keys from documents with CRLF line endings. `parseSectionCore` splits on `\n`, leaving a trailing `\r` on every line; child keys then fail the key match, so the parsed section comes back empty, and `updateSectionInYaml` replaces the block with only the values the caller passed. Every other key in the section is gone from the written file.

Repro through the public splice path:

```ts
updateSectionInYaml("wifi:\r\n  ssid: net\r\n", "wifi", { use_address: "10.0.0.9" })
// => "wifi:\r\n  use_address: 10.0.0.9\n"   ssid is gone
```

Found while adding a CRLF round trip test to the troubleshooting dialog's use_address writer in #1597; that path now refuses CRLF documents outright (`findNetworkSection` returns null, falling back to the copyable snippet) so it cannot lose data. The GUI section editor rides the same `parseSectionCore` / `updateSectionInYaml` machinery, so a config saved from a Windows editor with CRLF endings is likely exposed to the same key loss on any section-form save. `test/util/yaml-section-*.test.ts` has no CRLF coverage.

Possible directions: tolerate `\r` in the lexer/parser line handling (and preserve it on untouched lines, emitting new lines with the document's ending), or normalize the document to LF at the editor boundary before any section machinery touches it. Either way the fix belongs in the shared machinery, with round trip tests, after which the CRLF refusal in `use-address-yaml.ts` can be lifted.

bdraco:

The section machinery split documents on `\n` only, so every line of a CRLF config kept a trailing `\r`; the key regexes then failed to match, the parse came back empty, and a section form save wrote back only the caller's values, dropping everything else in the block. The read side had the same problem, a CRLF config opened the form empty and the navigator lost `name:` labels.

Documents are now split through shared helpers in a new dependency free `yaml-doc-lines.ts` module: `splitYamlDocLines` strips the CR per line (including a bare CR on an unterminated final line), and `yamlDocEol` picks the document's ending by majority vote with ties going to LF, so one pasted CRLF line in an LF file normalizes to LF instead of converting the whole file. The section entry points rejoin with that ending, keeping a Windows edited file CRLF throughout with untouched lines round tripping byte for byte. The same helper now feeds the other line scans in the machinery (`parseYamlTopLevelSections`, `findAddedSection`, `assessBusHostability`), the sensitive value scanner, and the superseded delete rebase check, which compared lines byte for byte and would have declined on normalized mixed ending buffers.

Round trip tests cover parse, update (map, list item inline key, globals, block scalar body), remove, append, both mixed ending directions, the bare final CR, and the LF path.

Once this merges, the CRLF refusal added to the use_address writer in #1597 can be lifted on that branch.

- fixes #1601

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/apply-removal.ts
+++ b/src/components/device/apply-removal.ts
@@ -6,6 +6,7 @@
  */
 import type { ESPHomeAPI } from "../../api/esphome-api.js";
 import type { AutomationLocation } from "../../api/types/automations.js";
+import { splitYamlDocLines } from "../../util/yaml-section-lexer.js";
 import { removeSectionFromYaml } from "../../util/yaml-section-values.js";
 import { resolveCurrentSectionLine } from "../../util/yaml-sections.js";
 import { applyYamlDiff } from "./automation-editor/serialise.js";
@@ -78,10 +79,12 @@ export async function applyRemoval(
 }
 
 /** Lines removed going from *before* to *after*, or ``null`` when
- *  the edit is not a pure contiguous removal. */
+ *  the edit is not a pure contiguous removal. Compared CR-stripped:
+ *  the section machinery normalizes a mixed-ending document's line
+ *  endings, and an ending-only difference is not a removal. */
 function removedLines(before: string, after: string): string[] | null {
-  const b = before.split("\n");
-  const a = after.split("\n");
+  const b = splitYamlDocLines(before);
+  const a = splitYamlDocLines(after);
   let start = 0;
   while (start < a.length && b[start] === a[start]) start++;
   let endB = b.length - 1;
--- a/src/components/device/apply-removal.ts
+++ b/src/components/device/apply-removal.ts
@@ -6,7 +6,7 @@
  */
 import type { ESPHomeAPI } from "../../api/esphome-api.js";
 import type { AutomationLocation } from "../../api/types/automations.js";
-import { splitYamlDocLines } from "../../util/yaml-section-lexer.js";
+import { splitYamlDocLines } from "../../util/yaml-doc-lines.js";
 import { removeSectionFromYaml } from "../../util/yaml-section-values.js";
 import { resolveCurrentSectionLine } from "../../util/yaml-sections.js";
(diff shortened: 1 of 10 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

bdraco:
@esphbot review

bdraco:
Addressed in the last two commits: appendSectionToYaml now normalizes the existing body through the same split and join, yamlDocEol is a majority vote with ties going to LF so a stray pasted CRLF line normalizes instead of converting the whole file, and the suite now covers the list item inline key path, globals, a CRLF block scalar body, and both mixed ending directions. Also CR-stripped the remaining raw splits inside the section machinery (parseYamlTopLevelSections, findAddedSection, assessBusHostability); the navigator was losing name: on CRLF docs through the same regex class. Skipped the stripTrailingCr consolidation, sensitive-scan's LineSource branch strips per line so the private helper stays either way and the cross module import buys nothing.

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Round three notes addressed: splitYamlDocLines now strips a bare CR on an unterminated final line so the sensitive scanner cannot fail open on a truncated CRLF write, both helpers moved to a new dependency free yaml-doc-lines.ts so the scanner keeps its zero import property and the lexer stays about lexing, and the deliberate whole document normalization is now noted at the eol site for the emitYamlDraft case. PR summary updated to match the final scope.

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1606, opened 2026-08-01
Fix CRLF-blind entity name pool and address CRLF review nits

### LINKED ISSUE #1605: Default entity name uniqueness pool is empty on CRLF configs
collectInstanceScalars feeds readInstanceScalar raw split("\n") lines, so on a CRLF config every line keeps its trailing CR. The name regex ends (.*)$ and JS regexes treat CR as a line terminator, so name lines never match; id and platform only survive through their trailing \s*$. defaultEntityName then sees an empty pool and can suggest names that collide with existing ones, and the id pool in default-component-id works only by accident of the regex shape.

Surfaced by the post merge review on #1602, which fixed the same failure class in the section machinery. The fix is to split through splitYamlDocLines from yaml-doc-lines.ts (added in #1602) and pin it with a CRLF test.

bdraco:

Follow up to the post merge review on #1602. The default entity name uniqueness pool was empty on CRLF configs: collectInstanceScalars fed readInstanceScalar raw split lines, the name regex ends `(.*)$` which cannot match a CR bearing line, so suggested names could collide with existing ones. It now splits through the shared `splitYamlDocLines`.

Also from the review notes: `yamlDocEol` is now an allocation free index scan instead of two whole document splits, since it runs on every debounced section form keystroke; and the documented tie to LF rule plus the degenerate inputs get direct tests in a new `yaml-doc-lines.test.ts`.

- fixes #1605

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/util/yaml-doc-lines.ts
+++ b/src/util/yaml-doc-lines.ts
@@ -26,10 +26,15 @@ export function splitYamlDocLines(yaml: string): string[] {
  * The document's line ending, by majority vote (ties go to LF). A
  * mixed-ending document normalizes toward its dominant ending, so one
  * pasted CRLF line in an LF file costs that line its CR on the next
- * save instead of converting the whole file.
+ * save instead of converting the whole file. An index scan, not a
+ * split: this runs on every debounced section-form keystroke.
  */
 export function yamlDocEol(yaml: string): "\r\n" | "\n" {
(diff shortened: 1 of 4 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1607, opened 2026-08-02
CR-strip the remaining readInstanceScalar line feeds

bdraco:

Sweeps the last CRLF blind readInstanceScalar feeders flagged in the #1606 review. collectTakenIds dropped every `*_id` key on CRLF configs (the same collision hazard as #1605, one file over), readPlatformBoard reported board and variant as null so board mismatch detection went quiet, and the pin scans in config-entry-yaml-scan under reported used pins. All seven raw splits now go through the shared `splitYamlDocLines`; each fix is pinned by a CRLF test verified to fail against the previous code.

- follow up to #1606, same defect class as #1605

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1608, opened 2026-08-02
CR-strip the substitutions and automation line scans

bdraco:

Follow up from the #1607 review, closing out the CRLF defect class. parseSubstitutions returned an empty map on CRLF configs (the header regex matched but the `(.*)$` value regex could not), so pin numbers, navigator labels, and id reference names all rendered raw `$var` tokens; it now splits through `splitYamlDocLines`. The sixth readInstanceScalar feeder in yaml-automations gets the same swap, its sub entity name read was CRLF blind and navigator automation rows fell back to raw ids. Also adds the missing regression test for findUsedPins on a CRLF long form pin block, the one converted path in #1607 without one.

Each fix is pinned by a test verified to fail against the previous code.

- follow up to #1607, same defect class as #1605

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/util/substitutions.ts
+++ b/src/util/substitutions.ts
@@ -8,6 +8,8 @@
  * reach those sources without a backend round-trip.
  */
 
+import { splitYamlDocLines } from "./yaml-doc-lines.js";
+
 /** Strip surrounding quotes / an inline ``# comment`` from a raw scalar. */
 function rawScalar(raw: string): string {
   const v = raw.trim();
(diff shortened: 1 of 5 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

PR #1609, opened 2026-08-02
Extract a shared reachability follower for the drawer and troubleshoot dialog

### LINKED ISSUE #1600: Extract a shared reachability follower for the drawer, troubleshoot dialog, and status report
There are now three independent consumers of `subscribeDeviceReachability`, each carrying its own copy of the subscribe, reconnect, failure, and teardown handling:

1. The device drawer (`src/components/dashboard/device-drawer-content/reachability.ts`): `reconcileSubscription` / `openSubscription` / `teardownSubscription` plus the 1 Hz tick, with a `device:generation` failure key and rate limited logging.
2. The offline troubleshooting dialog (#1597): `_startReconcile` / `_reconcileSubscription` / `_subscribe` / `_teardownSubscription` plus `_subscribedGeneration`, `_failedGeneration`, and `_subscribePending`, with the failed gate cleared on reopen.
3. The device status report prefill (#1599): `captureReachabilitySnapshot`, a one shot subscribe, first event, unsubscribe, bounded by a timeout.

The copies already diverge in bookkeeping style (failure keyed on device plus generation vs generation alone, `isCurrent()` staleness checks vs an in flight pending flag), and both long lived copies had review rounds fixing the same class of race independently: the dialog's in flight double subscribe and stale failed gate were caught by review after the drawer had already solved its equivalents differently. The next WS reconnect or ack race fix will land in whichever copy the reporter happened to hit and silently miss the others.

Follow up: extract one shared reachability follower, either a `ReachabilitySubscriptionController` ReactiveController (peer of `src/util/dialog-open-controller.ts`) parameterized on enabled state, device name, and the api handle, or a small `util/reachability-follower.ts` the hosts delegate to. It should own the generation tracking, the per generation failure memo, the in flight guard, teardown, and the 1 Hz reconcile tick; the one shot capture in #1599 can then be a thin wrapper that follows until the first event and stops.

Sketch:

```ts
// src/util/reachability-follower.ts
export interface ReachabilityFollowerOptions {
  api: () => ESPHomeAPI | undefined;
  /** esphome device name to follow; empty string means idle. */
  deviceName: () => string;
  /** Gate: follow only while true (drawer open, dialog open). */
  enabled: () => boolean;
  onEvent: (state: ReachabilityStateEvent) => void;
}

export class ReachabilityFollower implements ReactiveController {
  constructor(host: ReactiveControllerHost, options: ReachabilityFollowerOptions) {}

  // hostConnected/hostDisconnected own the 1 Hz reconcile tick.
  // Internals (private): subscription handle, subscribed generation,
  // per (device, generation) failure memo with one warn per key, and
  // the in-flight guard so a pending subscribe is never doubled.

  /** Clear the failure memo; call on an explicit user retry (reopen). */
  retry(): void {}

  /** Tear down now (after-hide), independent of host disconnect. */
  stop(): void {}
}

// One-shot capture (#1599) becomes a thin wrapper:
export async function captureReachabilitySnapshot(
  api: ESPHomeAPI,
  deviceName: string,
  timeoutMs = 3000
): Promise<ReachabilityStateEvent | null> {
  // follow until the first event or the timeout, then stop.
}
```

Hosts then shrink to construction plus `retry()` on reopen: the drawer drops its `_subscribedDevice` / `_subscribedGeneration` / `_failedSubscribeKey` / `_loggedFailureKey` fields and the dialog drops `_subscription` / `_reconcileTimer` / `_subscribedGeneration` / `_failedGeneration` / `_subscribePending`.

Refactor only, no behavior change intended. Best done after esphome/device-builder-frontend#1597 and esphome/device-builder-frontend#1599 both merge so the extraction unifies all three call sites at once.

bdraco:

Extracts one shared `ReachabilityFollower` (a Lit ReactiveController in `src/util/reachability-follower.ts`) from the two hand-rolled copies of the `subscribeDeviceReachability` lifecycle: the device drawer's `reconcileSubscription` / `openSubscription` / `teardownSubscription` / `syncTickInterval` free functions and the troubleshoot dialog's `_startReconcile` / `_reconcileSubscription` / `_subscribe` / `_teardownSubscription` methods. The copies had already diverged in bookkeeping (failure keyed on `device:generation` with a rate-limited warn vs generation alone; synchronous field registration vs an explicit pending flag), and review rounds fixed the same race class in each independently.

The follower owns the subscription handle, the subscribed name and connection generation, a per `name:generation` failure memo with a once-per-key warn, stale-attempt discard on both resolution and every event push (currency is a per-attempt id that teardown bumps, so a teardown-and-return to the same target within one subscribe round-trip cannot double-subscribe or leak the first stream — the property the dialog's old `_subscribePending` flag guarded), and the 1 Hz reconcile tick. Hosts declare what to follow through a `deviceName()` closure (null means idle) plus `onEvent` / `onTeardown` callbacks and an opt-in `tickRender` for second-precision age re-renders; the follower reconciles itself after every host update (`hostUpdated`), so neither host maintains a trigger list, the interval arms and disarms itself off the same signal, and `hostDisconnected` tears everything down. The unified semantics are the drawer's proven discipline applied to both hosts.

The drawer keeps only its render functions in `device-drawer-content/reachability.ts` and drops seven host fields (including the write-only `_tick` render counter); the dialog drops five. The dialog's untracked explainer-only mode now falls out structurally: an untracked target reads as idle, so reconcile tears down and disarms instead of needing the special timer stop the last review round added.

Deliberate deviations from the issue sketch, called out for review:

- `reachability-snapshot.ts` (the one-shot status-report capture) stays standalone. Its contract is a bounded promise where a subscribe rejection resolves `null` immediately; routing it through the poll-driven follower would turn that into a full timeout wait and drag generation and interval concerns it deliberately lacks.
- One benign behavior change: reopening the dialog onto the same device keeps the established stream instead of dropping and resubscribing (reconcile sees an identical target), and the last reachability snapshot is kept alongside it, since the kept stream redelivers nothing to repaint a blanked section. The reopened-onto-another-device case still tears down; tests pin both shapes.

The new `test/util/reachability-follower.test.ts` unit-tests the whole lifecycle (in-flight no-op, stale resolve, failure memo and warn rate-limit, generation-bump recovery, retry, idle teardown, tick recovery, disconnect), which also closes an existing gap: the drawer's subscription logic previously had no tests at all.

**Related issue or feature (if applicable):** fixes #1600

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/dashboard/device-drawer-content.ts
+++ b/src/components/dashboard/device-drawer-content.ts
@@ -232,8 +232,10 @@ export class ESPHomeDeviceDrawerContent extends LitElement {
     `;
   }
 
-  protected updated(changed: Map<string, unknown>) {
-    super.updated?.(changed);
+  // willUpdate, not updated: controllers' hostUpdated (the follower's
+  // reconcile) runs before the host's updated, so a memo cleared there
+  // would miss this cycle and wait out a 1 Hz tick.
+  protected willUpdate(changed: Map<string, unknown>) {
     // The dashboard re-binds device on every DEVICE_UPDATED push (state flap,
     // IP/version change). Only reset state when the drawer is reused for a
     // *different* device — compare configuration so same-device updates
--- a/src/components/dashboard/device-drawer-content.ts
+++ b/src/components/dashboard/device-drawer-content.ts
@@ -127,7 +127,7 @@ export class ESPHomeDeviceDrawerContent extends LitElement {
   @state() _ipExpanded = false;
 
   // Registered via addController; drives itself from host updates.
-  readonly _follower = new ReachabilityFollower(this, {
+  private readonly _follower = new ReachabilityFollower(this, {
     api: () => this._api,
     deviceName: () => (this.drawerOpen && this.device ? this.device.name : null),
     onEvent: (state) => {
@@ -244,6 +244,11 @@ export class ESPHomeDeviceDrawerContent extends LitElement {
       (previousDevice?.configuration ?? null) !== (this.device?.configuration ?? null);
     if (deviceTargetMoved) {
       this._ipExpanded = false;
+    }
+    if (deviceTargetMoved || changed.has("drawerOpen")) {
+      // A device switch or drawer reopen is an explicit fresh chance;
+      // don't carry a failed-subscribe memo across it.
+      this._follower.retry();
     }
   }
 }
--- a/src/components/dashboard/device-drawer-content.ts
+++ b/src/components/dashboard/device-drawer-content.ts
@@ -122,17 +122,14 @@ export class ESPHomeDeviceDrawerContent extends LitElement {
   // snapshot.value + (now - anchor) / 1000, advanced by the 1Hz tick.
   @state() _reachabilityAnchorMs = 0;
 
-  // Tick counter the relative-time renderer reads from to force a 1Hz re-render.
-  @state() _tick = 0;
-
   // Collapsed by default — flips when the user clicks the chevron on a
   // multi-IP device (typical when IPv6 is in play).
   @state() _ipExpanded = false;
 
-  private readonly _follower = new ReachabilityFollower(this, {
+  // Registered via addController; drives itself from host updates.
+  readonly _follower = new ReachabilityFollower(this, {
     api: () => this._api,
-    deviceName: () =>
-      this.drawerOpen && this.device && this._api ? this.device.name : null,
+    deviceName: () => (this.drawerOpen && this.device ? this.device.name : null),
     onEvent: (state) => {
       this._reachability = state;
       this._reachabilityAnchorMs = Date.now();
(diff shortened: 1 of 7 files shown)


bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - `isCurrent()` can't tell two concurrent attempts apart — leaks a subscription on teardown-then-resubscribe

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

PR #1620, opened 2026-08-06
Use direct map iteration for dependent-job lookups

### LINKED ISSUE #1617: Use direct map iteration for single-match job lookups in commands.ts
`src/components/command-dialog/commands.ts` has two single-match job lookups written as `[...jobs.values()].find(...)`, which copies the whole jobs map into an array before scanning:

- `deriveFollowCommandType` (dependent lookup for a live COMPILE head)
- the dependent flash lookup in `followJob`'s result handler

`findDependentUpload` in the same file already uses the cheaper `for (const j of jobs.values())` with an early return, and the wake re-follow added in #1616 does the same; these two should be made consistent with it.

Related consideration while in there: the command dialog is the one job-following surface that does not consume `activeJobsContext` (the per-configuration live-job map the app shell already maintains); some of these predicate scans might collapse into a context read, though the matching semantics differ (any live job vs a specific shape), so that needs its own evaluation.

bdraco:

Replaces the two remaining `[...jobs.values()].find(...)` lookups in commands.ts with a shared `findDependent` helper that iterates the jobs map directly, matching the style `findDependentUpload` already used; that helper now delegates too. The chain tail's type pair lives in one `CHAIN_FLASH_TYPES` constant, and `deriveFollowCommandType` maps the found dependent through the existing `JOB_TYPE_TO_COMMAND` table instead of re-branching on its type.

- fixes #1617

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/command-dialog/commands.ts
+++ b/src/components/command-dialog/commands.ts
@@ -18,6 +18,9 @@ const JOB_TYPE_TO_COMMAND: Record<JobType, CommandType> = {
   [JobType.RENAME]: "rename",
 };
 
+// The flash tail a compile head chains into.
+const CHAIN_FLASH_TYPES: readonly JobType[] = [JobType.UPLOAD, JobType.RENAME];
+
 // A live COMPILE with a held dependent is a chain head — attach in the
 // chain's command mode so success reflects the flash, not just the build.
 // Terminal reattach keeps plain compile mode: it's a log-review path and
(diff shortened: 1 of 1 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Both taken: `UPLOAD_TYPES` hoisted beside `CHAIN_FLASH_TYPES`, and the constant's comment now records that these are the only job types carrying depends_on.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Reworded to "the only job types that ever hold a non-empty depends_on".

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1630, opened 2026-08-09
Satisfy dependency checks from package-resolved components

bdraco:

Adding a component to a config built on a remote package showed a false "requires other components" detour; an adopted Apollo device pulling `github://ApolloAutomation/...` was told to add the ESP32 platform even though the package already provides `esp32:`. The gate only scanned the editor buffer's top level keys, so anything supplied by `packages:` was invisible.

The backend already resolves these configs in the background (`--only-generate` writes StorageJSON on every save), so the device row's `loaded_integrations` plus `target_platform` is the ground truth. The device page provides that list through a Lit context, every add-component dialog mount consumes it, and `findMissingDependencies` widens its presence set with it, but only when the YAML root merges other sources; plain configs keep the live literal scan so deleting a block re-flags its dependents immediately. The widening applies to dependency satisfaction only, single-instance catalog hiding stays on the literal scan so a local override block on top of a package remains addable.

Known limitations are tracked as follow-ups: dotted deps such as `ota.esphome` still cannot be satisfied from a package (#1631), and `depends_on_component` field visibility keeps the literal scan (#1632).

- fixes: user report, no linked issue

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/add-component-deps.ts
+++ b/src/components/device/add-component-deps.ts
@@ -26,6 +26,11 @@ const PLATFORM_DOMAINS: ReadonlySet<string> = new Set(Object.values(ComponentCat
  *  - `resolvedComponents` (the device's backend-resolved list) carries it
  *    and the YAML root-merges other sources — see `withMergedSourcePresence`.
  *
+ * `resolvedComponents` satisfies bare deps only: it is a flat list of bare
+ * names, and matching a dotted dep by stem would false-satisfy (`ota.esphome`
+ * vs the always-loaded `esphome` core key), so dotted deps stay on the
+ * configured-platform scan even for merged-source configs.
+ *
  * `presentComponents` may be passed precomputed to avoid re-parsing
  * the top-level blocks when the caller already has them.
  */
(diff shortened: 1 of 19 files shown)


bdraco:
@esphbot review

bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Second add-component-dialog mount is not threaded — the false detour persists on the "+ Add new <domain>" path
> - Core-locked widening in filters.ts is unreachable — `add-config-dialog` is the only `lockedCategories` producer and doesn't forward the prop

bdraco:
Addressed the review: the two flagged dialog mounts were real misses, and rather than threading the prop into each one the device page now provides the resolved list through a Lit context that every add-component dialog, form, and catalog consumes directly; a future mount can no longer forget the wiring, and the core locked filter branch is reachable from the real add-config dialog. Also documented the dotted dep limitation (stem matching would false satisfy ota.esphome via the always loaded esphome core key) and the merged config staleness tradeoff. @esphbot review

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1639, opened 2026-08-12
Pass the typed address override through to network logs

bdraco:
# What does this implement/fix?

The logs picker's advanced "Device IP or hostname" box was ignored; `launchLogsWithMethod` dropped the `port` argument for the `ota` method and always opened the default OTA target, so the backend ran `esphome logs --device OTA` and the CLI still did MQTT IP discovery even when the user typed the address. Pass the typed address through to the logs dialog, matching how `applyInstallMethod` handles the same picker for installs. With an explicit address the CLI uses it verbatim and skips the MQTTIP lookup entirely.

**Related issue or feature (if applicable):**

- fixes https://github.com/esphome/esphome/issues/18311

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [ ] New feature (non-breaking change which adds functionality) — `new-feature`
- [ ] Enhancement to an existing feature — `enhancement`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [ ] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `docs`
- [ ] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Checklist

- [x] The code change is tested and works locally.
- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.
- [x] Tests have been added to verify that the new code works (where applicable).

Issue #539, opened 2026-06-01
States toggle is shown for serial logs but only applies to OTA/network logs


bdraco:
### The problem

When viewing **serial** logs (a server serial port, e.g. `/dev/cu.usbserial-110`, or browser Web Serial), the logs dialog still shows the **States** toggle in the toolbar. States are an API / OTA concept; they only arrive over the device's network connection, so there is nothing to toggle on a serial session. Clicking it just tears the stream down and respawns it with a `--no-states` flag that has no effect on serial output.

### Steps to reproduce

1. Open Logs for a device and pick a serial source (Plug into computer running ESPHome Device Builder, i.e. a server serial port; the screenshot is on `/dev/cu.usbserial-110`).
2. The toolbar shows a **States** button next to expand / download / clear.

### Expected

The States toggle should appear only for the network / OTA log source, not for serial sources (server serial or Web Serial). For Web Serial it is already hidden; the gap is server serial.

### Notes / likely cause

In `src/components/logs-dialog.ts` the States toggle (`renderTermToggle`) renders whenever the session is **not** passive. Web Serial is passive (toggle hidden), but server serial is opened via `open(port)` and becomes the same `ota` session as a real OTA / network session, so it inherits the toggle. The gate should be the actual source being network / OTA, not just non passive; the `ota` session carries the target port, so a network session (`port === "OTA"`) versus a serial device path can be distinguished there.

PR #1640, opened 2026-08-12
Keep the States toggle for typed-address network logs

bdraco:

`isOtaNetwork` gated the States toggle on the port being the `OTA` sentinel, so a typed address override from the logs picker (or a post install hand off carrying one) hid the toggle even though the stream still runs over the backend API where `--no-states` applies. The predicate's real intent since #539 is to exclude server serial device paths, so test for that axis directly: a new `isServerSerialPath` helper classifies `/dev/...` and `COMn` as serial, everything else as network. Session tagging was considered instead, but the job reattach path restores the port from the backend `FirmwareJob` as a bare string, so a tag would still need this same classification at that boundary; the backend classifies serial ports by path shape the same way.

- follow-up to the review finding on #1639 (https://github.com/esphome/device-builder-frontend/pull/1639#issuecomment-5263017699)

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/logs-session.ts
+++ b/src/components/logs-session.ts
@@ -25,7 +25,6 @@
  * - ``dead``         a Web Serial reopen failed; the port is gone. Start runs
  *                    the reconnect hook (the #636 "click Start to reconnect").
  */
-import { OTA_PORT } from "../api/types/streaming.js";
 export type LogsSession =
   | { readonly kind: "idle" }
   | {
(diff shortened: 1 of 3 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Suggestion 1 added in b3568cf5: the toggle-restart case now opens on a typed address and asserts both spawns carry it. Skipping the extended `\\.\COMn` form on purpose; the backend's port list comes from pyserial's list_ports, which yields bare `COMn` on Windows, so the extended form never reaches the frontend and the miss would only show a no-op toggle anyway.

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1655, opened 2026-08-17
Match board locks by dotted path so nested clk fields lock

bdraco:

Board locked presets never locked nested fields in the visual editor. On esp32-poe-iso the ethernet clk group (pin and mode) rendered editable while mdc_pin and mdio_pin showed the board lock; the board data already carries locked_pins keyed as clk.pin plus a locked clk mapping preset, but every lock consumer matched flat keys only.

The pin wiring guard now matches locked_pins against the field's dotted path and reads nested presets through getIn. The value dependent lock overlay matches an id-less featured entry by its section (ethernet has no id preset, so the overlay previously never matched that section at all) and descends mapping presets into nested children, locking each leaf that still sits on its preset value. The duplicate card pin scan resolves dotted locked_pins keys the same way. Verified in the live editor against esp32-poe-iso: every board defined ethernet field shows the lock including clk pin and mode, and moving clk pin off the board wiring releases only that field.

- None, reported directly with a screenshot of the esp32-poe-iso ethernet editor

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/pin/renderer.ts
+++ b/src/components/device/pin/renderer.ts
@@ -10,7 +10,11 @@ import type { BoardPin } from "../../../api/types/boards.js";
 import type { ConfigEntry } from "../../../api/types/config-entries.js";
 import { PinFeature, PinMode } from "../../../api/types/config-entries.js";
 import { findUsedPins, sectionEndLine } from "../../../util/config-entry-yaml-scan.js";
-import { isPlainObject, isPrimitiveOrNullish } from "../../../util/nested-values.js";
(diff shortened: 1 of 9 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1657, opened 2026-08-18
Surface Web Serial unavailability in the flash receiver

### LINKED ISSUE #1656: [web.esphome.io] Safari: flash-receiver never reports "Web Serial not supported", just hangs on "waiting…"
## Steps to reproduce
1. Use Safari on macOS (yes I know it is unsupported; that is not the point for this issue)
2. Have a Home Assistant installation running the ESPHome Device Builder add-on.
3. Compile any ESP YAML file into new firmware.
4. Click "Open USB Flasher".
Image

5. The flasher page opens and shows no error about Web Serial.

## Expected behaviour
The flasher page (opened as `https://web.esphome.io/#nonce=SOMERANDOMNUMBER&origin=http%3A%2F%2Fhomeassistant.local%3A8123`) should show that Web Serial is not supported on Safari — the same message you get from a plain visit to web.esphome.io.

## Actual behaviour
The flasher page (same URL shape) shows "waiting…" and never resolves. On Safari this is a dead end with no explanation, which is a frustrating experience for regular Safari users who have no way to know their browser is the problem.
Image

## Double-check performed
Opening the flasher page with no `nonce` / hash at all *does* correctly show the "Web Serial not supported" message on the same Safari install. So Web Serial support detection works — it's just not wired up on the nonce/opener code path.
Image

## Environment
- Hardware: Home Assistant Green
- Home Assistant Core: 2026.8.2
- Supervisor: 2026.07.5
- Operating System: 18.2
- Frontend: 20260729.7
- ESPHome Device Builder: 2026.7.4
- Browser: Safari on macOS

## Root cause

> please note I have very little experience with javascript coding, and the assumptions below are created with help of AI chat (Claude)

The flasher app shell picks one of two components to render, based purely on `window.opener` and the hash (`esphome-web-app.ts`):

```ts
private _flasherMode =
  window.opener != null && parseFlasherParams(window.location.hash) != null;
```

- **No opener, no hash** → renders `<esphome-web-dashboard>`, which checks support up front:

  ```ts
  // esphome-web-dashboard.ts:54-56
  if (!isWebSerialSupported()) {
    return html`<esphome-web-unsupported-card></esphome-web-unsupported-card>`;
  }
  ```

- **Opener + nonce present** (i.e. exactly the case when Home Assistant's dashboard opens the flasher for a real flash) → renders `<esphome-web-flash-receiver>` instead. That component never calls `isWebSerialSupported()` / `webSerialAvailability()` anywhere. `navigator.serial` is only touched deep inside `_runInstall()` (`esphome-web-flash-receiver.ts:217`), which only runs after the user clicks "Connect & install" *and* firmware has already arrived over `postMessage`.

So on an unsupported browser, the receiver still happily starts the `FlashHandshake`, receives the firmware, and sits in "connecting"/"waiting" state forever — there is no code path that ever tells the user their browser can't do this. This is exactly the path Home Assistant's dashboard always uses (it always passes `nonce` + `origin`), so **every Safari user hitting "Open USB Flasher" hits this dead end**, while a bare visit to web.esphome.io correctly reports it.

### What were you trying to do?

Install / flash firmware

### Browser and operating system

Safari 26.5.2 (21624.2.5.11.8), MacOS 26.5.2 (25F84)

### Device / board

ESP32 (any)

### Console / log output

```shell
see above
```

### Anything else?

_No response_

wouterpotters:

The flash-receiver page (`<esphome-web-flash-receiver>`, the tab the dashboard
hands firmware to over `postMessage` when it can't flash itself) never checked
Web Serial support — only `<esphome-web-dashboard>` did. On a browser without
Web Serial (e.g. Safari, or an insecure context), the receiver rendered its
normal flash UI and sat there after firmware arrived, and no error was relayed
to the opener, so the dashboard — already handed off — waited out its full
10-minute watchdog on a tab that could never flash.

`ESPHomeWebFlashReceiver` now checks `webSerialAvailability()` once at
construction and, when unavailable:

- renders the existing `<esphome-web-unsupported-card>` (inside the normal
  `.wrap` width container) instead of the flash UI, and
- relays an `error` state to the opener as soon as firmware arrives — before
  retitling the tab or retaining the firmware buffer — so the dashboard fails
  immediately with a message the user can act on, instead of hanging.

Both the unsupported-browser and insecure-context messages are covered by
tests.

- fixes #1656

## Types of changes

## Checklist

- [x] The code change is tested and works locally, using `window.open("http://localhost:5174/#nonce=test123&origin=http://localhost:5174")`.
- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

## Remarks for reviewers

- Only tested by hand on Safari (macOS); not exercised on Windows/Linux or in
  other browsers.

### DIFF
--- a/src/web/flash-receiver/esphome-web-flash-receiver.ts
+++ b/src/web/flash-receiver/esphome-web-flash-receiver.ts
@@ -68,7 +68,8 @@ export class ESPHomeWebFlashReceiver extends LitElement {
   // The dashboard always hands off through this component (opener + nonce),
   // so unlike <esphome-web-dashboard> this is the only place that ever tells
   // a Safari (or insecure-context) user their browser can't do this at all.
-  private readonly _unsupported = webSerialAvailability() !== "available";
+  private readonly _serialAvailability = webSerialAvailability();
+  private readonly _unsupported = this._serialAvailability !== "available";
   // Supersedes a pending boot-log acquisition (a second manual flash during
   // the re-enumeration wait, or an unmount mid-await).
   private _bootLogsGen = 0;
@@ -107,6 +108,20 @@ export class ESPHomeWebFlashReceiver extends LitElement {
   }
 
   private _onFirmware(msg: FirmwareMessage): void {
+    if (this._unsupported) {
+      // The tab already shows <esphome-web-unsupported-card> (see render());
+      // relay it as an error too so the dashboard — now handed off — doesn't
+      // just sit waiting up to its own timeout for a tab that can never flash.
+      // Bail before retaining the firmware or retitling the tab "Flashing …":
+      // nothing below applies to a page that can't flash.
+      this._setState(
+        "error",
+        this._serialAvailability === "insecure-context"
+          ? this._localize("web.unsupported.insecure")
+          : this._localize("web.unsupported.browser")
+      );
+      return;
+    }
     this._firmware = msg;
     // Name the tab + card after the device so several concurrent flash tabs are
     // distinguishable (legacy did the same with the transmitted device name).
@@ -115,18 +130,6 @@ export class ESPHomeWebFlashReceiver extends LitElement {
       document.title = this._localize("web.flash.tab_title", {
         name: msg.deviceName,
       });
-    }
-    if (this._unsupported) {
-      // The tab already shows <esphome-web-unsupported-card> (see render());
-      // relay it as an error too so the dashboard — now handed off — doesn't
-      // just sit waiting up to its own timeout for a tab that can never flash.
-      this._setState(
-        "error",
-        webSerialAvailability() === "insecure-context"
-          ? this._localize("web.unsupported.insecure")
-          : this._localize("web.unsupported.browser")
-      );
-      return;
     }
     this._setState(
       "connecting",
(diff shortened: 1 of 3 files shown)


bdraco:
It probably shouldn't have gotten here in the first place the dialog shouldn't show in safari if no webserial
https://github.com/esphome/device-builder-frontend/issues/1656

esphbot (approved):
> [!TIP]

bdraco:
Addressed the review in 25c94e12:

- **#1 (full-bleed card):** the unsupported branch now reuses the existing `.wrap` container, and the test asserts the card is inside `.wrap` (plus no `esphome-web-card`, which is what the old `.wrap === null` assertion was really after).
- **#2 (tab title / retained firmware):** the `_unsupported` bail is now the first statement of `_onFirmware`, before `document.title` and `this._firmware` are touched; tests pin both.
- **"Computed once" contradiction:** `webSerialAvailability()` is now read once into `_serialAvailability`, with `_unsupported` derived from it, and the `insecure-context` arm has a test.

esphbot (approved):
> [!TIP]

bdraco (approved):
Thanks @wouterpotters

bdraco:
@esphbot review

PR #1660, opened 2026-08-18
Migrate device table to TanStack Table v9

bdraco:

Takes the TanStack Table v9 major that dependabot proposed in #1649; v9 is a breaking rewrite, so the bump alone fails typecheck and tests. The row model factories move into a shared `tableFeatures` object (new `table-features.ts`), `VisibilityState` becomes `ColumnVisibilityState`, column defs use `sortFn`, and the table types gain the features generic. Behavior is unchanged; sorting, filtering, column visibility and pagination keep the same controlled state flow.

Supersedes #1649.

- supersedes #1649

## Types of changes

- [x] Dependencies bump — `dependencies`

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "@lit/reactive-element": "2.1.2",
     "@mdi/js": "7.4.47",
     "@replit/codemirror-indentation-markers": "^6.5.3",
-    "@tanstack/lit-table": "^8.21.3",
+    "@tanstack/lit-table": "^9.1.2",
     "codemirror": "^6.0.2",
     "esptool-js": "^0.6.1",
     "improv-wifi-serial-sdk": "^2.8.1",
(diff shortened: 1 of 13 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Both suggestions addressed in 270c207b. Added a test pinning the auto resolved alphanumeric order on the platform column; the fixture includes host2/host10 and a case mixed value so a silent basic fallback fails the assertion rather than passing with the same order. Tightened the table-features docstring to name the unregistered datetime and basic outcomes. Moved DeviceRow into a leaf device-row.ts so table-features and table-columns no longer import each other; table-columns re-exports the type, so call sites are unchanged.

bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Suggestion 1 was already addressed in c0f1e39d, which the review notes it could not see; it pins the text entry by sorting the digit-free comment column over kitchen, Bedroom, attic and asserting attic, Bedroom, kitchen, an order that fails under both basic and textCaseSensitive. Suggestion 2 is fixed in the commit above, the two device-row type imports are now one statement.

@esphbot review

bdraco:
@esphbot review

PR #1663, opened 2026-08-19
Spell out the migration sentence keys so translators can grep them

bdraco:

Translators and a grep for orphaned keys both flagged `config_migration_change_key` / `_field` / `_fold` / `_convert` / `_action` as unused after #1662; they are used, but through a template literal built from `change.kind`, so the literal key never appears in the source. The lookup is now an explicit per kind map, so every key is greppable from `en.json` and a new backend kind has to be given a sentence here instead of silently falling through to the raw key.

For reference, the placeholders: `{old}` / `{new}` are YAML option names, `{scope}` is the block or `domain.platform` they live in; e.g. `voc` changed to `voc_index` in `sensor.sgp4x`, or `rgb_order` and its flags folded into `channel_colors` in `light.esp32_rmt_led_strip`.

- follow up to #1662 (Discord translator report)

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/config-migration-copy.ts
+++ b/src/components/device/config-migration-copy.ts
@@ -1,11 +1,21 @@
 /** Phrasing for the config-migration nudge: one sentence per ``MigrationChange``. */
 import { html, type TemplateResult } from "lit";
-import type { MigrationChange } from "../../api/types/editor.js";
+import type { MigrationChange, MigrationChangeKind } from "../../api/types/editor.js";
 import type { LocalizeFunc } from "../../common/localize.js";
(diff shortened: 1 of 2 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Added the per kind table test; each of the five kinds is pinned to its own sentence key, so a copy paste swap like fold pointing at _field now reds the suite. Agreed on the lockstep pushback, no runtime fallback wanted.

bdraco:
@esphbot review

PR #1665, opened 2026-08-20
Use wa-tooltip on the editor toolbar and reveal toggle

bdraco:

The editor toolbar's icon buttons (reveal sensitive values, diff toggle, the three layout switches) used native title attributes, which show the browser's slow unstyled tooltip. They now use the same sibling wa-tooltip pattern as the dashboard card actions, so hovering shows the styled tooltip immediately. The reveal toggle is shared, so the migration preview dialog's eye button gets the same treatment.

Buttons keep their aria labels and drop title. Touch devices stay tooltip free through the existing global wa-show suppression, verified with touch emulation. One tradeoff, the diff toggle's tooltip cannot show while the button is disabled, since disabled buttons do not fire hover events; the native title did show there.

Deliberately deferred to a follow up so this stays small: the install action buttons and the editor actions kebab in the same chrome still use native title.

- follow-up to #1664

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.


bdraco:
@esphbot review

bdraco:
Cleanup pass in 8eacb470. The three layout buttons collapsed into one local render helper, which also stops localizing each label twice; the shared reveal toggle documents its one instance per shadow root constraint instead of growing an id parameter; the tooltip anchor test helper graduated from test/web to test/ since it now has consumers in two areas, and the dialog test uses it too. Also noted in the description that the install action buttons and the actions kebab keep native title for a follow up.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Suggestion 1 is in as the tightened helper you sketched: expectTooltipsAnchored now also requires the anchor ids to be distinct within the shadow root, so a second reveal toggle rendered into one root fails the suite instead of silently piling both tooltips onto the first button. The disabled diff toggle tradeoff stays as stated, and the save button plus kebab remain for the follow up already noted in the description.

@esphbot review

PR #1671, opened 2026-08-21
Show the device comment on the tile card

Daniel085:

The legacy dashboard showed a device's `esphome.comment` on its tile; the new tile view doesn't. This renders the comment on the card under the configuration filename — quiet color, italic (matching the archived-devices dialog's comment rows), truncated to one line, and only rendered when the device carries one. The drawer and the opt-in table column already surface the comment; the tile was the one surface missing it.

No backend change needed — `ConfiguredDevice.comment` is already on the wire.

Verified in the live dashboard against a dev backend (three configs, two with comments), in dark and light themes: commented devices show the line, uncommented devices render no node, and card heights in a grid row stay equalized either way.

- https://github.com/orgs/esphome/discussions/3719 (Bring back display of Comment on Tile view)

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

Note on the test suite: the two new tests pass and the changed files are clean; the full local run shows 107 failures that reproduce identically on a clean `main` checkout under Node 25 (CI pins Node 24), so they're environmental, not from this change.

### DIFF
--- a/src/components/dashboard/render-content.ts
+++ b/src/components/dashboard/render-content.ts
@@ -131,6 +131,7 @@ export function renderCardGrid(
             data-configuration=${device.configuration}
             .name=${device.friendly_name || device.name}
             .configuration=${device.configuration}
+            .comment=${device.comment ?? ""}
             .state=${rt.state}
             .labelIds=${device.labels ?? []}
             ?has-pending-changes=${device.has_pending_changes === true}
--- a/src/components/device-card.ts
+++ b/src/components/device-card.ts
@@ -80,6 +80,8 @@ export class ESPHomeDeviceCard extends LitElement {
 
   @property({ attribute: false }) name = "";
   @property() configuration = "";
+  // ``esphome.comment`` from the YAML — free-form user text.
+  @property({ attribute: false }) comment = "";
   @property() state: DeviceState = DeviceState.UNKNOWN;
   // Truthy name_add_mac_suffix in the YAML — OFFLINE/UNKNOWN verdicts
   // are meaningless for such a device, so the badge shows "No status".
@@ -242,6 +244,11 @@ export class ESPHomeDeviceCard extends LitElement {
               ${renderEncryptionIcon(this)}
             </div>
             <p class="device-config truncate">${this.configuration}</p>
+            ${
+              this.comment
+                ? html`<p class="device-comment truncate">${this.comment}</p>`
+                : nothing
+            }
           </div>
           ${renderStatusBadge(this)}
         </div>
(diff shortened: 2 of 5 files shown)


bdraco:
I think this was intentionally not carried over from the old dashboard because it took up a lot of space on the dashboard page.  The downside to adding a comment is that esphome also bakes it into the flash on ESP32 and I think thats rodata on ESP8266 so RAM.  Something to probably fix on the esphome side before we would encourage more use cases

edit: looks like we fixed that in esphome/esphome#12554

esphbot (approved):
> [!TIP]

bdraco:
> ### 🟢 Suggestions
> 1. Pass-through binding (and its `null` coercion) has no test, despite an established pattern for exactly this
> [src/components/dashboard/render-content.ts:133](https://github.com/esphome/device-builder-frontend/blob/9057353b90c656460e5bf3a60d1a42847051b8e4/src/components/dashboard/render-content.ts#L133)
> The two new tests set `comment` directly on the card, so they cover the template branch but never exercise the one piece of logic this PR actually adds outside a template: `device.comment ?? ""`.
> 
> Why it matters: `ConfiguredDevice.comment` is `string | null` (`src/api/types/devices.ts:81`), and `test/_make-configured-device.ts:24` defaults it to `null`. If someone later drops the `??` or renames the field, both card tests still pass and the regression ships silently — a `null` would bind through and render `"null"` on every uncommented tile.
> 
> The repo already has a cheap pattern for this: `test/components/dashboard/render-content-migration-wiring.test.ts` and `render-content-encryption-wiring.test.ts` mount the real grid via `renderInto(renderCardGrid(host, [device]))` and assert on the resulting `<esphome-device-card>`. Two cases would close it:
> 
> * `makeConfiguredDevice({ comment: "Garage" })` → card's `.comment === "Garage"`
> * `makeConfiguredDevice({ comment: null })` → card's `.comment === ""` and no `.device-comment` node
> 
> ```
> .comment=${device.comment ?? ""}
> ```
> 
> 2. Truncated comment has no way to be read in full from the tile
> [src/components/device-card.ts:249](https://github.com/esphome/device-builder-frontend/blob/9057353b90c656460e5bf3a60d1a42847051b8e4/src/components/device-card.ts#L249)

Both suggestions are real issue and worth addressing

bdraco:
> * This would be the first _always-on_ comment surface — the table column is hidden by default and the drawer is on demand — which is exactly the space concern @bdraco raised. Gating it behind a preference would answer that, but needs a maintainer decision first.

Adding a new preference is worse, and since comments are opt in. I think its ok to add back once the other comments are addressed

bdraco:
Rebased onto main and addressed both review suggestions: the truncated comment line now carries the full text in its native title tooltip (same pattern as the logs dialog's truncated source chip), and a new render-content wiring test pins the card grid's comment ?? '' pass-through for both a real comment and a null one.

esphbot (approved):
> [!TIP]

bdraco (approved):

PR #1689, opened 2026-09-02
Ignore dot-prefixed top-level anchor keys in the section parser

### LINKED ISSUE #2651: [Bug] Dot-prefixed top-level anchor key (.defaultfilters:) misattributed to preceding section (TIME) in structured editor
### What happened?

**Description**
ESPHome core explicitly ignores any top-level configuration key starting with a dot (`.`) — this is a documented convention used to define YAML anchors (`&anchor`) without triggering schema validation errors (see https://esphome.io/guides/yaml/#anchors-aliases-and-overriding-values). Actual compilation with `esphome compile` is unaffected and produces a correct firmware.

However, in the structured/visual editor, the list items under such a dot-prefixed key are not recognized as belonging to an ignored key. Instead, they appear to be attributed to whatever real component section immediately precedes them in the file — in this case `time:` — and are rendered as bogus, unnamed "Time Source" cards.

**Steps to Reproduce**
1. Create/open a config where a `time:` section (a list, e.g. `platform: sntp`) is immediately followed by a dot-prefixed top-level key containing a list of YAML anchors, e.g.:
   ```yaml
   time:
     - platform: sntp
       id: my_time
       timezone: Europe/Vienna
       on_time_sync:
         then:
           - logger.log: "Synchronized system clock"

   .defaultfilters:
     - &moving_avg
       sliding_window_moving_average:
         window_size: 12
         send_every: 6
     - &throttle_avg
       throttle_average: 5s
     - &throttle_time
       throttle: 60s
     - &invert
       lambda: 'return max(-x, 0.0f);'
     - &pos
       lambda: 'return max(x, 0.0f);'
     - &abs
       lambda: 'return abs(x);'
   ```
2. Open the config in the Device Builder structured/visual editor.
3. Look at the left-hand navigator under "TIME".

### Expected Behavior
Only the actual `time:` entries should be listed under TIME (in this example: 1 entry — "SNTP Time Source"). The `.defaultfilters:` key and its list contents should be ignored entirely by the structured editor, consistent with how ESPHome core treats dot-prefixed top-level keys.

### Actual Behavior
The TIME section shows 7 entries: the 1 real SNTP time source, plus 6 additional, unnamed "Time Source" cards — one for each anchor entry (`moving_avg`, `throttle_avg`, `throttle_time`, `invert`, `pos`, `abs`) defined under `.defaultfilters:`.

This is purely a display/navigation issue in the structured editor — it does not affect YAML validity or compilation.

### Screenshot
Image

### Environment
- Device Builder version: 1.13.1
- ESPHome core version: 2026.8.2
- Browser/OS: Chrome 150.0.7871.115 / Windows 11 24H2
- Confirmed still present after updating to the latest available Device Builder release as of 02.09.2026.

### Additional Context
- This may be related to the general class of issues already being addressed around unknown/stray top-level keys, e.g. `frontend#1319` ("Offer an indent auto-fix when a stray top-level key belongs to the section above") and `frontend#1323` ("Hold the section switch while an unknown top-level key is being typed"). Those fixes don't appear to cover the specific case of ESPHome's dot-prefixed "ignored key" convention.
- Workaround in the meantime: edit this part of the config via the Monaco YAML view instead of the structured editor.
- Suggested fix: when the structured editor's parser encounters a top-level key starting with `.`, it should treat it (and its entire nested block) as ignored/hidden, matching ESPHome core's own behavior, rather than merging its list items into the previous recognized section.

### Dashboard version

2026.8.2

### How did you install it?

Home Assistant App

### Where is it running?

Home Assistant App

### Device YAML config

```yaml
not device specific
```

### Relevant logs / errors

```shell

```

### Anything else?

_No response_

bdraco:

ESPHome ignores top-level keys starting with a dot, the documented way to park YAML anchors. The section parser did not recognise them, so the preceding section never closed and the anchor list items rendered as bogus unnamed cards under it; the save path had the same blind spot, so editing the section above through the structured form silently deleted the dot key's header line.

The parser now closes the previous section at a dot-prefixed key and emits no section for it, and the shared block terminator treats a column-0 dot key as a sibling boundary, which fixes the save and delete splices too. The remaining column-0 scanners with their own inline matchers were widened the same way: the configured-platforms scan, the pin-usage scan and its section-end bound, the instance-scalar pool, and the cursor block walk (`findTopLevelBlock` now answers null inside a dot block while still walking past compact-sequence dashes, and shares the lexer's key regex instead of its own copy).

At review request, `ESPHOME_YAML_INDENT` moved out of the CodeMirror language module into a leaf `esphome-yaml-indent.ts`, breaking the import cycle the lexer import would otherwise close; `yaml-error-analysis` derives `YAML_INDENT_STEP` from it instead of a hand-rolled literal.

- fixes https://github.com/esphome/device-builder/issues/2651

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/yaml-editor.ts
+++ b/src/components/yaml-editor.ts
@@ -27,7 +27,8 @@ import {
 } from "../util/codemirror-theme.js";
 import { initialDarkMode } from "../util/dark-mode.js";
 import { editorSearchPhrases } from "../util/editor-search-phrases.js";
-import { ESPHOME_YAML_INDENT, esphomeYaml } from "../util/esphome-yaml-lang.js";
+import { ESPHOME_YAML_INDENT } from "../util/esphome-yaml-indent.js";
+import { esphomeYaml } from "../util/esphome-yaml-lang.js";
 import { fireEvent } from "../util/fire-event.js";
 import { idleCompletion } from "../util/idle-completion.js";
 import { createYamlCompletionSource } from "../util/yaml-completion.js";
--- a/src/util/config-entry-yaml-scan.ts
+++ b/src/util/config-entry-yaml-scan.ts
@@ -25,6 +25,11 @@ import { splitYamlDocLines } from "./yaml-doc-lines.js";
 import { readInstanceScalar } from "./yaml-instance-scalars.js";
 import { indentOf } from "./yaml-line-walker.js";
 import {
+  IGNORED_TOP_LEVEL_KEY_RE,
+  TOP_LEVEL_KEY_RE,
+  TOP_LEVEL_KEY_START_RE,
+} from "./yaml-section-lexer.js";
+import {
   collectIdsAtPath,
   findFieldLine,
   parseYamlTopLevelSections,
@@ -228,14 +233,17 @@ export function findUsedPins(
   // Indentation of an open free-text block scalar; continuation lines
   // indented deeper than this are prose and skipped. -1 when none is open.
   let blockScalarIndent = -1;
+  let inIgnoredBlock = false;
   for (let i = 0; i < lines.length; i++) {
     const line = lines[i];
-    const topMatch = line.match(/^([a-zA-Z_][a-zA-Z0-9_]*):/);
-    if (topMatch) {
-      currentDomain = topMatch[1];
+    const topMatch = line.match(TOP_LEVEL_KEY_RE);
+    if (topMatch || IGNORED_TOP_LEVEL_KEY_RE.test(line)) {
+      currentDomain = topMatch ? topMatch[1] : "";
+      inIgnoredBlock = !topMatch;
       blockScalarIndent = -1;
       continue;
     }
+    if (inIgnoredBlock) continue;
     const lineNo = i + 1;
     if (
       excludeFromLine !== undefined &&
@@ -297,7 +305,7 @@ export function sectionEndLine(yaml: string, fromLine?: number): number | undefi
   for (let i = fromLine; i < lines.length; i++) {
     const line = lines[i];
     if (line === "") continue;
-    if (/^[a-zA-Z]/.test(line)) return i;
+    if (TOP_LEVEL_KEY_START_RE.test(line)) return i;
   }
   return lines.length;
 }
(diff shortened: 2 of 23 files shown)


bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - findTopLevelBlock now returns null for a column-0 list dash, not just dot keys

bdraco:
1: fixed in 5e750393; a dot key answers null, other column-0 lines are skipped again, compact-sequence test added.
2: intended; the scanner is documented best-effort and alias-blind, noted for a follow-up.
3: no change; with 1 fixed the walk skips non-key column-0 lines again, so the rationale holds.
4: fixed in 5e750393.
Description now lists the extra scanner updates.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
1: fixed in 6a71c560; ESPHOME_YAML_INDENT moved to a leaf module, importers migrated, the cycle is gone, and yaml-error-analysis's hand-rolled indent literal now derives from it.
2: fixed in 6a71c560; sectionEndLine test covers the dot and underscore bounds.
3: acknowledged, stands as intended.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Both fixed in the new head: unquote's docblock keeps the semantic reason and drops the dissolved chain, sectionEndLine's states the inclusive-last-line return.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
1: taken in the new head; RE_TOP_LEVEL_KEY is gone, both walkers use the lexer's TOP_LEVEL_KEY_RE.
The unscoped-card note stands as a follow-up, an improvement over the old mis-attribution.
Description now covers the indent-constant extraction.

@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Taken in the new head; the flag is gone, the cleared domain plus the existing !currentDomain guard covers the block, GPIO7 test unchanged.

@esphbot review

PR #1712, opened 2026-09-05
Close the migration preview and troubleshoot dialogs through the wrapper's hide

### LINKED ISSUE #1708: Move the migration preview and troubleshoot dialogs onto the after hide close path
Follow up to #1707, which added `DialogOpenController.requestClose()` and moved the picker, desktop update and update all dialogs onto the after hide close path so a body rendered only while open survives the hide animation.

Two more dialogs gate their body on `open` but bind `@request-close` instead of `@after-hide`: `config-migration-preview-dialog.ts` and `troubleshoot-dialog.ts`. The request close event fires synchronously at the start of the hide, so their flag flips on the first frame and the body blanks for every close path, including Escape and the close button.

Switch both to `@after-hide=${this._dialog.onAfterHide}`, route any programmatic close through `this._dialog.requestClose()`, and add a one shot guard if an action button stays live through the animation, as #1707 did for the picker pick and the update all confirm.

bdraco:

Follow up to #1707. The config migration preview and troubleshoot dialogs render their body only while open but bound `@request-close`, which the wrapper emits synchronously at the start of every hide, so the flag flipped and the body blanked on the first frame of the animation for every close path, Escape and the close button included.

Both now bind `@after-hide` only and route their programmatic `close()` through `DialogOpenController.requestClose()`, so the body stays through the hide and the flag drops when the wrapper reports the hide finished. The migration preview's confirm button gets the same one shot guard as the picker pick and the update all confirm, since it stays live through the animation. Checked live on the troubleshoot dialog: 50 ms after `close()` the body and the native dialog are still up, 1.5 s later both are closed.

- fixes #1708

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/device/config-migration-preview-dialog.ts
+++ b/src/components/device/config-migration-preview-dialog.ts
@@ -60,13 +60,17 @@ export class ESPHomeConfigMigrationPreviewDialog extends LitElement {
     `,
   ];
 
+  /** Set on the first confirm; the button stays live through the hide animation. */
+  private _confirmed = false;
+
   open() {
     this._revealSensitive = false;
+    this._confirmed = false;
     this._dialog.open = true;
   }
 
   close() {
-    this._dialog.open = false;
+    this._dialog.requestClose();
   }
 
   protected render() {
@@ -76,7 +80,6 @@ export class ESPHomeConfigMigrationPreviewDialog extends LitElement {
         .label=${this._localize("device.config_migration_preview_title", {
           configuration: this.configuration,
         })}
-        @request-close=${this._dialog.onRequestClose}
         @after-hide=${this._dialog.onAfterHide}
       >
         ${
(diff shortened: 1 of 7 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Took all three: the reopen test now dispatches after-hide between close and open and checks the diff was torn down, so it exercises a real close again; the double click test continues through after-hide and a reopen and asserts the second confirm fires, pinning the _confirmed reset; and the wrapper stub plus open flag accessor live in test/_dom.ts as stubDialogRequestClose and dialogOpen beside baseDialog, used by both suites here.

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Took both nits: the update-all and picker open-contract suites now use stubDialogRequestClose and dialogOpen from test/_dom.ts instead of their local stubs, and the migration preview header names the close path and confirm guard coverage.

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1698, opened 2026-09-05
Render the catalog picker body only while the dialog is open

bdraco:
# What does this implement/fix?

The automation catalog picker rendered its full body while closed. The by target tab is one row per entity times matching action, and every action node mounts its own picker, so opening an automation in a config with hundreds of entities built tens of thousands of hidden rows and icons. A performance trace from the linked issue shows one 3 second render creating 19,847 `wa-icon` elements and the DOM growing from 100k to over a million nodes; the page stays sluggish afterwards because that subtree never goes away.

The picker now renders the search field, tabs and rows only while it is open, and drops them on close. Locally, a find jump into an automation with 11 action nodes and 500 entities went from 13,248 hidden picker rows and a 393 ms task to zero rows and no long task; an opened picker still lists its rows.

**Related issue or feature (if applicable):**

- fixes https://github.com/esphome/device-builder/issues/2657

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [ ] New feature (non-breaking change which adds functionality) — `new-feature`
- [ ] Enhancement to an existing feature — `enhancement`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [ ] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `docs`
- [ ] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Checklist

- [x] The code change is tested and works locally.
- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.
- [x] Tests have been added to verify that the new code works (where applicable).

PR #1715, opened 2026-09-06
Mount one navigator and inline its row icons

bdraco:

The device page mounted the navigator twice, a drawer copy for mobile and a desktop copy, and hid one with CSS, so every row was built twice. Each row also mounted one or two wa-icon elements and every domain subgroup started open, so a config with a few hundred entities held well over a thousand icon elements before the user touched anything. Same shape as #1698 and #1702, on a list that is usually larger than the picker.

One navigator is mounted per layout now: the drawer copy on mobile, the desktop copy on desktop, hidden by CSS while the sidebar is collapsed so its search and reveal state survive. The row and subgroup glyphs are inline svg through mdiSvg, and domain subgroups start collapsed except the one holding the last selected row, which stays open after a deselect or Back; a hand toggle still wins over the selection, and search still opens everything.

Measured on a config with 425 rows: one navigator instead of two, 14 wa-icon elements inside it instead of roughly 430 per copy, 105 rows rendered instead of 425.

- related to https://github.com/esphome/device-builder/issues/2657

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

bdraco:
Took 1 and 3 in 892563bb: a flat single-row domain no longer takes the open latch (the flatten rule is shared as isFlatDomain), and the group callbacks are required on the view. 2 is left: keeping both copies mounted is the cost this PR removes, and a resize across 900px is rare enough that lifting the per-instance state to the page can wait for a report.

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
No further changes: a hand-collapsed domain winning over an external reveal is the pre-existing contract the reveal tests pin, the resize state stays deferred, and the walker's directive limit predates this PR and is the same one that keeps #1716's candidate list on map.

esphbot (approved):
> [!TIP]

PR #1732, opened 2026-09-10
Keep the compile running when the install dialog is dismissed

bdraco:
# What does this implement/fix?

Closing the install dialog with X, Escape, or the backdrop cancelled the running compile, the same as pressing Stop. Now a dismissal releases the compile to the background queue and shows a toast pointing at Firmware tasks; only Stop cancels. The command dialog also offers a Download firmware button once a compile finishes, so a build that finished in the background, or one retried from Firmware tasks, still hands over the binary. The button reuses the three-dot Download flow, which compiles only when nothing is built.

Verified in the browser against a local backend: dismiss mid-compile leaves the job running to completion, Stop still cancels it, and both the background build and the retried build offer Download firmware from Firmware tasks.

**Related issue or feature (if applicable):**

- fixes https://github.com/esphome/device-builder/issues/2677

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [ ] New feature (non-breaking change which adds functionality) — `new-feature`
- [ ] Enhancement to an existing feature — `enhancement`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [ ] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `docs`
- [ ] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Checklist

- [x] The code change is tested and works locally.
- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.
- [x] Tests have been added to verify that the new code works (where applicable).

PR #1733, opened 2026-09-10
Render the Firmware tasks dialog inside the layout

bdraco:

The Firmware tasks dialog was rendered as a sibling of the layout, so events from its nested command and install dialogs bubbled past the listeners app-shell binds on the layout; the dialog re-fired them itself and app-shell carried a duplicate binding. This moves the dialog inside the layout so those events reach the existing listeners directly, and deletes the four forwarders plus the duplicate binding. The dialog still closes itself on open-settings so Settings does not stack over it; it just lets the event keep bubbling instead of re-firing it. A comment at the render site records why the placement matters, a happy-dom test pins the bubble-and-close behaviour, and the host gets display: contents now that it sits in the layout flow. Page-scoped events (open editor, clean build, download firmware, show logs) keep their local handlers; the clean-build one becomes a stable class property instead of an inline arrow, so the listener is not re-bound on every progress tick.

Verified in the browser: from the nested command dialog, open-settings opens Settings on the requested section, open-reset-build-env opens the reset confirm, and open-firmware-jobs still opens the dialog.

- follow-up to #1732

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/src/components/app-shell.ts
+++ b/src/components/app-shell.ts
@@ -596,6 +596,11 @@ export class ESPHomeApp extends LitElement {
         @open-guided-tour=${this._onOpenGuidedTour}
       >
         ${this._router.outlet()}
+        <!-- Inside the layout on purpose: its nested command and install
+             dialogs fire open-* events the layout listeners above handle. -->
+        <esphome-firmware-jobs-dialog
+          @firmware-history-cleared=${() => onFirmwareHistoryCleared(this)}
+        ></esphome-firmware-jobs-dialog>
       </esphome-layout>
       <esphome-command-palette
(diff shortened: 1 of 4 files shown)


bdraco:
@esphbot review

bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

PR #1735, opened 2026-09-10
Isolate the registerMdiIcons tests from shared mock history

bdraco:

The registerMdiIcons tests registered the library once in a beforeAll and then asserted on that call count from a later test. Vitest 5 turns on clearMocks by default, so the call history is wiped before each test and the "exactly once" assertion sees zero calls, which is what fails on #1729.

Each test now loads a fresh copy of the module and makes its own registration, so no test depends on an earlier one or on how mock history is carried between tests; the suite passes on vitest 4 and 5.

- unblocks https://github.com/esphome/device-builder-frontend/pull/1729

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

### DIFF
--- a/test/util/register-icons.test.ts
+++ b/test/util/register-icons.test.ts
@@ -1,8 +1,8 @@
-import { beforeAll, describe, expect, test, vi } from "vitest";
+import { beforeEach, describe, expect, test, vi } from "vitest";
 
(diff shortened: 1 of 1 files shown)


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1736, opened 2026-09-10
Resolve package_import_url hosts through a table keyed by domain

bdraco:

Follow-up to #1721. The three per-host branches in previewPackageImportUrl differed only in the URL template and the service literal, so they collapse into one table keyed by shorthand domain; the service union now derives from the table keys instead of being widened by hand for each new host. The module header and the refSegment comment still described the GitHub blob route as if it were universal, both now cover all three hosts.

The lookup is an own-property check rather than the in operator, since the shorthand grammar admits domains like constructor or toString that would otherwise resolve to a prototype function; a test pins that those fall through to plain text. No behaviour change for the existing shorthands.

- follows up https://github.com/esphome/device-builder-frontend/pull/1721

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
@esphbot review

PR #1750, opened 2026-09-19
Tolerate catalog fields the backend omits at their default

bdraco:

The backend is about to stop sending component catalog fields that hold their default, the way board data already works (companion PR linked below). A config entry will then arrive with only `key`, `type`, `label` and whatever differs from its default, which cuts a component body to about a third of its size.

This makes the frontend tolerate that ahead of the backend change. The `ConfigEntry` and `ComponentCatalogEntry` types mark every such field optional, so the compiler finds each place that assumed a field was present; that turned up six, all fixed here. The one list every consumer reads directly, a component's `config_entries`, is backfilled once in `getComponentBodies`, as `getAvailableAutomations` already does. Nothing compared these fields strictly against `null` or `false`, so an absent field reads the same as the default it replaces.

The shared `makeConfigEntry` factory was a hand kept table of thirty defaults whose comments said the type had no optional fields; it now builds only the three guaranteed fields, as does the component test factory, so the 77 test files that use them exercise wire shaped entries. One of the six fixes prevented a real crash, a catalog index row with no `dependencies` in the core configuration dialog, and now has a test.

It is harmless against today's backend, which still sends every field. Checked against a backend with the change in a browser: the Wi-Fi section renders its fields, validation, nested block and advanced settings from the slim payload with no console errors.

- companion backend PR: https://github.com/esphome/device-builder/pull/2762 (merge this one first)

## Screenshots

No visual change.

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.


bdraco:
@esphbot review

bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

PR #1751, opened 2026-09-19
Type catalog list rows without the body fields they never carry

bdraco:

Follow up to #1750. `components/get_components` returns slim list rows that never carry `config_entries` or `required_groups`, but `PagedComponentsResponse.components` was typed `ComponentCatalogEntry[]`, which promises both. Every current reader happens to guard with `?? []`, so nothing is broken, but the compiler had nothing to say to the next reader who does not; that is the same kind of crash #1750 fixed for `dependencies`.

This adds `ComponentCatalogIndexEntry`, the full entry without those two fields, and uses it for the list response, `fetchAllComponents`, the catalog component, its filters and card renderer, and the YAML completion and hover catalog index. A full entry is still assignable wherever a row is expected. No helper that now takes a row read a body field, so there is no behaviour change. The filters test builds its rows without casts, and one assertion from #1750 is tightened from falsy to undefined.

- none

## Screenshots

No visual change.

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
Both applied in 9deeb90e9afbf412804f3bdb45a0d599d5d407f7: `ComponentCatalogEntry` now extends `ComponentCatalogIndexEntry` so a new body field cannot leak into the row type, and the `add-component` event detail is typed as the index row.

@esphbot review

esphbot (approved):
> [!TIP]

PR #1754, opened 2026-09-19
Read absent automation catalog flags as false and backfill body config_entries

bdraco:

The backend is about to serialise the automation catalog models the way it already serialises the component catalog (esphome/device-builder#2762): a field holding its declared default is left off the wire. For triggers, actions, conditions, light effects and filters that means an absent `is_device_level`, `supports_list`, `is_control_flow`, `has_else_branch` or `accepts_condition_list` reads as false and an absent `applies_to` or `accepts_action_list` reads as empty. This PR makes those fields optional in the types, adjusts the three readers that indexed `applies_to` without a guard, and backfills `config_entries` on every body from `automations/get_bodies` the way `getAvailableAutomations` already does for its rows, so renderers keep reading `config_entries.length` synchronously.

Against the current backend nothing changes; every field is still sent. Lands before the backend change so the deploy order does not matter.

- none

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.


bdraco:
@esphbot review

PR #1761, opened 2026-09-20
Mask a secure child's value in a block's summary of what it holds

bdraco:

Follow up to #1759, from a Copilot comment that landed after it merged. An enabled block with no field to show summarises what it holds ("Sets Divider: 1"). That summary listed every primitive child value, so a secure string child such as a password or API key would have been printed in plain text, while its own field masks it.

The summary now shows a secure child as `••••••`; every other value reads as before. Test added.

- follow up to https://github.com/esphome/device-builder-frontend/pull/1759#discussion_r4056441224

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
On the review of 71a73d4, as of 343b8958:

- An undeclared key now goes through `maskSensitiveLine`, the repo's own YAML credential masker, rather than a second notion of sensitive; the test adds an `ota_password` the catalog does not declare.
- Also folded in here, from the last review of #1759 (same helper): the notice is built over the keys the form does not paint instead of only when nothing paints, so a demanded block that paints some fields still names a seeded child the filter drops. An entity block with fields is unchanged. Test added; emc2101 still reads "Sets Divider: 1" in the browser.

@esphbot review

bdraco:
Correction to my comment above: this PR had already merged at 71a73d4 when I pushed the second commit, so neither the undeclared key masking nor the unpainted keys notice is in it. Both are in #1762 instead. I have put this PR's description back to what it actually contains.

PR #1763, opened 2026-09-20
Leave the search term alone when the Filters menu clears filters

davidcoulson:

The Filters menu's **Clear filters** now clears the facet selections only, and leaves the search box alone. The search box's own × still clears the search, and the no-results **Clear filters** button still clears both.

Before this, both buttons shared `_clearAllFilters`, so opening the Filters menu and clearing it also wiped whatever you had typed. That's surprising in the menu: its badge deliberately counts facet selections only — a lone search term isn't a pill there (#1160) — so the button was clearing something the menu never showed, and you lose the search you'd narrowed to. With 150-odd devices I hit it constantly: filter by label, type a name, decide the label was wrong, clear it, and now the name's gone too.

The no-results case is left as it was. There the search term is one of the things hiding every device, so clearing it is the point of the button.

- `_clearFacets` — new; empties the five facet arrays, leaves `_search`.
- `_clearAllFilters` — now `_resetSearch()` + `_clearFacets()`.
- `renderFacets` wires the popover's `clear-filters` event to `_clearFacets`; `renderNoResultsExtras` keeps `_clearAllFilters`.

No copy changes, so no translation keys move.

- esphome/device-builder#1160, which is what the bare `#1160` in the existing
  comments points at — it lives in the backend repo, so a bare number here
  links to a TypeScript bump instead. My comments now spell it out in full;
  happy to fix the pre-existing bare ones in this PR or leave them alone.

## Screenshots

No pixel changes — the button, its label and the menu are identical. The difference is state, so it doesn't show in a still. To see it:

1. Type something in the search box (e.g. `rrn00`).
2. Open **Filters** and tick a label.
3. Click **Clear filters** inside the menu.

Before: the search box is empty and the full device list is back. After: the search box still says `rrn00`, the list is still narrowed by it, and only the label filter is gone.

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.

Verified against a live Device Builder (154 devices) with `pnpm run dev`, plus three new tests: `_clearFacets` keeps the search and drops the facets, `_clearAllFilters` still does both, and `renderFacets` wires the popover to the facet-only handler.


bdraco:
The linked pr is a typescript bump.

davidcoulson:
Sorry — I meant esphome/device-builder#1160, the backend issue the bare `#1160` in the existing comments refers to. A bare number resolves in this repo, hence the TypeScript bump.

Fixed the description, and spelled the reference out in full in the comments I added. Say the word if you'd like the pre-existing bare `#1160`s in `render-facets.ts`, `dashboard.ts` and the two tests qualified the same way and I'll fold that in.

esphbot (approved):
> [!TIP]

bdraco (approved):

PR #1766, opened 2026-09-20
Stop masking the eap key, which is a path to a PEM file

bdraco:

The YAML editor masked the value of `key:` under `wifi: eap:` as if it were a credential, so `key: client.key` showed as dots. That value is a path: esphome runs it through `cv.file_` and reads the PEM from the file (`wifi/wpa2_eap.py`), the same as the `certificate:` beside it, which was already shown.

The `eap` entry is removed from the parent scoped sensitive keys, so the path is visible in the editor and everywhere else the scanner is used. The `eap` password stays masked, and `key:` under `encryption:` is unchanged.

- none

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.


bdraco:
@esphbot review

PR #1767, opened 2026-09-20
Ask the credential rules about a key instead of a synthetic YAML line

bdraco:

Follow up to #1762, from a review that landed after it merged. The block summary ("Sets Divider: 1") decides whether to mask a value by building two lines of synthetic YAML and checking whether the masker changed them. That works, but it answers a question about a key by rewriting text, so it depends on the masker's line pattern (a key with a leading digit never matched) and has to know YAML indentation.

The scanner's two built in rules, always sensitive keys and keys that are sensitive only under a given parent, are now one exported predicate that its own two call sites use, and `isSensitiveKeyUnder(parent, key)` adds the editor's suffix rule on top. The summary calls that directly. No masking behaviour changes for YAML text.

Also adds the test the review asked for: an ordinary enabled entity block whose fields are painted shows no notice.

- follow up to https://github.com/esphome/device-builder-frontend/pull/1762#issuecomment-5749385973

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

bdraco:
On the review here (23a7679): all three notes are in #1769. The predicate folds case itself, the commented line branch finds the parent and asks once instead of carrying a `scoped` flag and an `undefined` sentinel, and the top of stack lookup is pinned by a local typed as possibly undefined.

PR #1770, opened 2026-09-20
Mask a commented credential that follows a live sibling

bdraco:

A commented out `# key:` under `encryption:` was masked when it came first under its parent and left in the clear when it followed a live sibling:

```yaml
api:
  encryption:
    algorithm: aes
    # key: c2VjcmV0
```

A comment pops nothing off the scanner's ancestor stack, so `algorithm` was still on top and was taken for the commented key's parent. Any ancestor in the chain to the left of the key's column now counts. The stack is a single chain, so this can only over mask; tests pin both directions, including a comment at column zero, one after the parent's block has ended, and one under a parent that does not scope the key.

This matters for crash reports, which mask comments; the editor leaves comments visible on purpose. The bug is older than the recent predicate refactors; Copilot spotted it while reviewing #1769.

Two review nits from #1769 ride along: the no parent case is an explicit guard again instead of an empty string map key, and the stack entry has a named type shared by the stack and the top of stack local.

- follow up to https://github.com/esphome/device-builder-frontend/pull/1769#discussion_r4057471379

## Types of changes

## Checklist

- [x] `pnpm run lint` passes.
- [x] `pnpm run test` passes.


bdraco:
@esphbot review

esphbot (approved):
> [!TIP]

